PR #38, opened 2023-09-22
FileWriter Implementation

bsowell:
I got this to work for my simple test case (write annotated PDF images to a local directory), but it still needs some work before it is ready to ship. I'm not likely to get to it before launch, so thought I'd stick it in a draft PR. 

The main design challenge comes from how to effectively reuse Ray's existing writer infrastructure. Ray has a `FileBasedDatasource` that does most of what we want. In the latest version, they even added support for writing each record as a separate file, which I think is the behavior we want by default. Normally you would just subclass `FileBasedDatasource` and implement the `_write_row` method, but unfortunately it currently hardcodes the filename to a long combination of the dataset UUID, the taskid, and the block/row_id. This might be okay if you are treating your data as rows that you want to process in bulk, but specifying the filename seems like a key requirement when dealing with documents. 

So that leads us to a choice -- implement something custom or try to reuse what's there. I took the latter approach -- I still extended the `FileBasedDatasource` class to get methods like `on_write_start`, but I overrode the entire `write` method rather than just `_write_row`. To do this I had to import some internal Ray methods like `_resolve_paths_and_filesystem`, which are not really intended to be used externally. These methods encapsulate quite a bit of complex functionality such as standardizing access across different "filesystems" like S3 and GCS that I really don't want to rewrite, though this dependency may not be stable across ray versions. Eventually I think the right thing to do is to add support for custom filenames to the `FileBasedDatasource` class in Ray itself. Based on some comments in the code, this is line with their plans, but I have no idea how long it will take. 

If people are okay with this approach, the remaining items are:
- Clean up the parameters, many of which are not used. 
- Write tests, including tests against multiple filesystems. 
- Fix random linting/testing issues.

### DIFF
--- a/sycamore/tests/unit/test_writer.py
+++ b/sycamore/tests/unit/test_writer.py
@@ -1,6 +1,53 @@
 from sycamore import DocSet, Context
+import sycamore
+from sycamore.data import Document
 from sycamore.plan_nodes import Node
 from sycamore.writers import OpenSearchWriter
+
+from pathlib import Path
+
+from sycamore.writers.file_writer import default_filename, default_doc_to_bytes, json_properties_content
+
+
+def generate_docs(num: int, type: str = "test", text=True, binary=False) -> list[Document]:
+    docs = []
+    for i in range(num):
+        doc = Document(
+            {
+                "doc_id": f"doc_{i}",
+                "type": "test",
+                "elements": {"array": []},
+                "properties": {"filename": f"doc_{i}.dat"},
+            }
+        )
+
+        if text:
+            doc["text_representation"] = f"This is text content {i}"
+        if binary:
+            doc["binary_representation"] = f"This is binary content {i}".encode("utf-8")
+
+        docs.append(doc)
+
+    return docs
+
+
+def _check_doc_path(
+    docs: list[Document], path: Path, filename_fn=default_filename, doc_to_bytes_fn=default_doc_to_bytes
+):
+    docs_by_filename = {filename_fn(doc): doc for doc in docs}
+    paths = [p for p in path.iterdir() if p.is_file()]
+    assert len(paths) == len(docs)
+
+    for p in paths:
+        with open(p, "rb") as infile:
+            expected = doc_to_bytes_fn(docs_by_filename[p.name])
+            actual = infile.read()
+
+            assert expected == actual
+
+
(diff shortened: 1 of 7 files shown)


bsowell:
There seems to be something that is environment specific about mypy -- it was working fine for me locally. Will address.

sycamore/tests/unit/test_writer.py
bohou-aryn, line 1: do we clean up write out files either in s3 or local?

sycamore/writers/file_writer.py
```diff
@@ -0,0 +1,155 @@
...
+            for row in block.iter_rows(public_row_format=True):
+                doc = Document(row)
+                filename = filename_fn(doc)
+
+                write_path = os.path.join(path, filename)
+                logger.debug(f"Writing file at {write_path}")
+
+                # with fs.open_output_stream(write_path) as outfile:
+                with _open_file_with_retry(write_path, lambda: fs.open_output_stream(write_path)) as outfile:
+                    if prefer_text and doc.text_representation is not None:
```
bohou-aryn, line 150: is there case, we want write both text and binary into same file, e.g an image? or is there a consideration for writing based on document type instead of a flag?

sycamore/writers/file_writer.py
```diff
@@ -0,0 +1,155 @@
...
+        base_name = str(uuid.uuid4())
+    else:
+        base_name = str(doc.doc_id)
+
+    if extension is not None:
+        return f"{base_name}.{extension.lstrip('.')}"
+    return base_name
+
+
+def doc_path_filename(extension: str, suffix: str):
```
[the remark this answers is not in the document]
bsowell: Ack. Will do.

sycamore/writers/file_writer.py
bsowell: Good question. This also completely ignores properties. Maybe calling this `files` is too expansive, since it's really just writing the main content. I think we would need a different writer for something like JSON, even if it still goes to a file. 

Agreed on document type. Maybe we have some kind of is_text/is_binary flag that we can use to distinguish? Not sure if we wan to add that in this PR or wait until we have a more holistic approach.

bohou-aryn (approved):

pparmar30 (approved):

PR #69, opened 2023-09-28
Quick lint rule to detect references to notion.so in docs

alexaryn:


bsowell (approved):
Looks reasonable to me. Your find knowledge clearly exceeds my own, but don't see any problems. I can imagine a few edge cases -- for example it is possible to publish public sites with notion, and we might use something like that for a design doc, but that'll be the exception rather than the rule.

bsowell:
Hmm. Any idea why it failed on the push, but succeeded on the PR? For that matter, do we need it to run on every branch and PRs? That means it will always run twice, right?

alexaryn:
Thanks for looking.  It's not a work of art, and I think it won't hard-block any commits.  We can tweak as necessary.  Eventually we can grow it into a Python script if we want more complex logic.

My theory for why the push failed and the pull-request succeeded is that my git repo still had the "bad" file in it when I pushed.  I tried to do some rebasing and re-pushing afterward, but couldn't trigger the workflow again.  I'd like to keep an eye on it to see what happens in the future.  We can always remove the workflow if it's problematic.  On the other hand, there are other repos where this workflow could be handy.

PR #70, opened 2023-09-29
add demo vid as additional resource

HenryL27:
adds the demo video to the sycamore docs page


bsowell (approved):
Nice. Maybe add a link in the readme too under resources?

PR #78, opened 2023-10-03
UnstructuredPdfPartitioner to inherit parent properties

alexaryn:
Modify UnstructuredPdfPartitioner to inherit parent document properties down to child elements.  For now, just "path", but extensible.  This allows properties to be directly available in search results without the complication and expense of retrieving the parent document.


bohou-aryn (approved):
Might worth to add a unit test.

Issue #57, opened 2023-09-28
Complete API Documentation


bsowell:
We need to complete documentation of the main classes and transforms in Sycamore.

pparmar30:
I am currently working on this

PR #81, opened 2023-10-03
Add and Update API docs

pparmar30:
Resolves https://github.com/aryn-ai/sycamore/issues/57


docs/source/APIs/index.rst
```diff
@@ -1,7 +1,9 @@
...
+=================
 
 .. toctree::
-   :maxdepth: 1
+   :maxdepth: 2
 
-   functions.rst
+   functions/functions.rst
+   data/data.rst
+   transforms/transforms.rst
```
bsowell, line 9: Minor: I would order this as data structures, then transforms, then functions. I think it will be easier to understand the functions once you look at the other two categories.

sycamore/functions/document.py
```diff
@@ -30,6 +57,27 @@ def split_and_convert_to_image(doc: Document) -> list[Document]:
...
+        default_color: The default color for bounding boxes when the element type is unknown.
+
+    Example:
+
+         .. testcode::
+
+            drawer = DrawBoxes(font_path="path/to/font.ttf", default_color="blue")
+            image_docs =
+                [Document(binary_representation=image_data, properties={"size": (width, height), "mode": "RGB"})]
+            labeled_image_docs = drawer(image_docs)
```
bsowell, line 78: Did you generate these examples? It's interesting b/c none of them reference the DocSet, which is how we would expect people to use them. Ideally, I think a better example would be showing how to use this in a map function.

sycamore/functions/chunker.py
```diff
@@ -9,7 +9,25 @@ def chunk(self, tokens: list[Any]) -> list[Any]:
 
 
 class TokenOverlapChunker(Chunker):
-    def __init__(self, chunk_token_count=1000, chunk_overlap_token_count=100) -> None:
+    """
+    TokenOverlapChunker is a class for chunking tokens into smaller segments while allowing for token overlap.
```
bsowell, line 13: This is really for chunking text rather than tokens. In fact token is a bit of a misnomer here, since we aren't using an actual tokenizer -- just words/characters.

sycamore/functions/document.py
```diff
@@ -8,6 +8,33 @@
...
 def split_and_convert_to_image(doc: Document) -> list[Document]:
+    """Split a document into individual pages as images and convert them into Document objects.
+
+    This function takes a Document object, which may represent a multi-page document, and splits it into individual
+    pages. Each page is converted into an image, and a new Document object is created for each page. The resulting
+    list contains these new Document objects, each representing one page of the original document and elements making
+    up the page.
+
+    The input Document object should have a binary_representation attribute containing the binary data of the
+    document (e.g., a PDF).Each page's elements are preserved in the new Document objects, and page-specific properties
```
bsowell, line 19: Since this only works for PDFs, I would say "containing the binary data of the pdf document" and drop the e.g..

sycamore/docset.py
```diff
@@ -17,7 +17,10 @@
 
 
 class DocSet:
-    """DocFrame is a distributed computation framework for Documents."""
+    """
+    A DocSet, short for “documentation set,” is a distributed collection of documents bundled together for processing.
```
bsowell, line 21: Doesn't have to come right away, but I think eventually we may want method documentation for DocSet. To some extent that may be more useful than the transform docs, since most users will probably invoke the transforms via the DocSet APIs.

docs/source/APIs/transforms/transforms.rst
```diff
@@ -0,0 +1,16 @@
+.. _Ref-Transforms:
+
+Transforms
+===========
+
```
bsowell, line 5: Maybe we should add a note here that transform classes are primarily of interest to developers looking to extend Sycamore? Most users of Sycamore won't need to interact with these classes or know about things like Nodes.

bsowell (commented):
Great start. Thanks for doing this! Few comments to start with, but I think this is something we can incrementally improve over time.

bsowell:
Also, can you fix the linting issue?

docs/source/APIs/index.rst
pparmar30: Updated

sycamore/functions/document.py
pparmar30: Updated

sycamore/functions/chunker.py
pparmar30: Updated

sycamore/functions/document.py
pparmar30: Updated

sycamore/docset.py
pparmar30: Updated most of them

docs/source/APIs/transforms/transforms.rst
pparmar30: Updated

bsowell:
@pparmar30 Thanks for the changes. Only thing is I think you inadvertently broke the test_html_to_opensearch by adding a `filter`

     TypeError: DocSet.filter() missing 1 required positional argument: 'f'

bsowell (approved):

PR #84, opened 2023-10-04
Add FileMetadataProvider

baitsguy:
This change adds a file-level metadata provider. The provider is invoked for each "document" read by BinaryScan and it's returned dictionary is merged with the element properties. The first metadata provider just reads from a JSON file where properties are keyed by the filePath. This allows us to build workflows against simple crawler results.

```
s3://mybucket/directory/
...TencentSort2016.pdf
...ExoshuffleCloudSort2022.pdf
...file3.pdf
...manifest.json
```

manifest.json would look like:
```json
{
  "s3://mybucket/directory/TencentSort2016.pdf": {
      "remote_url": "http://sortbenchmark.org/TencentSort2016.pdf",
      "indexed_at": "2023-10-04 and some time"
  },
  "s3://mybucket/directory/file2.pdf": {
      "remote_url": "http://sortbenchmark.org/ExoshuffleCloudSort2022.pdf",
      "indexed_at": "2023-10-04 and some time"
  }
}
```
and then in sycamore, the following would read in our files and associated metadata
```python
context.read.binary("s3://mybucket/directory/",
  binary_format="pdf",
  metadata_provider=JsonManifestMetadataProvider("s3://mybucket/directory/manifest.json")
)
```

This can also be used to inject document level IDs. We'll want to create a set of reserved system-level properties after this (e.g. prefixed with _, like '_remote_url", "_document_id", "_created_at")

### DIFF
--- a/sycamore/tests/unit/scans/test_binary_scan.py
+++ b/sycamore/tests/unit/scans/test_binary_scan.py
@@ -27,36 +27,20 @@ class TestBinaryScan:
         remote_url = "https://en.wikipedia.org/wiki/Binary_search_algorithm"
         indexed_at = "2023-10-04"
         manifest = {
-            base_path + "/wikipedia_binary_search.html": {
-                "remote_url": remote_url,
-                "indexed_at": indexed_at
-            },
-            "other file.html": {
-                "remote_url": "value",
-                "indexed_at": "date"
-            },
-            "non-dict element": {
-                "key1": "value1",
-                "key2": [
-                    "listItem1",
-                    "listItem2"
-                ]
-            },
-            "list property": [
-                "listItem1",
-                "listItem2"
-            ]
+            base_path + "/wikipedia_binary_search.html": {"remote_url": remote_url, "indexed_at": indexed_at},
+            "other file.html": {"remote_url": "value", "indexed_at": "date"},
+            "non-dict element": {"key1": "value1", "key2": ["listItem1", "listItem2"]},
+            "list property": ["listItem1", "listItem2"],
         }
-        tmp_manifest = tempfile.NamedTemporaryFile(mode='w+')
+        tmp_manifest = tempfile.NamedTemporaryFile(mode="w+")
         try:
             json.dump(manifest, tmp_manifest)
             tmp_manifest.flush()
             manifest_path = tmp_manifest.name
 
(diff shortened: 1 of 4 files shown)


pparmar30:
- How does one create manifest.json?
- If it is up to the user to create it, is there a way to auto-extract some metadata depending on the source (s3 vs local path)?

baitsguy:
> * How does one create manifest.json?
> * If it is up to the user to create it, is there a way to auto-extract some metadata depending on the source (s3 vs local path)?

The one here is intended to be generated by a crawler that we'll write. Once we start reading properties from this metadata for downstream tasks, it's up to the creator to ensure the fields are added. 

> is there a way to auto-extract some metadata depending on the source 
We can probably consider adding one, for e.g. add s3 object metadata. By default we don't really get much from Ray's binary reader

sycamore/tests/unit/scans/test_binary_scan.py
```diff
@@ -17,3 +21,29 @@ def test_partition(self):
...
+        base_path = str(TEST_DIR / "resources/data/htmls/")
+        remote_url = "https://en.wikipedia.org/wiki/Binary_search_algorithm"
+        indexed_at = "2023-10-04"
+        manifest = {
+            base_path + "/wikipedia_binary_search.html": {"remote_url": remote_url, "indexed_at": indexed_at},
+            "other file.html": {"remote_url": "value", "indexed_at": "date"},
+            "non-dict element": {"key1": "value1", "key2": ["listItem1", "listItem2"]},
+            "list property": ["listItem1", "listItem2"],
+        }
+        tmp_manifest = tempfile.NamedTemporaryFile(mode="w+")
```
bsowell, line 35: More of a convenience, but you can use the tmp_path fixture in pytest if you want to avoid having to manage this yourself. https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

bsowell (approved):
LGTM. We'll probably need some more experience to figure out how we will use this in practice, but factoring looks nice.

sycamore/tests/unit/scans/test_binary_scan.py
baitsguy: Nicee! Let me add this in the follow up where I create system properties

Issue #16, opened 2023-09-08
Standardize format for element coordinates.


bsowell:
Currently some operations access element coordinates, for example to determine whether the document is one column or two column: https://github.com/aryn-ai/sycamore/blob/c11108998d59b671d91e087ea68a2b14efcea1d3/python/sycamore/execution/transforms/partition.py#L36

This currently relies on the internal coordinate representation produced by Unstructured. Since we hope to eventually have more partitioners that don't rely on Unstructured, we should have our own independent specification for coordinates.

PR #87, opened 2023-10-05
Standarize the bounding box representation.

bohou-aryn:
We use ratio to document page width/height as unit for coordinates.
Related : https://github.com/aryn-ai/sycamore/issues/16


sycamore/transforms/partition.py
```diff
@@ -23,9 +23,9 @@ def _elements_reorder_comparator(element1: Element, element2: Element) -> int:
     # following function checks if the x0 point of the element is in the
     # left column
     def element_in_left_col(e: Element) -> bool:
-        width = e.properties["coordinates"]["layout_width"]
-        x0 = e.properties["coordinates"]["points"][0][0]
-        return x0 / width <= 0.5
+        if e.bbox is None:
+            raise RuntimeError("Element BBox is None")
```
pparmar30, line 27: Do we want to raise an error or fail safe?

bohou-aryn: boundinbox/coordinates for element should not be None at least based on current algorithm, when you say fail safe, how do we behave here? return false/true by default?

pparmar30 (approved):

bsowell (approved):

PR #103, opened 2023-10-12
Transform to denormalize specified properties from parents to children

alexaryn:
The new SpreadProperties transform copies properties from a Document to its Elements.  This denormalization can be useful to provide efficient access to the properties from the retrieved documents at search time.  The current use is for 'title' and 'path' in examples/simple_ingest.py.  Its use is optional.

This supersedes the previous code in the PDF partitioner that inherited the path from parents.  A unit test is included.

### DIFF
--- a/sycamore/tests/unit/transforms/test_spread_properties.py
+++ b/sycamore/tests/unit/transforms/test_spread_properties.py
@@ -1,0 +1,37 @@
+from sycamore.data import Document
+from sycamore.transforms import SpreadProperties
+
+
+class TestSpreadProperties:
+    def test_spread_properties(self):
+        dict0 = {
+            "doc_id": "doc_id",
+            "type": "pdf",
+            "text_representation": "text",
+            "binary_representation": None,
+            "parent_id": None,
+            "properties": {"path": "/docs/foo.txt", "title": "bar"},
+            "elements": {
+                "array": [
+                    {
+                        "type": "UncategorizedText",
+                        "text_representation": "text1",
+                        "properties": {"filetype": "text/plain", "page_number": 1},
+                    },
+                    {
+                        "type": "UncategorizedText",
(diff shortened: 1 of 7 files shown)


sycamore/tests/unit/transforms/test_spread_properties.py
```diff
@@ -0,0 +1,37 @@
+from sycamore.data import Document
+from sycamore.transforms import SpreadProperties
+
+
+class TestSpreadProperties:
```
pparmar30, line 5: We should also add a test which actually operates on the dataset to verify the underlying Ray calls.

alexaryn: I looked at the other tests and they're very confusing with all the mocking calls.  Do we need to repeat that testing here?

bsowell: Do we need mocking. Can we just run the transform on one of the docs we have committed (or synthetic data)?

alexaryn: I am testing the logic, but just not through Ray.  Maybe someone can show me a straightforward way to test more of the wrapper code.

bohou-aryn:
I feel we should put this in the explode transformation, before explode, this is not needed.

sycamore/tests/unit/transforms/test_spread_properties.py
alexaryn: I'll add a unit test that uses execute().

alexaryn:
@bohou-aryn, I think it's still optional if API users want this behavior.  Their UI could also fetch properties from the parent documents.  It could be better in some cases.

pparmar30 (approved):

bohou-aryn (approved):

PR #104, opened 2023-10-12
Increase nbmake timeout to address integ test issues.

bsowell:


alexaryn (approved):
LGTM.  I assume removing the paths restriction was deliberate.

bsowell:
> LGTM. I assume removing the paths restriction was deliberate.
I'm actually going to undo that -- it was just to trigger the tests on this PR to make sure they work!

Issue #96, opened 2023-10-08
Support for PPTX files


bsowell:
We've had some requests to support PowerPoint. We can likely add very basic support via Unstructured, but we should also investigate some more sophisticated segmentation methods.

PR #107, opened 2023-10-12
Add support for PPTX files

pparmar30:
Addresses https://github.com/aryn-ai/sycamore/issues/96


sycamore/transforms/partition.py
```diff
@@ -49,6 +49,91 @@ def partition(self, document: Document) -> Document:
...
+            )
+
+            context = sycamore.init()
+            pdf_docset = context.read.binary(paths, binary_format="pptx")
+                .partition(partitioner=pptx_partitioner)
+
+    """
+
+    @staticmethod
+    def to_element(dict: dict[str, Any]) -> Element:
```
bsowell, line 84: Does the powerpoint partitioner return a bounding box? Do we have to do anything to convert it to the bbox representation?

sycamore/transforms/partition.py
```diff
@@ -49,6 +49,91 @@ def partition(self, document: Document) -> Document:
...
+        **kwargs
+    ):
+        self._include_page_breaks = include_page_breaks
+        self._include_metadata = include_metadata
+        self._include_slide_notes = include_slide_notes
+        self._chunking_strategy = chunking_strategy
+        self._kwargs = kwargs
+
+    def partition(self, document: Document) -> Document:
+        from unstructured.partition.pptx import partition_pptx
```
bsowell, line 116: Question rather than a comment -- is there a Ray reason for importing these as part of the method rather than at the top?

sycamore/tests/unit/transforms/test_partition.py
```diff
@@ -122,3 +127,13 @@ def test_partition_html(self, mocker, path, partition_count) -> None:
         docset = partition.execute()
         doc = docset.take(limit=1)[0]
         assert len(doc["elements"]["array"]) == partition_count
+
+    @pytest.mark.parametrize("path, partition_count", [(TEST_DIR / "resources/data/pptx/design.pptx", 71)])
+    def test_partition_pptx(self, mocker, path, partition_count) -> None:
+        scan = mocker.Mock(spec=BinaryScan)
```
bsowell, line 133: Are we actually executing the partition logic here? What is the mocking doing for us?

sycamore/transforms/partition.py
```diff
@@ -49,6 +49,91 @@ def partition(self, document: Document) -> Document:
...
+    Example:
+         .. code-block:: python
+
+            pptx_partitioner = UnstructuredPPTXPartitioner(
+                include_page_breaks=True,
+                strategy="auto",
+                infer_table_structure=True,
+                ocr_languages="eng",
+                max_partition_length=2000,
+                include_metadata=True,
```
bsowell, line 74: These look like the pdf arguments?

sycamore/transforms/partition.py
pparmar30: It does not! We will prob have to implement the partitioner if we wanted that.

sycamore/transforms/partition.py
pparmar30: Nop! Just to bring in the dependency as needed.

sycamore/tests/unit/transforms/test_partition.py
pparmar30: Yes, we are. The following two lines are setting up the mock behavior :
```
execute: Callable[[], Dataset] = _make_scan_executor(path, "pptx")
mocker.patch.object(scan, "execute", execute)
```

pyproject.toml
```diff
@@ -31,6 +31,7 @@ boto3 = ">=1.28.61"
 # See https://stackoverflow.com/a/76647180
 torch = "2.0.0"
 pdfminer-six = "20221105"
+python-pptx = "^0.6.22"
```
bsowell, line 34: Do you still need this if you are using unstructured?

bsowell (approved):

pyproject.toml
pparmar30: Yes! The tests were failing without it.

Issue #105, opened 2023-10-12
[FEATURE] Element Merging


HenryL27:
**Is your feature request related to a problem? Please describe.**
We often see partitioning models 'over-partition' documents. That is, split documents at such granularity that a lot of context gets lost in the resulting elements, which are too small and too many to be meaningful. It would be nice to be able to merge elements after-the-fact into larger elements.

**Describe the solution you'd like**
A new transform (maybe call it `merge_elements`?) to merge elements based off of some user-provided logic (create an `ElementMerger` class?) within a document. Merging should not be a full collapse of the elements, just mutate the list of elts to a smaller list of bigger elts.

**Describe alternatives you've considered**
This could also be part of the partitioning process - perhaps some kind of partitioning post-processing step to squish things back together. Then you don't have to worry about merging element properties. But it makes the partition interface more complicated.

**Additional context**

HenryL27:
Merging the `text_representation` is easy, simply concatenate. We need to make decisions about how to deal with all the other properties though.
- Type: maybe we can concatenate these too? Just strings after all. so you end up with "Title text text text" or something. Probably the cleanest way of showing what elements make up this merged element.
- Binary representation: the bits might not have been split so cleanly, so concatenation might not successfully reconstruct
- bbox: This should become a list of bounding boxes. Not sure what that means for the data model.
- properties: Hard to say what to do here since these are pretty free. Union the keys, surely, but what do you do with the overlappiing keys w diff values? not sure

PR #115, opened 2023-10-17
merge_elements transform

HenryL27:
transform for greedily merging small elements into larger elements with more context.

Resolves: #105


sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,130 @@
...
+            element2 (Tuple[Element, int]): the second element (and number of tokens in it)
+
+        Returns:
+            Tuple[Element, int]: a new merged element from the inputs (and number of tokens in it)
+        """
+        elt1, tok1 = element1
+        elt2, tok2 = element2
+        new_toks = 0
+        new_elt = Element()
+        new_elt.type = "Section"
```
baitsguy, line 54: How are we handling table (or other typed) elements?

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,130 @@
...
+            new_elt.bbox = BoundingBox(
+                min(elt1.bbox.x1, elt2.bbox.x1),
+                min(elt1.bbox.y1, elt2.bbox.y1),
+                max(elt1.bbox.x2, elt2.bbox.x2),
+                max(elt1.bbox.y2, elt2.bbox.y2)
+            )
+        # Merge properties by taking the union of the keys
+        for k, v in elt1.properties.items():
+            new_elt.properties[k] = v
+        for k, v in elt2.properties.items():
```
baitsguy, line 82: so element 2's properties are going to override? I think this merging is non-trivial, do we have an open issue for this

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,130 @@
...
+
+class ElementMerger(ABC):
+    @abstractmethod
+    def merge_elements(self, document: Document) -> Document:
+        pass
+
+class GreedyElementMerger(ElementMerger):
+
+    def __init__(
+        self,
```
baitsguy: can you define types for these

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Args:
+            document (Document): A document with elements to be merged. 
+
+        Returns:
+            Document: The same document, with its elements merged
+        """
+        if len(document.elements) < 2:
+            return document
+        token_counts = [len(self.tokenizer.encode(e.text_representation or "")) for e in document.elements]
```
baitsguy, line 100: we have a tokenizer interface already, let's use that instead

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,130 @@
...
+        new_elts = [(document.elements[0], token_counts[0])]
```
baitsguy, line 101: why are we using tuples/treating elements and token counts separate? Token count is a property of element I'd think

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,130 @@
...
+from typing import Tuple
+
+
+
+class ElementMerger(ABC):
+    @abstractmethod
+    def merge_elements(self, document: Document) -> Document:
+        pass
+
+class GreedyElementMerger(ElementMerger):
```
baitsguy: Should be explicit this is a text based merger, we'll want other types of mergers that will also be using the greedy approach (e.g. merge overlapping boxes until text is full)

sycamore/tests/unit/transforms/test_merge_elements.py
```diff
@@ -0,0 +1,111 @@
...
+                },
+                {
+                    "type": "UncategorizedText",
+                    "text_representation": "text2",
+                    "properties": {"filetype": "text/plain", "page_number": 2},
+                },
+                {
+                    "type": "Title",
+                    "text_representation": "title1",
+                    "bbox": [0.17606372549019608,
```
baitsguy: These tests need to be readable, can you change the coordinates to be something more readable, same with text_representation if we can reduce the size

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,130 @@
...
+            new_elt.text_representation = elt1.text_representation + '\n' + elt2.text_representation
+            new_toks = tok1 + 1 + tok2
+        # Merge bbox by taking the coords that make the largest box
+        if elt1.bbox == None and elt2.bbox == None:
+            pass
+        elif elt1.bbox == None or elt2.bbox == None:
+            new_elt.bbox = elt1.bbox or elt2.bbox
+        else:
+            new_elt.bbox = BoundingBox(
+                min(elt1.bbox.x1, elt2.bbox.x1),
```
baitsguy, line 74: can't this result in the bounding box include other elements? seems to assume that boxes are ordered visually

sycamore/transforms/merge_elements.py
HenryL27: tables special data is stored in properties, so that gets persisted. If two tables get merged then this implementation will drop the second table's "rows" and "columns" values. But otherwise 'merged' elements are just basic Elements

sycamore/transforms/merge_elements.py
HenryL27: element 2's properties get overwritten by element1's, yes. #105 is the issue for this feature; I didn't really know what to do with the keys. If you have a suggestion I'll take it

sycamore/transforms/merge_elements.py
HenryL27: ack

sycamore/transforms/merge_elements.py
HenryL27: it's not explicitly in the dataclass, so I didn't assume it. I agree, I could probably add that to the elements but I didn't want to step on anything else's toes

sycamore/transforms/merge_elements.py
HenryL27: ack. `GreedyTextElementMerger`?

sycamore/transforms/merge_elements.py
HenryL27: can totally result in overlapping boxes; I'm not convinced that's a huge issue though - maybe cold be a feature wherein you go back and forth with text merging and geometric merging until your elements are stable and large? The big assumption that this makes is that the elements are on the same page - not sure how we should be handling cases where that isn't true.

sycamore/tests/unit/transforms/test_merge_elements.py
HenryL27: ack

sycamore/transforms/merge_elements.py
HenryL27: ack

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,124 @@
...
+    Merge Elements into fewer large elements
+    """
+
+    def __init__(self, child: Node, merger: ElementMerger, **kwargs):
+        super().__init__(child, **kwargs)
+        self._merger = merger
+
+    def execute(self) -> Dataset:
+        input_dataset = self.child().execute()
+        dataset = input_dataset.map(generate_map_function(self._merger.merge_elements))
```
bohou-aryn, line 123: for tokenizer these heavy operation which requires loading models, we always try to minimize the number of times as much as possible, may use mp_class to enable Ray's actor exeuction.

sycamore/transforms/merge_elements.py
bohou-aryn: I feel we should not merge two different types of elements? e.g. author with title, or table with figure etc?

sycamore/transforms/merge_elements.py
bohou-aryn: not only bounding box, but also text. Are the ordering across multiple columns handled here(Roland do some work here)? In summary, is there any concern to have this as a standalone transformation without ordering?

sycamore/transforms/merge_elements.py
HenryL27: I guess it sort of depends on how we're handling types? My impression was that types are essentially inherited directly from unstructured, in which case we want to be able to merge 'Title' and 'NarrativeText' - that's the whole point. Basically we want to squash context together, and that can come from different sources. 

If we are gonna make our type system more uniform then it might make sense to filter only the 'Text' elements for this transform or something, but as it is I think anything with a text_representation can and should be merged (given sufficient room in the context window)

sycamore/transforms/merge_elements.py
HenryL27: multi-column is not a part of this transform, no. I think there's a different transform that does that? But yes, I do assume some ordering of the elements

sycamore/transforms/merge_elements.py
HenryL27: huggingface tokenizer doesn't download the model, though it does download some stuff, so prob worth that.

as for mp_class... what? how?

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,124 @@
...
+        elt1, tok1 = element1
+        elt2, tok2 = element2
+        new_toks = 0
+        new_elt = Element()
+        new_elt.type = "Section"
+        # Merge binary representations by concatenation
+        if elt1.binary_representation is None or elt2.binary_representation is None:
+            new_elt.binary_representation = elt1.binary_representation or elt2.binary_representation
+        else:
+            new_elt.binary_representation = elt1.binary_representation + elt2.binary_representation
```
pparmar30, line 60: Not sure if it matters, but I am assuming we will be able to decode it back to text if required?

HenryL27: I think so, but I haven't tested all kinds of documents. This is just doing concatenation of the bytes. If the data is in a format that doesn't concatenate into something meaningful then you won't get something meaningful. Maybe we should add more knobs, i.e.
ElementMerger takes as params a merge function for each element field. So something like 
```
ElementMerger(
    binary_merge_fn,
    text_merge_fn,
    bbox_merge_fn,
    type_merge_fn,
    property_merge_fn
)
```
Could also throw in a `should_merge_fn` and make it fully general? But again, you could just say, "You want different behavior? Then implement your own subclass! You only need to write one function, it's not that hard"

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,128 @@
...
+
+from sycamore.data import Document, Element, BoundingBox
+from sycamore.plan_nodes import NonCPUUser, NonGPUUser, Transform, Node
+from sycamore.transforms.map import generate_map_class_from_callable
+from sycamore.functions.tokenizer import Tokenizer
+
+from typing import Tuple
+
+
+class ElementMerger(ABC):
```
bsowell, line 13: This is such a generic interface, it's effectively equivalent to just passing the merger to the map function, right? The logic still makes sense, I guess I'm just wondering if there is anything more to put in the abstraction. Is it worth exposing an interface that just merges to elements, and then we are responsible for calling it? Is there any logic you expect to be common to all mergers?

sycamore/transforms/merge_elements.py
```diff
@@ -0,0 +1,128 @@
...
+            return False
+        return True
+
+    def _merge(self, element1: Tuple[Element, int], element2: Tuple[Element, int]) -> Tuple[Element, int]:
+        """
+        Merge two elements; the new element's fields will be set as:
+
+            type: "Section"
+            binary_representation: elt1.binary_representation + elt2.binary_representation
+            text_representation: elt1.text_representation + '\n' + elt2.text_representation
```
bsowell, line 37: Why the newline? OpenAI's embedding models actually recommend stripping newlines, though I don't know if it matters for sentence transformers.

sycamore/transforms/merge_elements.py
bsowell: I guess this is kind of what I was getting at with my abstraction question above. I don't feel that strongly about it at this point, but it does seem like it would be nice to not have to implement a separate `GreedyElementMergerThatConcatenatesBinary` and `GreedyElementMergerThatReplacesBinary`, etc. Guess it depends on how many points in the space we expect to be useful in practice.

sycamore/transforms/merge_elements.py
bsowell: Bounding boxes already overlap in some cases, so not too worried about that.

sycamore/transforms/merge_elements.py
HenryL27: I considered exposing a `should_merge` and a `merge_two` interface kind of deal, and then have built in logic that does the reduce operation

sycamore/transforms/merge_elements.py
HenryL27: Just following what Roland did (there should be some kind of separation in case the text_representations are already stripped - we don't want to end up with ".... sword" + "fish ...." = ".... swordfish ...."

sycamore/transforms/merge_elements.py
HenryL27: might also be worth providing a `preprocess_elements` and `postprocess_elements` interface?

sycamore/transforms/merge_elements.py
HenryL27: ok, added it to the dataclass and now remove it at the end (maybe we should keep them around idk?)

bsowell (approved):

Issue #111, opened 2023-10-16
Better pipeline naming


HenryL27:
**Is your feature request related to a problem? Please describe.**
When I run a sycamore script I get a bunch of logs from Ray describing the pipeline that look like
```
(ReadBinary->Map(BinaryScan._to_document)->Map(ray_callable)->Map(ray_callable)->Map(ray_callable)->FlatMap(ray_callable)->MapBatches(ray_callable)->Write pid=37456)
```
I don't know what all these `ray_callable` map functions are, but I'd like to.

**Describe the solution you'd like**
The `ray_callable` name comes from a wrapper function that gets passed to ray in map around whatever transform is being executed. Can we do some python shenanigans and set the name of the function dynamically? (e.g. `ray_callable.__name__ = "blah"`)

**Describe alternatives you've considered**

**Additional context**

PR #116, opened 2023-10-17
Inject function name while creating Ray function

pparmar30:
Resolves https://github.com/aryn-ai/sycamore/issues/111

### DIFF
--- a/sycamore/transforms/map.py
+++ b/sycamore/transforms/map.py
@@ -8,7 +8,16 @@ from sycamore.data import Document
 from sycamore.plan_nodes import Node, UnaryNode
 
 
+def rename(new_function_name: str):
+    def decorator(f):
+        f.__name__ = new_function_name
+        return f
+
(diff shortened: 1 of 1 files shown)


sycamore/transforms/map.py
```diff
@@ -8,7 +8,16 @@
...
+def rename(new_function_name: str):
+    def decorator(f):
+        f.__name__ = new_function_name
+        return f
+
+    return decorator
+
+
 def generate_map_function(f: Callable[[Document], Document]) -> Callable[[dict[str, Any]], dict[str, Any]]:
+    @rename(f.__name__)
```
bsowell, line 20: What is f.__name__ if you pass in an anonymous function or like a lambda?

pparmar30: A lambda function will show up as `<lambda>`

bsowell (approved):

Issue #113, opened 2023-10-16
Add  support for JSON documents


pparmar30:
We can likely add very basic support for JSON documents via Unstructured.

PR #117, opened 2023-10-17
Add basic support for JSON

pparmar30:
Resolves https://github.com/aryn-ai/sycamore/issues/113

### DIFF
--- a/sycamore/scans/file_scan.py
+++ b/sycamore/scans/file_scan.py
@@ -2,9 +2,10 @@ import json
 from abc import ABC, abstractmethod
 import boto3
 from typing import Any, Optional, Union, Tuple
+import uuid
 
 from pyarrow.filesystem import FileSystem
-from ray.data import Dataset, read_binary_files, read_json
+from ray.data import Dataset, read_binary_files
 from ray.data.datasource import FileExtensionFilter
 
 from sycamore.data import Document
@@ -12,8 +13,6 @@ from sycamore.plan_nodes import Scan
 
 
 def _set_id(doc: dict[str, Any]) -> dict[str, Any]:
-    import uuid
-
     doc["doc_id"] = str(uuid.uuid1())
     return doc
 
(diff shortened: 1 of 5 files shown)


sycamore/scans/file_scan.py
```diff
@@ -133,14 +138,82 @@ def format(self):
...
+        json_dict = json.loads(json_str)
+
+        if self._document_body_field is not None:
+            body = json_dict.pop(self._document_body_field, None)
+        else:
+            body = json_str
+
+        if body is not None:
+            document.text_representation = body
+            document.binary_representation = body.encode("utf-8")
```
bohou-aryn, line 179: r we able to save this encoding when _document_body_field is None?

sycamore/scans/file_scan.py
```diff
@@ -133,14 +138,82 @@ def format(self):
...
+        document = Document()
+
+        document.doc_id = str(uuid.uuid1())
+        document.type = "json"
+
+        json_str = dict["bytes"].decode("utf-8")
+        json_dict = json.loads(json_str)
+
+        if self._document_body_field is not None:
+            body = json_dict.pop(self._document_body_field, None)
```
bohou-aryn, line 173: here default should be none or json_str?

bsowell: This is the case where the caller specifies a field to use and then it's not present. I think `None` is probably reasonable in that case, though open to suggestions if you feel differently.

sycamore/scans/file_scan.py
bsowell: Not quite sure I understand the suggestion. Are you trying to avoid decoding and then encoding? I guess if _document_body_field is None then we could just use dict["bytes"]?

bohou-aryn: yes, quite minor, just directly use dict["bytes"] when _document_body_field.

bohou-aryn (approved):

PR #118, opened 2023-10-18
Add support for OpenAI embeddings.

bsowell:
This change also updates the type signature for embedding to avoid an unnecessary nested list and convert from ndarray.

### DIFF
--- a/sycamore/transforms/embed.py
+++ b/sycamore/transforms/embed.py
@@ -1,14 +1,20 @@
 import math
+import os
 from abc import ABC, abstractmethod
+from enum import Enum
 from typing import Optional
+
+import openai
 
 import ray
 from ray.data import ActorPoolStrategy, Dataset
 from sentence_transformers import SentenceTransformer
 
+from tenacity import retry, stop_after_attempt, wait_random, retry_if_exception_type
+
 from sycamore.data import Document
 from sycamore.plan_nodes import Node, Transform
-from sycamore.utils import generate_map_batch_function, generate_map_batch_class_from_callable
+from sycamore.utils import batched, generate_map_batch_function, generate_map_batch_class_from_callable
 
 
 class Embedder(ABC):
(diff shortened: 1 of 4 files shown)


sycamore/transforms/embed.py
```diff
@@ -85,7 +92,69 @@ def generate_embeddings(self, doc_batch: list[Document]) -> list[Document]:
...
+        batch_size: The Ray batch size.
+        model_batch_size: The number of documents to send in a single OpenAI request.
+    """
+
+    def __init__(
+        self,
+        model_name: str = OpenAIEmbeddingModels.TEXT_EMBEDDING_ADA_002.value,
+        batch_size: Optional[int] = None,
+        model_batch_size: int = 100,
+        api_key: Optional[str] = None,
```
bohou-aryn, line 118: could we explicitly set device it uses, otherwise, it might set to cuda if system has gpu available, or you could change the logic for setting device in base class.

sycamore/transforms/embed.py
```diff
@@ -85,7 +92,69 @@ def generate_embeddings(self, doc_batch: list[Document]) -> list[Document]:
...
+    def _openai_embeddings(self, text_to_embed: list[str]) -> list:
+        return openai.Embedding.create(input=text_to_embed, engine=self.model_name).data
+
+    def generate_embeddings(self, doc_batch: list[Document]) -> list[Document]:
+        # TODO: Add some input validation here.
+        # The OpenAI docs are quite vague on acceptable values for model_batch_size.
+
+        for batch in batched(doc_batch, self.model_batch_size):
+            text_to_embed = [
+                doc.text_representation.replace("\n", " ") for doc in batch if doc.text_representation is not None
```
bohou-aryn, line 149: why we need "\n" get replaced?

sycamore/transforms/embed.py
bsowell: Doesn't the super call on line 120 explicitly set it to CPU? That was my intention anyway.

sycamore/transforms/embed.py
bsowell: It was recommended in the OpenAI libraries (https://github.com/openai/openai-python/blob/284c1799070c723c6a553337134148a7ab088dd8/openai/embeddings_utils.py#L20).

sycamore/transforms/embed.py
```diff
@@ -85,7 +92,69 @@ def generate_embeddings(self, doc_batch: list[Document]) -> list[Document]:
...
+
+        assert api_key is not None, (
+            "You must provide an API key to "
+            "use the LLM. Either pass it in "
+            "the constructor or set the "
+            "OPENAI_API_KEY environment "
+            "variable."
+        )
+        self.api_key = api_key
+        self.model_name = model_name
```
pparmar30, line 133: Does it make sense to re-use the OpenAI class here? [Link](https://github.com/aryn-ai/sycamore/blob/main/sycamore/llms/llms.py#L31)

bsowell: I wasn't sure. On one hand, there is some common configuration, but on the other hand embedding models are different and have a different signature. We could have the LLM class have an embed method, but then we would need some options for LLMs that don't support embeddings, etc. I'm open to suggestions, but it sort of seemed like using the OpenAI LLM class would add more code than it saves.

pparmar30 (approved):

PR #124, opened 2023-10-21
Return self from DocSet operations.  No need to waste DocSets.

alexaryn:
This will allow code like this:
```
ds=context.read_binary(...)
ds.partition(...)
ds.merge(...)
ds.explode(...)
ds.embed(...)
ds.write.opensearch(...)
```

### DIFF
--- a/sycamore/docset.py
+++ b/sycamore/docset.py
@@ -163,7 +163,8 @@ class DocSet:
         """
         from sycamore.transforms import Limit
 
-        return DocSet(self.context, Limit(self.plan, limit))
+        self.plan = Limit(self.plan, limit)
+        return self
 
     def partition(
         self, partitioner: Partitioner, table_extractor: Optional[TableExtractor] = None, **kwargs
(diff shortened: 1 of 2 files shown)


sycamore/docset.py
bohou-aryn, line 1: we do need to keep the docset plan tree read only for history, right?

alexaryn: The plan still is, right?  I'm open to discussion on how we want the code to work, but the existing behavior struck me as surprising and I couldn't see a technical reason in the current usage.  I wasn't so sure if there was a future vision I didn't get.

bohou-aryn: I think this immutability is pretty common in database execution layer. A lot of users just want do the following
ds1 = ds.partition()
ds2 = ds1.explode()
ds1.show()

we lose the track/ref to the plan in different stages.

alexaryn:
It seems people like getting a new DocSet back from each method.  Closing.

Issue #130, opened 2023-10-25
modify the text that gets embedded


HenryL27:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to specify/modify what text gets embedded. Some models require a prefix (e.g. BAAI/bge-*), and often it would be nice to be able to add some extra information into the embedding. For instance, throwing in a string that says something like "This is an excerpt of <properties.title>".

**Describe the solution you'd like**
I want to optionally pass a function (Document -> str) that will get called on all documents to generate the string to embed in the embed transform. Default should be `doc -> doc.text_representation` or some such.

**Describe alternatives you've considered**
Could also just modify the text representation with normal transforms but I'd rather not load th modified representations into my search engine, so I'd need a way to roll back the modification. 

**Additional context**

PR #132, opened 2023-10-25
Provide a way to modify the text to embed

pparmar30:
Resolves https://github.com/aryn-ai/sycamore/issues/130

### DIFF
--- a/sycamore/transforms/embed.py
+++ b/sycamore/transforms/embed.py
@@ -102,8 +102,12 @@ class SentenceTransformerEmbedder(Embedder):
 
         text_batch = [self.pre_process_document(doc) for doc in doc_batch if doc.text_representation is not None]
         embeddings = self._transformer.encode(text_batch, batch_size=self.model_batch_size, device=self.device)
-        for doc, embedding in zip(doc_batch, embeddings):
-            doc.embedding = embedding.tolist()
+
+        i = 0
+        for doc in doc_batch:
+            if doc.text_representation is not None:
+                doc.embedding = embeddings[i].tolist()
+                i += 1
 
         return doc_batch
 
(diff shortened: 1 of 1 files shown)


HenryL27 (commented):
LGTM, Thanks!
I noticed that bedrock (unbatched) and ST embedders handle null `text_representation`s differently. Can we follow the bedrock logic for sentence transformers? (if no text representation, just skip) Not sure how that would integrate w the preprocess fn.

sycamore/transforms/embed.py
```diff
@@ -19,17 +19,23 @@
 from sycamore.utils import batched, generate_map_batch_function, generate_map_batch_class_from_callable
 
 
+def _pre_process_document(document: Document) -> str:
+    return document.text_representation if document.text_representation is not None else ""
```
bsowell, line 23: I would agree with what Henry said. Some of the embedding models will break on empty strings (which is why I did it differently for OpenAI/Bedrock). Maybe better to skip empty passages.

bsowell: Maybe could do something as simple as if you return null from the preprocess function then skip embedding the element.

sycamore/transforms/embed.py
```diff
@@ -90,7 +96,7 @@ def generate_embeddings(self, doc_batch: list[Document]) -> list[Document]:
 
         assert self._transformer is not None
 
-        text_batch = ["" if doc.text_representation is None else doc.text_representation for doc in doc_batch]
+        text_batch = [self.pre_process_document(doc) for doc in doc_batch if doc.text_representation is not None]
         embeddings = self._transformer.encode(text_batch, batch_size=self.model_batch_size, device=self.device)
         for doc, embedding in zip(doc_batch, embeddings):
```
HenryL27, line 101: well now doc_batch and embeddings could be different lengths bc maybe a doc got take out of the middle, so won't that cause potential errors?

pparmar30: good catch! Updated

HenryL27 (approved):

Issue #129, opened 2023-10-25
Use manifest as single source of truth for paths


HenryL27:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to change my manifest file and have that change what documents get ingested

**Describe the solution you'd like**
Here's the hack I threw together
```
class PathProvider(JsonManifestMetadataProvider):
    def get_paths(self):
        return list(self._path_to_metadata_map.keys())


class ManifestReader(DocSetReader):
    def binary(
        self,
        binary_format,
        parallelism,
        filesystem,
        metadata_provider: PathProvider,
        **resource_args
    ):
        paths = metadata_provider.get_paths()
        return super().binary(
            paths=paths,
            binary_format=binary_format,
            parallelism=parallelism,
            filesystem=filesystem,
            metadata_provider=metadata_provider,
            **resource_args
        )
```

**Describe alternatives you've considered**
The end goal here is to dedupe multilingual overlapping documents. Some kind of docset-level dedupe transform/operation would also suffice though it would be a lot more compute-heavy than simply not bringing in files I don't want.

It's probably also possible to bake this into the standard Reader methods without having to subclass it, this was just a quick thing I did.

**Additional context**

PR #133, opened 2023-10-26
Add manifest reader

pparmar30:
resolves https://github.com/aryn-ai/sycamore/issues/129

### DIFF
--- a/sycamore/reader.py
+++ b/sycamore/reader.py
@@ -33,6 +33,7 @@ class DocSetReader:
         )
         return DocSet(self._context, scan)
 
+    # TODO: Support including the metadata attributes in the manifest file directly
     def manifest(
         self,
         metadata_provider: FileMetadataProvider,
--- a/sycamore/reader.py
+++ b/sycamore/reader.py
@@ -33,6 +33,25 @@ class DocSetReader:
         )
         return DocSet(self._context, scan)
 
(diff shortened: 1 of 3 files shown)


sycamore/reader.py
```diff
@@ -33,6 +33,25 @@ def binary(
...
+        scan = BinaryScan(
+            paths,
+            binary_format=binary_format,
+            parallelism=parallelism,
+            filesystem=filesystem,
+            metadata_provider=metadata_provider,
+            **resource_args
+        )
+        return DocSet(self._context, scan)
+
```
HenryL27, line 54: would this work for html/non-binary?

pparmar30: It will work with html as we are still using Binary scan for HTMLs. We can add support for non-binaries as needed.

sycamore/reader.py
```diff
@@ -33,6 +33,25 @@ def binary(
         )
         return DocSet(self._context, scan)
 
+    def manifest(
+        self,
+        metadata_provider: FileMetadataProvider,
```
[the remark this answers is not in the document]
pparmar30: Added as a TODO

HenryL27 (approved):
lgtm, ty!

bsowell (approved):

PR #137, opened 2023-11-01
Parameter to help with debugging bounding box issues.

bsowell:
This is a minor thing, but I've found myself needing it twice in the past couple of weeks, so it seems worth adding. If you initialize the partitioner as `UnstructuredPdfPartitioner(retain_coordinates=True)`, then the resulting elements will have a `coordinates` property with the original Unstructured coordinates in addition to the bbox property that we compute.


sycamore/transforms/partition.py
```diff
@@ -199,7 +203,10 @@ def to_element(dict: dict[str, Any]) -> Element:
         properties = element.properties
         properties.update(dict.pop("metadata"))
         properties.update(dict)
-        coordinates = properties.pop("coordinates")
+        coordinates = properties.get("coordinates")
```
alexaryn, line 206: Why get() and pop() in the common case?  Why not have an else and do either/or?

PR #138, opened 2023-11-01
Re-lock poetry config to fix a warning around the 'executing' package.

eric-anderson:
Other changes to the lock file automatically happen when running poetry lock --no-update

Error was:
Warning: The locked version 2.0.0 for executing is a yanked version. Reason for being yanked: Released 2.0.1 which is equivalent but added 'python_requires = >=3.5' so that pip install with Python 2 uses the previous version 1.2.0.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1118,13 +1118,13 @@ test = ["pytest (>=6)"]
 
 [[package]]
 name = "executing"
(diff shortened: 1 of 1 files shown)


bsowell (approved):
Took me a few reads to understand that "executing" is the name of the dependency :).

PR #142, opened 2023-11-06
Merge in crawler

eric-anderson:
Merge the private crawler repo into the public sycamore repository.

Three main commits:

1. Prepare crawler repo by moving all files into a subdirectory
2. Create the merge of unrelated histories between the two repos
3. Fix the resulting merge so that it works; including some bugfixes that were missed in the crawler repo.


crawler/http/crawler/settings.py
```diff
@@ -0,0 +1,33 @@
...
+#     https://docs.scrapy.org/en/latest/topics/spider-middleware.html
+
+BOT_NAME = "crawler"
+
+SPIDER_MODULES = ["crawler.spiders"]
+NEWSPIDER_MODULE = "crawler.spiders"
+
+
+# Crawl responsibly by identifying yourself (and your website) on the user-agent
+USER_AGENT = "sycamore crawler (+https://github.com/aryn-ai/crawler)"
```
bsowell, line 17: Doesn't matter, but should this point to sycamore repo?

crawler/s3/tests/unit/test_s3_crawler.py
```diff
@@ -0,0 +1,101 @@
...
+from pathlib import Path
+from typing import Any
+
+import botocore.paginate
+
+from crawler.s3.crawler import s3_crawler
+from crawler.s3.crawler.s3_crawler import S3Crawler
+
+
+# TODO : Parth - investigate doing an integration test similar to the http one. Maybe using minio or localstack
```
bsowell, line 11: (not for the merge, but for reference we have a bucket for testing and the we can get the aws keys in github, so I think we can just have an integration test against actual s3.)

bsowell (approved):

eric-anderson:
Nuke and retry again, mypy on desktop behaves differently. Trying datetime.timezone.utc instead of datetime.UTC

PR #143, opened 2023-11-06
Merge in crawler

eric-anderson:
Merge the private crawler repo into the public sycamore repository.

Three commits for the merge:

1. Prepare crawler repo by moving all files into a subdirectory
2. Create the merge of unrelated histories between the two repos
3. Fix the resulting merge so that it works; including some bugfixes that were missed in the crawler repo.


crawler/http/crawler/settings.py
```diff
@@ -0,0 +1,33 @@
...
+#     https://docs.scrapy.org/en/latest/topics/spider-middleware.html
+
+BOT_NAME = "crawler"
+
+SPIDER_MODULES = ["crawler.spiders"]
+NEWSPIDER_MODULE = "crawler.spiders"
+
+
+# Crawl responsibly by identifying yourself (and your website) on the user-agent
+USER_AGENT = "sycamore crawler (+https://github.com/aryn-ai/crawler)"
```
bsowell, line 17: Not a big deal, but might be better to point to sycamore repo, since the separate crawler repo presumably won't exist/won't be public.

bsowell (approved):

crawler/http/crawler/settings.py
eric-anderson: Done.

pparmar30 (approved):

mashah (approved):
Looks good to me.

PR #144, opened 2023-11-07
Add workflows to publish docker images

pparmar30:


.github/workflows/http_crawler_docker_release.yml
```diff
@@ -0,0 +1,21 @@
...
+        uses: actions/checkout@v4
+
+      - name: Login to Docker Hub
+        run: docker login -u ${{ secrets.ARYN_DOCKERHUB_USERNAME }} -p ${{ secrets.ARYN_DOCKERHUB_PASSWORD }}
+
+      - name: Create a new docker builder
+        run: docker buildx create --name arynbuilder --use
+
+      - name: Build Docker image
+        run: docker buildx build -t  ${{ secrets.ARYN_DOCKERHUB_USERNAME }}/${{ secrets.ARYN_DOCKERHUB_HTTP_CRAWLER }}:latest -f crawler/http/Dockerfile --platform linux/amd64,linux/arm64 --push .
```
eric-anderson, line 21: nit:add newline to end of file.

.github/workflows/http_crawler_docker_release.yml
```diff
@@ -0,0 +1,21 @@
+name: HTTP Crawler Docker Image CI
+
+on:
+  push:
+      branches: [ "main" ]
```
eric-anderson, line 5: I don't think we want to push on every push to main. I think we should either only run on version tags (v.*), or restrict the action to run only if the crawler, poetry.lock, pyproject.toml, or README.md files have changed.

So https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags
or https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

Same comment for other build.

I'm slightly leaning toward running on a version because that what I expect we'll do for sycamore.

eric-anderson (changes requested):

PR #145, opened 2023-11-07
Add docker compose yaml files that run sycamore + crawler + arynai/opensearch + demoui

eric-anderson:
compose.yaml is the main file.
reset.yaml lets you reset persistent state back to empty.
sort-all.yaml downloads all of the sort benchmark files.

### DIFF
--- a/docker_compose/README.md
+++ b/docker_compose/README.md
@@ -1,0 +1,9 @@
+# Docker compose for sycamore
+
+Runs the entire sycamore stack + aryn demo UI.
+
+See compose.yaml for details and extensions, but in short:
+
+1. cd .../docker_compose
+1. docker compose up
+1. visit http://localhost:3000/
--- a/docker_compose/README.md
+++ b/docker_compose/README.md
@@ -4,6 +4,6 @@ Runs the entire sycamore stack + aryn demo UI.
 
 See compose.yaml for details and extensions, but in short:
 
-1. cd .../docker_compose
+1. cd path/to/sycamore/docker_compose
 1. docker compose up
 1. visit http://localhost:3000/
--- a/docker_compose/compose.yaml
+++ b/docker_compose/compose.yaml
@@ -2,12 +2,12 @@
 # % docker compose up
 #
 # What you should see:
-#   1. lots of waiting while containers are downloaded (or nothing if cached)
+#   1. wait while containers are downloaded (speed depends on your network), or nothing once cached
 #   2a. crawler should run and download 1 sample pdf
 #   2b. demo ui starting (in parallel with 2a)
-#   3. lots of logging as opensearch get started eventually 'Waiting for opensearch to exit.'
-#   4. sycamore container downloading pytorch files then running to import
-#   5. UI ready to go on port localhost:3000
+#   3. lots of logging as opensearch starts. Eventually 'Waiting for opensearch to exit.'
+#   4. sycamore container downloading pytorch files then running ray to import
+#   5. UI ready to go on port localhost:3000 (it's accessible before this but won't work)
 #
 # Troubleshooting/Extensions:
 #   1. Something is messed up, I just want to reset:
--- a/docker_compose/compose.yaml
+++ b/docker_compose/compose.yaml
@@ -1,0 +1,52 @@
+# To start, run:
+# % docker compose up
+#
+# What you should see:
+#   1. lots of waiting while containers are downloaded (or nothing if cached)
+#   2a. crawler should run and download 1 sample pdf
+#   2b. demo ui starting (in parallel with 2a)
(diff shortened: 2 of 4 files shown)


docker_compose/README.md
```diff
@@ -0,0 +1,9 @@
+# Docker compose for sycamore
+
+Runs the entire sycamore stack + aryn demo UI.
+
+See compose.yaml for details and extensions, but in short:
+
+1. cd .../docker_compose
```
austintlee, line 7: Three dots?  Is that a linux thing?

docker_compose/compose.yaml
```diff
@@ -0,0 +1,52 @@
+# To start, run:
+# % docker compose up
+#
+# What you should see:
+#   1. lots of waiting while containers are downloaded (or nothing if cached)
```
austintlee, line 5: A few minutes or go get a cup of coffee kind of waiting?  Can we say a few minutes?

docker_compose/compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+#   2a. crawler should run and download 1 sample pdf
+#   2b. demo ui starting (in parallel with 2a)
+#   3. lots of logging as opensearch get started eventually 'Waiting for opensearch to exit.'
```
austintlee, line 8: I don't think logs go to stdout by default?

docker_compose/compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+#   4. sycamore container downloading pytorch files then running to import
+#   5. UI ready to go on port localhost:3000
+#
+# Troubleshooting/Extensions:
+#   1. Something is messed up, I just want to reset:
+#      Follow the instructions in reset.yaml
+#   2. I want more sort benchmark files:
```
austintlee, line 15: So, this docker compose is only going to be used for retrieving the sort benchmark data?  Can we allow users to provide their own url?

docker_compose/reset.yaml
```diff
@@ -0,0 +1,17 @@
...
+version: '3'
+
+services:
+  reset:
+    image: ubuntu
+    volumes:
+      - opensearch_data:/tmp/opensearch_data
+      - crawl_data:/tmp/crawl_data
+    command: /bin/sh -c 'echo size-before && du -s /tmp/* && rm -rf /tmp/opensearch_data/* /tmp/crawl_data/* && echo size-after && du -s /tmp/*'
+
```
austintlee, line 14: I'm wondering if we should just provide a script (reset.sh) that does this instead of downloading and spinning up a container just to execute these commands.

pparmar30: We should probably do something similar to load all of sort_benchmark data as well.

docker_compose/README.md
eric-anderson: It was to indicate whatever is your path to the docker_compose directory.
Changed to cd path/to/sycamore/docker_compose

docker_compose/compose.yaml
eric-anderson: It depends on how fast your network is. When Mehul was getting 1MB/s at the conference it would be about 1-1.5 hours (4-5GB of downloads).  Updated comment.

docker_compose/compose.yaml
eric-anderson: I was going to figure out how to do additional things via compose after we got V1 out. Added to the V2 task list.

(You need to specify -a domain=... and -a url=... to the container run.

docker_compose/reset.yaml
eric-anderson: I don't know of a way to fiddle with containers except by being within a container.  The ubuntu container is tiny (30MB) so it doesn't seem worth trying to be smarter.

You can load all of the sort_benchmark data by running 'docker compose -f sort-all.yaml up' as documented in sort-all.yaml

eric-anderson: I guess I could add sort-all.sh reset.sh and compose.sh scripts that run the command at the top. Would that be valuable?

docker_compose/compose.yaml
eric-anderson: If you run the command as shown they did for Alex, Parth, and me.
If you run it with -d (detach) they don't.

docker_compose/reset.yaml
austintlee: I guess this guy had the same question - https://github.com/docker/compose/issues/4031.

pparmar30 (approved):

austintlee (approved):

docker_compose/reset.yaml
eric-anderson: It's not quite the same, I think they wanted to share a directory between the host and the container. I looked into doing that for some of the persistent state, but what I saw said it has issues on MacOS so I went another way.

PR #146, opened 2023-11-07
Add the scripts to dockerize opensearch in a way that works with the other sycamore components

eric-anderson:
In particular, the aryn-opensearch.sh script sets up all of the opensearch configuration necessary to let the Aryn demo UI work. It also coordinates with the importer to make sure that the importer only starts when opensearch is ready.

### DIFF
--- a/opensearch/Dockerfile
+++ b/opensearch/Dockerfile
@@ -1,5 +1,5 @@
 ARG TAG=2.11.0
-FROM opensearchstaging/opensearch:$TAG
+FROM opensearchproject/opensearch:$TAG
 USER 0
 RUN yum install -y jq && yum clean all
 USER 1000
--- a/opensearch/Dockerfile
+++ b/opensearch/Dockerfile
@@ -1,0 +1,10 @@
+ARG TAG=2.11.0
+FROM opensearchstaging/opensearch:$TAG
+USER 0
+RUN yum install -y jq && yum clean all
+USER 1000
+COPY opensearch.yml /usr/share/opensearch/config/
+COPY aryn-opensearch.sh .
+ENV discovery.type=single-node
+ENV DISABLE_SECURITY_PLUGIN=true
+CMD ./aryn-opensearch.sh
--- a/opensearch/aryn-opensearch.sh
+++ b/opensearch/aryn-opensearch.sh
@@ -1,0 +1,427 @@
+#!/bin/bash
+# TODO: eric - detect low disk space and error out.
+# on macos you fix it in docker desktop > settings > resources > scroll down > virtual disk limit
+main() {
+    local http_status_port=43477 # from rand
+    local http_server_dir=/tmp/http_server
+    local http_status_file="${http_server_dir}/statusz"
+    [[ ! -f "${http_status_file}" ]] || rm -f "${http_status_file}"
+    
+    BASE_URL=http://localhost:9200
+    ARYN_STATUSDIR=/usr/share/opensearch/data/aryn_status
+    mkdir -p "${ARYN_STATUSDIR}"
+
+    set -e
+    # TODO: eric - figure out how we can get the Aryn logs during startup
+    # and switchover to the opensearch ones once we've got everything running
+
+    LOG_BASE="${ARYN_STATUSDIR}/opensearch.log"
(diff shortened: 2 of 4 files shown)


opensearch/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG TAG=2.11.0
+FROM opensearchstaging/opensearch:$TAG
```
austintlee, line 2: I don't think you want to use 'opensearchstaging'.  I think this is only used for release candidates.  The correct namespace is 'opensearchproject'.

opensearch/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG TAG=2.11.0
```
austintlee, line 1: I think we can just go with 'latest'?

opensearch/aryn-opensearch.sh
```diff
@@ -0,0 +1,427 @@
+#!/bin/bash
+# TODO: eric - detect low disk space and error out.
```
austintlee, line 2: remove 'eric'.  do you want to open an issue and assign it to yourself?

opensearch/aryn-opensearch.sh
```diff
@@ -0,0 +1,427 @@
...
+    mkdir -p "${ARYN_STATUSDIR}"
+
+    set -e
+    # TODO: eric - figure out how we can get the Aryn logs during startup
+    # and switchover to the opensearch ones once we've got everything running
+
+    LOG_BASE="${ARYN_STATUSDIR}/opensearch.log"
+    if opensearch_up; then
+        echo "OpenSearch appears to already be running, not starting it again"
+    else
```
austintlee, line 21: not sure if this would useful, but docker supports healthchecks - https://docs.docker.com/engine/reference/builder/#healthcheck

opensearch/aryn-opensearch.sh
```diff
@@ -0,0 +1,427 @@
...
+  "phase_results_processors": [
+    {
+      "normalization-processor": {
+        "normalization": {
+          "technique": "min_max"
+        },
+        "combination": {
+          "technique": "arithmetic_mean",
+          "parameters": {
+            "weights": [0.111, 0.889]
```
austintlee, line 378: @HenryL27 is this what we use everywhere?

opensearch/Dockerfile
eric-anderson: Done.

opensearch/aryn-opensearch.sh
eric-anderson: The eric was the reference for the bug.
Replaced with TODO: http://...

will do all the other ones also.

opensearch/aryn-opensearch.sh
HenryL27: this is. Approximation of 1:8 (what that blog post found was best) bc they decided weights must sum to one. I've tried a couple of other weightings, but this has performed best on our benchmark so far.

opensearch/aryn-opensearch.sh
eric-anderson: I can't find any way to check the health from another container, so I don't think it's useful to the automation scripts.  I searched for ~5 min, resolve the conversation if you can't find a way to support cross-container healthchecks.

In this particular case, this part of the code supports running opensearch in the container and then logging into the container and running the script again to redo the configuration. Starting opensearch twice wouldn't work well in that case.

opensearch/Dockerfile
austintlee: But you have to keep updating this file.  We only care about >= 2.11, not == 2.11.  People can modify this locally so no big deal.

opensearch/aryn-opensearch.sh
austintlee: I was playing around with this the other day.

https://gist.github.com/austintlee/84d24b040992127512e60877d48b8f54

austintlee (approved):

opensearch/aryn-opensearch.sh
eric-anderson: Per discussion, this would let us sequence containers with depends_on, but we don't really want to do that.

PR #147, opened 2023-11-08
Speed up 'poetry lock'.

alexaryn:


pyproject.toml
```diff
@@ -16,18 +16,20 @@ packages = [{include = "sycamore"}]
...
 [tool.poetry.group.sycamore_library.dependencies]
-opensearch-py = "^2.2.0"
+opensearch-py = "^2.3.1"
 pandas = "2.1.1"
 pdf2image = "^1.16.3"
 ray = { extras = ["default"], version = "^2.7.0" }
 sentence-transformers = "2.2.2"
 pyarrow = "13.0.0"
-unstructured = "0.10.19"
+unstructured = "0.10.20"
```
bsowell, line 25: Should we allow newer minor versions (e.g. "^0.10.20")? Not sure myself, tbh. On one hand, would be nice to not have to manually upgrade. On the other hand, we have not had a tremendous amount of luck with our dependencies being backwards compatible. I'm fine with this.

pyproject.toml
```diff
@@ -16,18 +16,20 @@ packages = [{include = "sycamore"}]
...
+openai = "0.28.1"
```
bsowell, line 26: Thanks for doing this. I spent 2 minutes looking at the newest version and it seems that they changed a fair amount. I will add an issue to upgrade properly, but in the meantime this is good.

bsowell (approved):

pyproject.toml
alexaryn: Newer ones break stuff.  I tried.  We need to target a specific version and take the time to make it work.

pyproject.toml
alexaryn: We should upgrade OpenAI also, maybe at the same time.  And any other similar packages.

PR #161, opened 2023-11-09
Entity Extraction

mkyl:
This pull request adds two operators to Sycamore: `extract_schema` and `extract_properties`. 

The first operator works on the document level to extract a json-schema. The second operator is then able to extract that schema from a given snippet or document. This is appended to the properties for export to OpenSearch and looks something like this:

```
{'BenchmarkRun': {'name': 'TritonSort',
   'year': 2011,
   'recordsSorted': 100000000000,
   'sortTimeSeconds': 8274,
   'hardware': {'processor': '2 Quadcore processors',
    'memory': '24 GB memory',
    'storage': '16x500GB disks'},
   'participants': [{'name': 'Alex Rasmussen',
     'affiliation': 'University of California'},
    {'name': 'Michael Conley', 'affiliation': 'University of California'},
    {'name': 'George Porter', 'affiliation': 'University of California'},
    {'name': 'Amin Vahdat', 'affiliation': 'University of California'}]}}}
```


sycamore/transforms/schema.py
```diff
@@ -0,0 +1,180 @@
...
+        self._num_of_elements = num_of_elements
+        self._prompt_formatter = prompt_formatter
+
+    def extract_schema(self, document: Document) -> Document:
+        entities = self._handle_zero_shot_prompting(document)
+
+        try:
+            payload = entities["answer"]
+            pattern = r'```json([\s\S]*?)```'
+            match = re.match(pattern, payload).group(1)
```
baitsguy, line 68: you might want to look at https://www.npmjs.com/package/jsonrepair, I think it'll handle a few more formatting issues. If you don't see an issue that's fine

sycamore/docset.py
```diff
@@ -269,6 +270,19 @@ def extract_entity(self, entity_extractor: EntityExtractor, **kwargs) -> "DocSet
...
         entities = ExtractEntity(self.plan, entity_extractor=entity_extractor, **kwargs)
         return DocSet(self.context, entities)
+    
+    def extract_schema(self, schema_extractor: SchemaExtractor, **kwargs) -> "DocSet":
+        from sycamore.transforms import ExtractSchema
+
+        schema = ExtractSchema(self.plan, schema_extractor=schema_extractor)
+        return DocSet(self.context, schema)
+    
+    def extract_properties(self, property_extractor: PropertyExtractor, **kwargs) -> "DocSet":
```
bsowell, line 280: Is it worth having this separate from `extract_entity`? I get that they behave slightly differently, but from an external perspective they do almost the same thing. I wonder if we can unify the `PropertyExtractor` and `EntityExtractor`  classes?

sycamore/llms/llms.py
```diff
@@ -10,7 +10,7 @@
 
 class OpenAIModels(Enum):
     TEXT_DAVINCI = "text-davinci-003"
-    GPT_3_5_TURBO = "gpt-3.5-turbo"
+    GPT_3_5_TURBO = "gpt-3.5-turbo-1106"
```
bsowell, line 13: Thinking about how we want to handle model changes like this. Looks like OpenAI will eventually point gpt-3.5-turbo to gpt-3.5-turbo-1106, so I guess it's fine to upgrade like this, though maybe we should expose the different variants to try to limit surprising behavior changes? I don't think it's a huge deal for this PR, but as the number of models increases we will need to come up with a standard approach here.

sycamore/llms/prompts/default_prompts.py
```diff
@@ -73,3 +73,53 @@
     =========
     {{gen "summary"}}
     """
+
+SCHEMA_ZERO_SHOT_GUIDANCE_PROMPT_CHAT = """
```
bsowell, line 77: Does guidance help you here? We've been going back and forth on whether it's worth it. Curious what you think.

sycamore/transforms/schema.py
```diff
@@ -0,0 +1,180 @@
...
+            answer = json.loads(match)
+        except:
+            answer = entities["answer"]
```
bsowell, line 71: In this case `entities[answer]` is likely to be a string, right? Will that break something downstream? Just trying to understand what the error handling should look like.

sycamore/transforms/schema.py
```diff
@@ -0,0 +1,180 @@
...
+        properties.update({"_schema": answer, "_schema_class": self._entity_name})
+        document.properties = properties
+
+        return document
+
+    def _handle_zero_shot_prompting(self, document: Document) -> Any:
+        sub_elements = [document.elements[i] for i in range((min(self._num_of_elements, len(document.elements))))]
+
+        if self._llm.is_chat_mode:
+            prompt = SCHEMA_ZERO_SHOT_GUIDANCE_PROMPT_CHAT
```
bsowell, line 83: Do you see better results with chat mode?

sycamore/transforms/schema.py
```diff
@@ -0,0 +1,180 @@
...
+    def extract_properties(self, document: Document) -> Document:
+        entities = self._handle_zero_shot_prompting(document)
+
+        try:
+            payload = entities["answer"]
+            pattern = r'```json([\s\S]*?)```'
+            match = re.match(pattern, payload).group(1)
+            answer = json.loads(match)
+        except:
+            answer = entities["answer"]
```
bsowell, line 120: If this is the right way to deal with JSON llm response, let's extract it as a utility somewhere.

sycamore/transforms/schema.py
```diff
@@ -0,0 +1,180 @@
...
+class OpenAIPropertyExtractor(PropertyExtractor):
+    """
+    OpenAISchema uses one of OpenAI's language model (LLM) to extract property values once schema is established.
+    """
+
+    def __init__(
+        self,
+        #properties: list[str],
+        llm: LLM,
+        num_of_elements: int = 10,
```
bsowell, line 103: I know this is modeled after the entity extractor, but this `num_of_elements` always seemed a little clunky for things other than title/authors that are likely to be at the beginning of the doc. Any thoughts on a better interface? seems like users might want an easy way to run this over all elements either individually or in batches sized based on the context window.

sycamore/transforms/schema.py
```diff
@@ -0,0 +1,180 @@
+from abc import ABC, abstractmethod
```
bsowell, line 1: Super minor, but can we call this file something like `extract_schema.py` rather than just `schema.py`. Schema is such an overloaded term it could mean any number of things.

mkyl:
To do:
- [ ] Tests
- [ ] Documentation

mkyl:
(Including test on batch of documents)

sycamore/docset.py
mkyl: Sorry for the delay on getting back to this. My thinking here is that there a setting in which the user provides the schema themselves (so `extract_schema` is **not** needed) but still want us to extract the actual entities (so `extract_properties` is needed). The user may even want to modify the schema produced by us before using it to extract entities.

sycamore/llms/llms.py
mkyl: This will auto-upgrade in eight days, so I think it may make sense to test the performance on this branch and then force the change. But also happy to reverse this patch and deal with this on a separate branch.

sycamore/llms/prompts/default_prompts.py
mkyl: I'm not using any guidance-specific features at the moment.

sycamore/transforms/schema.py
mkyl: For schema extraction, a small error in the generation should be fine. Because the subsequent call to the LLM to extract the entity using the json-like schema should recover. If we want to be conservative we may want to decline to populate a schema here.

sycamore/transforms/schema.py
mkyl: The system instruction to only generate JSON is very helpful, yes. (OpenAI actually requires it if you want to enable their forthcoming "json mode")

sycamore/transforms/schema.py
mkyl: done 👍

sycamore/transforms/schema.py
mkyl: Done 👍

sycamore/transforms/schema.py
mkyl: At the moment not having issues with malformed JSON but we can revisit if testing reveals problems with this.

mkyl:
Cleaned up the notebooks so this is not a +16,000 patch.

sycamore/transforms/schema.py
mkyl: I agree many users would want a rolling buffer. I also think there might be a clearer way to express the semantics of bulk schema detection rather than have two methods `extract_schema` and `extract_batch_schema`. Let's discuss on Monday.

mkyl:
The output of `entities-ntsb.ipynb` should resemble:
```json
{"location": "ARDMORE, OK",
 "dateAndTime": "1995-03-02T08:15:00Z", 
"aircraft": "CESSNA 208B", 
"definingEvent": "STRUCTURAL ICE ACCUMULATED ON THE AIRPLANE AT A RAPID RATE WHILE THE PILOT WAS EXECUTING A PROCEDURE TURN DURING AN NDB APPROACH TO HIS DESTINATION AIRPORT. THE PILOT APPLIED MAXIMUM AVAILABLE POWER BUT WAS UNABLE TO ARREST THE RATE OF DESCENT. THE PILOT MAINTAINED 110 KNOTS DURING THE APPROACH AND AFTER BREAKING OUT INTO VISUAL CONDITIONS ELECTED TO PERFORM A FORCED LANDING IN AN OPEN FIELD 2 MILES SHORT OF THE DESTINATION AIRPORT. THE PILOT WAS UNABLE TO ARREST HIS DESCENT AND THE AIRPLANE IMPACTED THE GROUND IN A 5 DEGREE NOSE DOWN ATTITUDE. THE PILOT ALSO STATED THAT HIS FORWARD VISIBILITY WAS COMPLETELY OBSCURED BY THE ACCUMULATION OF CLEAR ICE ON THE WINDSHIELD. THE AIRPORT MANAGER INFORMED THE PILOT THAT FREEZING DRIZZLE WAS PRESENT AT THE AIRPORT WHEN THE PILOT CALLED THE UNICOM DURING THE INITIAL PORTION OF THE INSTRUMENT APPROACH.", 
"flightConductedUnder": "Part 135: Air Taxi & Commuter - Non-scheduled",
 "probableCause": "THE PILOT\'S CONTINUED FLIGHT INTO ADVERSE WEATHER CONDITIONS. FACTORS WERE THE ICING CONDITIONS PREVAILING AT THE DESTINATION AIRPORT, AND THE PILOT\'S INABILITY TO MAINTAIN VISUAL LOOKOUT DUE TO WINDSHIELD ICING.", 
"accidentNumber": "FTW95FA129"}
  ```

File: sycamore/utils/extract_json.py [hunk 10]
```diff
@@ -0,0 +1,11 @@
+import re
+import json
+
+
+def extract_json(payload):
+    pattern = r"```json([\s\S]*?)```"
```
bsowell, line 6: Is the fact that this is wrapped in ```json()``` specific to OpenAI? or is it coming from somewhere else (like guidance)? If we wanted to support a different llm at some point (say Anthropic), would we need a different implementation of this method? If it's OpenAI specific we should make that clear in the name/location.

sycamore/transforms/extract_schema.py
```diff
@@ -0,0 +1,216 @@
...
+        super().__init__(child, **resource_args)
+        self._schema_extractor = schema_extractor
+
+    def execute(self) -> "Dataset":
+        input_dataset = self.child().execute()
+        dataset = input_dataset.map(generate_map_function(self._schema_extractor.extract_schema))
+        return dataset
+
+
+class ExtractBatchSchema(Transform):
```
bsowell, line 170: I'm having a hard time really grokking the batch vs non-batch implementation. Maybe you can walk me through it when we talk.

bsowell (commented):
Couple of things I'd like to clarify, but otherwise looking good. It would be good to write some docs for this.

PR #163, opened 2023-11-10
Upgrade pyarrow version and minor version bumps to other deps.

bsowell:
Addresses https://github.com/advisories/GHSA-5wvp-7f3h-6wmm


bsowell:
Hmm. Something looks messed up here -- it's not able to install the repo in github actions, though it worked locally. Let me see if I can figure out what's going on.

PR #169, opened 2023-11-13
Change simple_ingest and s3_ingest to use GTE-small embedding model.

alexaryn:
The GTE-small embedding model is superior to MiniLM, yielding better relevance in retrieval.


eric-anderson:
Can you add a "why we are doing this" to the description?

baitsguy (approved):
Discussed that we need to update all our references to this model

PR #171, opened 2023-11-15
Merging/breaking elements based on heuristics including bbox

alexaryn:
There are a few pieces of code that manipulate/mark elements:

- sort_top_to_bottom - reorder elements by page_number and bbox
- mark_drop_tiny - remove text shorter than a minimum size
- mark_drop_header_footer - remove text at extreme top or bottom
- mark_break_page - break before each new page
- mark_break_column - break at transitions from 2 columns to full-width (and re-sort)
- mark_break_tokens - break to keep element size within limit

- mark_bbox_preset - is a shorthand for a preconfigured set of the above

MarkedMerger does the merging according to the various marks on elements.

This is the same behavior that was tested and did well in the relevance benchmark.

### DIFF
--- a/sycamore/transforms/mark_misc.py
+++ b/sycamore/transforms/mark_misc.py
@@ -1,0 +1,142 @@
+from ray.data import Dataset
+
+from sycamore.data import Document
+from sycamore.functions.tokenizer import Tokenizer
+from sycamore.plan_nodes import Node, Transform, SingleThreadUser, NonGPUUser
+from sycamore.transforms.map import generate_map_function
+
+# TODO:
+# - make breaks balanced in size
+# - maybe move token counting elsewhere to avoid duplicate work
+
+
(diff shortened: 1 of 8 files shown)


sycamore/transforms/mark_misc.py
```diff
@@ -0,0 +1,142 @@
...
+from sycamore.functions.tokenizer import Tokenizer
+from sycamore.plan_nodes import Node, Transform, SingleThreadUser, NonGPUUser
+from sycamore.transforms.map import generate_map_function
+
+# TODO:
+# - make breaks balanced in size
+# - maybe move token counting elsewhere to avoid duplicate work
+
+
+class MarkDropTiny(SingleThreadUser, NonGPUUser, Transform):
```
baitsguy, line 13: We'll probably want more tests for these once we've settled on the apis

baitsguy (approved):
Logic looks good, I think we'll want to add more tests later and see how the preset fits into the whole picture

PR #174, opened 2023-11-17
Move library dependencies back under [tool.poetry.dependencies]

bsowell:
This is necessary to ensure that the sycamore library remains installable with pip.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,8 +14,6 @@ packages = [{include = "sycamore"}]
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -43,9 +41,6 @@ furo = { version = "^2023.9.10", optional = true }
 myst-parser = { version = "^2.0.0", optional = true }
 sphinx = { version = "^7.2.5", optional = true }
 
-[tool.poetry.group.crawler_s3.dependencies]
```
eric-anderson, line 46: Leave this group in; I think it will let me install only crawler_s3 dependencies for those Dockerfiles.

eric-anderson (approved):

PR #180, opened 2023-11-20
Simplify running the http crawler.

eric-anderson:
You used to need to specify both a url and a domain, but we can infer the likely domain from the URL, so avoid needing to write it twice.

Also add support for a prefix check so people can download only a specific sub-tree.

### DIFF
--- a/crawler/http/crawler/settings.py
+++ b/crawler/http/crawler/settings.py
@@ -16,6 +16,9 @@ NEWSPIDER_MODULE = "crawler.spiders"
 # Crawl responsibly by identifying yourself (and your website) on the user-agent
 USER_AGENT = "sycamore crawler (+https://github.com/aryn-ai/sycamore)"
 
+# Crawl at about 1 QPS
+DOWNLOAD_DELAY = 1
+
 # Obey robots.txt rules
 ROBOTSTXT_OBEY = True
 
--- a/crawler/http/crawler/spiders/sycamore.py
+++ b/crawler/http/crawler/spiders/sycamore.py
@@ -4,7 +4,7 @@ import os
 import re
 import scrapy
 import time
-import urllib.parse
+import urllib
 
 
 class SycamoreSpider(scrapy.Spider):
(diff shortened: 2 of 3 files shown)


crawler/http/crawler/settings.py
```diff
@@ -16,6 +16,9 @@
 # Crawl responsibly by identifying yourself (and your website) on the user-agent
 USER_AGENT = "sycamore crawler (+https://github.com/aryn-ai/sycamore)"
 
+# Crawl at about 1 QPS
+DOWNLOAD_DELAY = 1
```
austintlee, line 20: Where is this used?

crawler/http/crawler/spiders/sycamore.py
```diff
@@ -4,24 +4,31 @@
...
             self.start_urls = [kwargs["url"]]
-        elif "domain" in kwargs or "url" in kwargs:
-            raise RuntimeError("Should have both -a domain=... and -a url=...")
+            if "prefix" in kwargs:
+                self.prefix = kwargs["prefix"]
+            if "domain" in kwargs:
+                self.allowed_domains = [kwargs["domain"]]
+            else:
+                u = urllib.parse.urlparse(kwargs["url"])
+                d = u.hostname.removeprefix("www.")
```
austintlee, line 26: Does this still work if the url is given like "https://www.aryn.ai"?

crawler/http/crawler/settings.py
eric-anderson: Magically inside of scrapy, same as pretty much everything else in the config file.  I tested it and it was about 1/second.
https://docs.scrapy.org/en/latest/topics/settings.html#download-delay

crawler/http/crawler/spiders/sycamore.py
eric-anderson: Since that's exactly the string I used to test it, yes.  It downloaded from docs.aryn.ai and blog.aryn.ai 

Why would you think it wouldn't?

austintlee (approved):

PR #184, opened 2023-11-27
Element splitter to prevent text elements with too many tokens.

alexaryn:


examples/s3_ingest.py
```diff
@@ -38,8 +38,10 @@
 ds = (
     ctx.read.binary(paths, binary_format="pdf", filesystem=fsys)
     .partition(partitioner=UnstructuredPdfPartitioner())
-    .merge(merger=GreedyTextElementMerger(tokenizer=tokenizer, max_tokens=512))
     .extract_entity(entity_extractor=OpenAIEntityExtractor("title", llm=davinci_llm, prompt_template=title_template))
+    .merge(merger=GreedyTextElementMerger(tokenizer=tokenizer, max_tokens=512))
+    .spread_properties(["path", "title"])
+    .split_elements(tokenizer=tokenizer, max_tokens=512)
```
austintlee, line 44: I'm wondering if this should just be rolled into the `merge` transform and have it merge without creating too large chunks.  Otherwise, one has to remember to call `split_elements` after doing `merge`.

alexaryn: You're not wrong.  The merge and split are dirty laundry to a certain extent.  I think Vinayak's idea of a chunker that is passed to the partitioner would clean this up a bit.  Another way to go would be to have the embedder do the chunking it needs.  For power-users, it may still be useful to have top-level ad-hoc merge and split operators.

sycamore/transforms/split_elements.py
```diff
@@ -0,0 +1,112 @@
+from ray.data import Dataset
+
+from sycamore.data import Document, Element
+from sycamore.functions.tokenizer import Tokenizer
+from sycamore.plan_nodes import Node, Transform, SingleThreadUser, NonGPUUser
+from sycamore.transforms.map import generate_map_function
+
+
+class SplitElements(SingleThreadUser, NonGPUUser, Transform):
```
HenryL27, line 9: Can we do the OOP thing and turn this into an ElementSplitter or some such interface and then subclass as something like the `RecursiveDivisionElementSplitter`

HenryL27 (approved):
sorry this took a while :)

PR #185, opened 2023-11-27
Fix text chunking for html importing to improve result quality.

eric-anderson:
In particular, text chunking to get larger partitions.  Note fix is from Vinayak.

### DIFF
--- a/importer/docker_local_import.py
+++ b/importer/docker_local_import.py
@@ -13,7 +13,7 @@ sys.path.append("../sycamore")
 sys.path.append("/app")
 
 import sycamore
-from sycamore.functions import HuggingFaceTokenizer
+from sycamore.functions import HuggingFaceTokenizer, TextOverlapChunker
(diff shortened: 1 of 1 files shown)


baitsguy (approved):

importer/docker_local_import.py
```diff
@@ -330,7 +330,12 @@ def import_html(paths):
     ctx = sycamore_init()
     (
         ctx.read.binary(paths, binary_format="html", filter_paths_by_extension=False)
-        .partition(partitioner=HtmlPartitioner())
+        .partition(
+            partitioner=HtmlPartitioner(
+                extract_tables=True,
+                text_chunker=TextOverlapChunker(chunk_token_count=4000, chunk_overlap_token_count=400),
```
austintlee, line 336: There is an alternative to this.  When you feed search results to an LLM, you always include the elements surrounding the results.  That might give applications more control.

PR #186, opened 2023-11-29
Remove unneeded mapping in OpenSearch index settings.

alexaryn:


HenryL27:
One thing I've noticed: the properties (whose types are not set in the settings) come out with type 
`{"type": "text", "fields": {"keyword": {"type": "keyword"}}}`
Can you confirm that this change doesn't do that to the text_representation (or that we don't care if that happens)?

alexaryn:
@HenryL27, I mostly did this to sync with what the container folks are doing.  @baitsguy insists it's OK.  I've done a test and queries return reasonable results.  I'm not quite sure what the technical interpretation of "don't care" is in this context.

baitsguy:
Seems like text_representation already has a similar pattern so this shouldn't change anything:

```
...
        "text_representation": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        },
```

It seems like this implies that text_representation is stored as "text" (for text queries) but also as a keyword (to be used in filtering/joins)

baitsguy (approved):

PR #187, opened 2023-11-29
Simple transform to perform regex replacement on Elements.

alexaryn:
Includes DocSet.coalesce_whitespace() which is a simple application of regular expression replacement.

### DIFF
--- a/sycamore/docset.py
+++ b/sycamore/docset.py
@@ -335,39 +335,17 @@ class DocSet:
 
         Example:
             .. code-block:: python
+            from sycamore.transforms import COALESCE_WHITESPACE
             ds = context.read.binary(paths, binary_format="pdf")
                 .partition(partitioner=UnstructuredPdfPartitioner())
-                .regex_replace([
-                    (r"\d+", "1313"),
-                    (r"old", "new"),
-                ])
+                .regex_replace(COALESCE_WHITESPACE)
+                .regex_replace([(r"\d+", "1313"), (r"old", "new")])
                 .explode()
         """
         from sycamore.transforms import RegexReplace
 
         plan = RegexReplace(self.plan, spec, **kwargs)
         return DocSet(self.context, plan)
-
-    def coalesce_whitespace(self, **kwargs) -> "DocSet":
-        """
-        Reduce long repetitions of whitespace characters to a single space
-        in the text_representation of every Element in each Document.
-
-        Example:
-            .. code-block:: python
-            ds = context.read.binary(paths, binary_format="pdf")
-                .partition(partitioner=UnstructuredPdfPartitioner())
-                .coalesce_whitespace()
-                .explode()
-        """
-        return self.regex_replace(
-            [
-                (r"\s+", " "),
-                (r"^ ", ""),
-                (r" $", ""),
-            ],
-            **kwargs,
-        )
 
     def map(self, f: Callable[[Document], Document], **resource_args) -> "DocSet":
         """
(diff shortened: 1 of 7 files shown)


sycamore/docset.py
```diff
@@ -328,6 +328,47 @@ def merge(self, merger: ElementMerger, **kwargs) -> "DocSet":
...
+        """
+        return self.regex_replace(
+            [
+                (r"\s+", " "),
+                (r"^ ", ""),
+                (r" $", ""),
+            ],
+            **kwargs,
+        )
+
```
HenryL27, line 371: I think in the interest of not adding too much to `docset` we can elide this method? We can just do this in the script, right?

alexaryn: We can.  I was thinking to provide ease-of-use, and self-documentation for common cases.  I think the growth of DocSet is inevitable and we should make a plan for how to organize it.  Or, maybe we realize that DocSet is just a doorway to transformer classes and we drop the habit of creating 1:1 helper methods for transformers.  I dunno.

eric-anderson: What if we make the transformer:
def regex_replace(self, res, **kwargs) -> "DocSet":
    # handle single re, list of res here.

and have
COALESCE_WHITESPACE = [ (r"\s+", " "), (r"^ ", ""), (r" $", "") ]

so people can call it as
docset
   .regex_replace(transforms.COALESCE_WHITESPACE)

baitsguy: I also think a regex_replace/apply is a little more scalable than coalesce_whitespace. I'd even think having a defined coalesce function and having the user pass that in to .map is fine

sycamore/tests/unit/transforms/test_regex_replace.py
```diff
@@ -0,0 +1,54 @@
...
+            "properties": {"path": "/docs/foo.txt", "title": "bar"},
+            "elements": [
+                {
+                    "type": "UncategorizedText",
+                    "text_representation": " \t foo   \t\t\n\t \t\n\n \t   bar\t  ",  # noqa: E501
+                },
+            ],
+        }
+    )
+
```
eric-anderson, line 33: Can you add tests for incorrect arguments?
1. Tolerate empty list and do nothing (maybe print a warning, not sure)
2. Not a list, some wrong things in the list (e.g. string rather than pair)
3. Not pairs in the list

sycamore/transforms/regex_replace.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Args:
+        child: The source node or component that provides the documents
+        spec: A list of tuples of regular expressions and substitutions,
+              to be executed in order via re.sub()
+        kwargs: Additional resource-related arguments that can be passed to the operation
+
+    Example:
+        .. code-block:: python
+            rr = RegexReplace(child=node, spec=[(r"\s+", " "), (r"^ ", "")])
```
eric-anderson, line 29: Since you don't need the kwargs and don't use them in examples, remove them here?

eric-anderson (changes requested):

sycamore/tests/unit/transforms/test_regex_replace.py
```diff
@@ -52,3 +61,13 @@ def test_via_execute(self, mocker):
...
         assert elems[0].text_representation == "foo bar"
+
+    def expectValid(self, spec):
+        RegexReplace(None, spec)
+
+    def expectInvalid(self, spec):
+        try:
+            RegexReplace(None, spec)
+            raise AssertionError("RegexReplace accepted bad spec")
+        except Exception:
```
eric-anderson, line 72: make sure to try to catch the error you expect.
except RuntimeError:
optionally 
except RuntimeError as r:
  if r.args[0] != "RegexReplace spec is not list[tuple[str, str]]":
     raise AssertionError("Wrong runtime error")

eric-anderson (approved):

PR #198, opened 2023-12-20
Add async-timeout dependency

eric-anderson:
I've seen this now several times on linux. While writing: https://github.com/aryn-ai/sycamore/pull/197 (a comment only change); I got a bunch of inexplicable failures in testing. Manually installing the module made it go away.

I also saw it while doing the local development guide on linux: https://github.com/aryn-ai/quickstart/blob/0591e5bbc9040f12545c94981f98e4f4ddb469e0/sycamore-local-development-example.md?plain=1#L92

poetry.lock file already has a version locked for this dependency, so I chose a version number that roughly matched with it.


bsowell (approved):
I confess I don't quite understand why putting this here makes a difference if it was already in poetry.lock, but I'm fine with it given that it does.

PR #199, opened 2023-12-20
Near-Duplicate Detection in Sycamore: Document Tagging and Document Dropping

alexaryn:
Lots of explanation in simhash.py.  Adds two transforms:
- Sketcher - adds SimHash list to each Document
- SketchUniquify - drops documents based on SimHash, for small docsets


sycamore/docset.py
```diff
@@ -395,6 +395,24 @@ def regex_replace(self, spec: list[tuple[str, str]], **kwargs) -> "DocSet":
         plan = RegexReplace(self.plan, spec, **kwargs)
         return DocSet(self.context, plan)
 
+    def chain(self, cls: Type[Node], **kwargs) -> "DocSet":
```
bsowell, line 398: I like this idea I think. We've been debating how to expose transforms without having to modify The DocSet class every time, and this seems like a reasonable approach, though the whole Node/Transform set of interfaces isn't necessarily super easy to use. 

Minor comments: 
- Should it be `Type[Transform]`? That only excludes scans, I think, which don't make sense here anyway. 
- If it were me I would probably call this something like `apply` or `apply_transform`, but I'm not going to fight too hard about naming.

sycamore/functions/rabin_karp.py
```diff
@@ -0,0 +1,88 @@
...
+
+class RkHash:
+    """
+    Class RkHash encapsulates a 64-bit Rabin-Karp hash function.  It is
+    meant to hash one byte [0,255] at a time.  It's not the best hash
+    function, but has the nice property of allowing previous inputs to
+    be removed.  This makes it efficient for sliding-window applications.
+    """
+
+    __slots__ = ["shift", "prime", "primeShift", "val", "inv"]
```
bsowell, line 12: I don't see __slots__ used very often. Does it help here with performance?

sycamore/functions/rabin_karp.py
```diff
@@ -0,0 +1,88 @@
...
+            ary.append(seed)
+        self.hasher = hasher
+        self.ary = ary
+        self.idx = 0
+        self.width = width
+
+    def __str__(self) -> str:
+        h = str(self.hasher)
+        a = ""
+        for ii, ch in enumerate(self.ary):
```
bsowell, line 63: This is just for debugging right? I believe that concatenating strings this way is notoriously slow in python -- probably join + a comprehension would be faster (though obviously would have to measure). Not a big deal.

sycamore/functions/rabin_karp.py
```diff
@@ -0,0 +1,88 @@
...
+        if width < 1:
+            raise ValueError
+        self.shift = 8
+        self.prime = 36028797018963913  # largest prime < 2^55
+        self.primeShift = self.prime * self.shift
+        self.val = 0
+        self.inv = reduce(lambda x, _: (x << self.shift) % self.prime, range(width - 1), 1)
+
+    def __str__(self) -> str:
+        return "(rk:%d:%d)" % (self.val, self.inv)
```
bsowell, line 24: You are of course free to format strings however you want, but just in case you hadn't seen them, I find Pythons [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) to be one of the biggest quality of life improvements in recent versions. Course that may just be that I also forget the stupid % codes.

sycamore/data/document.py
```diff
@@ -78,6 +78,14 @@ def embedding(self) -> Optional[list[float]]:
     def embedding(self, embedding: list[float]) -> None:
         self.data["embedding"] = embedding
 
+    @property
+    def simHashes(self) -> Optional[list[int]]:
```
bsowell, line 82: Do we want simHashes to be a top-level field in all Documents? I guess this is a bit like the question of what transforms to add to DocSet, but it's not immediately obvious to me why this should be top-level vs a property. If we do want to keep it at the top-level, do we want to call it something more generic (fingerprint, dedupe_val, ...) to allow for potentially other algorithms in the future, or do we think we would add another top-level field?

sycamore/transforms/sketcher.py
```diff
@@ -0,0 +1,124 @@
...
+           dataset = xform.execute()
+    """
+
+    def __init__(self, child: Node, threshold: float = 16, **kwargs):
+        super().__init__(child, **kwargs)
+        self.threshold = threshold
+
+    def execute(self) -> Dataset:
+        ds = self.child().execute()
+        ds = ds.materialize()
```
bsowell, line 104: Doesn't this materialize on multiple nodes (in the object store)? Guess I was expecting a take_all.

bsowell (commented):
First pass. Need a few more minutes to go through the rest, but I'm not going to get to it until after some meetings. Bunch of questions, but don't see any big structural issues.

sycamore/transforms/sketcher.py
bsowell: Thinking about this a bit more, I'm trying to compare `materialize()` and then `iter_rows()` with just doing a `take_call`. I guess the former materializes in the aggregate memory of the cluster and then iterates over the rows on a single node, so theoretically you don't need the whole thing in memory. But since you are still storing all the simHashes in memory it might not matter much?

bsowell:
Also, do you know why the linting and unit tests are failing? I know the integ tests are sometimes flakey, but the others used to pass reliably.

sycamore/docset.py
alexaryn: Could we call it `transform`?

sycamore/functions/rabin_karp.py
alexaryn: I just ran a benchmark and it's pretty much the same.  I'll take it out.

sycamore/functions/rabin_karp.py
alexaryn: Yes, just for debugging.  I changed it though.

sycamore/functions/rabin_karp.py
alexaryn: sprintf-style (%) formatting of ints is faster than format(), which is what f-strings are.  One nice thing about f-strings, though, is that they support the 'b' type for binary output, which % doesn't for some reason.  https://stackoverflow.com/questions/40714555/python-string-formatting-is-more-efficient-than-format-function

sycamore/data/document.py
alexaryn: I was thinking it's similar to embeddings, so followed that pattern.  I wanted to name it specifically so that there wouldn't be a name clash if we wanted to use another algorithm in the future.  The only major requirement is that they end up in OpenSearch in such a way that we can perform retrieval-time deduplication.

sycamore/transforms/sketcher.py
alexaryn: I'm no Ray expert, but I think I agree with your later comment.  The important thing is that what I'm storing in memory is a tiny subset of each Document.  Reading the doc, it seems that take_all() is more likely to OOM the local host, but that's speculation.

bsowell (approved):
LGTM. Certainly can't claim that I didn't miss something in the low level calculations, but integration into Sycamore looks reasonable. I expect this will need more documentation before it can be widely used.

PR #205, opened 2024-01-04
Correctly handle OpenAI model fallback.

### LINKED ISSUE #204: TEXT_DAVINCI deprecation crashes instead of falling back to gpt-3.5-turbo-instruct
The behavior of https://github.com/aryn-ai/sycamore/commit/f5b8854743c9497afccc0a06b4d005c56f36660b was intended to automatically replace usage of `text-davinci-003` with `gpt-3.5-turbo-instruct` after printing a warning. However, due to a bug here https://github.com/aryn-ai/sycamore/blob/92051a564f1a3197c731a52d277b285932a44be4/sycamore/llms/llms.py#L57, we pass the Enum object rather than the value, which causes operations to crash. 

Until we fix this, the workaround is to use the gpt-3.5-turbo-instruct model explicitly.

bsowell:
Resolves #204

### DIFF
--- a/sycamore/llms/llms.py
+++ b/sycamore/llms/llms.py
@@ -51,10 +51,12 @@ class OpenAIClientParameters:
 
 
 class OpenAI(LLM):
-    def __init__(self, model_name, api_key=None, params: OpenAIClientParameters = OpenAIClientParameters(), **kwargs):
+    def __init__(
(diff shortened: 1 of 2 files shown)


bsowell:
WIll fix linting.

bohou-aryn (approved):

PR #208, opened 2024-01-09
Augment text

### LINKED ISSUE #206: Augmenting text representation with metadata
**Is your feature request related to a problem? Please describe.**
A lot of the time the actual text representation of a chunk will not include enough information to generate a useful embedding or perform adequate keyword search queries. Including some of the metadata that has been extracted in earlier pipeline steps as part of the text representation for the chunk can greatly increase performance. e.g.:
```
document {
  text_representation: "maintenance procedure: step 1. [blah] step 2. [blah] step 3. [blah]",
  properties {
    part_name: "zebulon fork A3"
  }
}
```
Let's provide a way to include the part name in the text, e.g.:
```
document {
  text_representation: "this pertains to the part "zebulon fork A3". maintenance procedure: step 1. [blah] step 2. [blah] step 3. [blah]",
  properties {
    part_name: "zebulon fork A3"
  }
}
```

**Describe the solution you'd like**
Implement a transform `augment_text` that takes a list of format strings. Replace the text representation of each document by formatting and joining the format strings. We take a list of such fmt strings because some documents may be missing certain properties, and so rather than voiding the entire operation in that case, we can simply drop the segment that's missing a value. Likely the original `text_representation` should be the final element, but I don't think that's a requirement. My recommendation is an API like so:
```
docset.augment_text([
    "This pertains to the part {doc.properties['part_name']}.",
    "{doc.text_representation}"
])
```

**Describe alternatives you've considered**
**Alt 1.**
```
docset.augment_text(prefix = ["This pertains to the part {doc.properties['part_name']}."], include_original_text=True, suffix=[])
```
This acts the same way, but treats the `text_representation` specially. I think it's unnecessarily messy though.
**Alt 2.**
Pass in a single format string
```
docset.augment_text("This pertains to the part {doc.properties['part_name']}. {doc.text_representation}")
```
If a document doesn't have a part name, this will cause problems. I mean, in this case it might work out ok but due to the error handling simply leaving the text_representation as it is, which would probably be intended, but in cases with more complicated augmentations I don't think the partial failure mode should be "do nothing"
**Alt 3**
Pass in a function
```
docset.augment_text(lambda doc: f"This pertains to the part {doc.properties['part_name'}. {doc.text_representation}")
```
This lets the user control exactly what happens - they just have to provide a function that takes a document and spits out a string. In the example I've done something simple, but this could be arbitrarily complex. But if we're going with this route, then the user might as well just use `docset.map()`, because the body of augment_text would just be something like
```
def augment_text(self, fn):
    def f(doc):
        doc.text_representation = fn(doc)
        return doc
    self.map(f)
```
**Alt 4**
Do nothing and just recommend people use `map` as in Alt 3. I feel like this is a feature we want directly exposed in the API though.
**Alt 5**
Alt 2, but you give it a list of format strings to fall back to. e.g.
```
docset.augment_text(["This pertains to the part {doc.properties['part_name']}. {doc.text_representation}", "There is not part name for this document. {doc.text_representation}"])
```
I just think this is extremely clunky.

**Additional context**
One interesting nuance to all this is that we're likely adding a whole bunch of tokens to the text representation, meaning that we're also likely to overflow the embedding models' context window if we're not careful. I'm not really sure what can be done to mitigate this other than just manually tuning the segmentation since the partitioning happens first and the number of tokens we're adding is a little random. Additionally, adding splitting steps after the text augmentation brings us right back to the place from which this augmentation came from. Maybe further down the line we can augment augmentation to also take in a splitter or something so it can dynamically calculate how many tokens it will add to a document and then split if it needs to. But doing something like this would be an argument for Alternative 1.

HenryL27:
resolves #206


HenryL27:
currently three transforms:
1. `FStringTextAugmentor` uses eval to lazily evaluate f-strings. This is pretty lightweight but has some rather serious security issues - users could do things like `FStringAugmentor(["{__import__('os').system('rm -rf /')}", "{exit()}"])` and then watch the entire ray cluster nuke itself. But on the other hand, users could do that anyway with map.
2. `UDFTextAugmentor` asks for a function that maps documents to strings and updates docs' text representations accordingly. This is probably the most heavyweight API, but also the most powerful b/c the user can do whatever the heck they want. But also it's really just a thin layer over `map`.
3. `JinjaTextAugmentor` asks for a [jinja2](https://jinja.palletsprojects.com/en/3.1.x/templates/) template string and throws the jinja rendering engine at it (using a sandboxed env for more safety). This may be the best solution in terms of power, safety, and simplicity, but it also has the benefit of bringing in another rather large library to the sycamore dependency list; and as we all know, sycamore is the mother of all libraries.

FString and Jinja augmentors also allow the user to specify a set of modules to use in the augmentation - so I can do something like this:
```python
from sycamore.transforms.augment_text import JinjaTextAugmentor
from sycamore.transforms.regex_replace import COALESCE_WHITESPACE
import pathlib
template = '''This document is from {{ pathlib.Path(doc.properties['path']).name }}.
The title is {{ doc.properties['title'] }}.
The authors are {{ doc.properties['authors'] }}.
{% if doc.text_representation %}
    {{ doc.text_representation }}
{% else %} 
    There is no text representation for this
{% endif %}
'''
aug = JinjaTextAugmentor(template=template, modules={"pathlib": pathlib})
aug_docset = exp_docset.augment_text(aug).regex_replace(COALESCE_WHITESPACE)
aug_docset.show(show_binary=False, truncate_content=False)
```

sycamore/transforms/augment_text.py
```diff
@@ -0,0 +1,162 @@
...
+                There is no text representation for this
+            {% endif %}
+            '''
+            aug = JinjaTextAugmentor(template=template, modules={"pathlib": pathlib})
+            aug_docset = exp_docset.augment_text(aug).regex_replace(COALESCE_WHITESPACE)
+            aug_docset.show(show_binary=False, truncate_content=False)
+    """
+
+    def __init__(self, template: str, modules: dict[str, Any] = {}):
+        from jinja2.sandbox import SandboxedEnvironment
```
bsowell, line 133: How does the SandboxedEnvironment work? Does it create a separate Python interpreter or something? Just want to understand the implications since it's run in parallel for every task.

bsowell: I find myself liking the Jinja syntax more than I expected, though definitely agree about adding infinite dependencies. @baitsguy since you have actually tried to construct the metadata, what do you think? I suppose if I were to vote I would say keep the UDF version and the Jinja version.

bsowell:
I find myself liking the Jinja syntax more than I expected, though definitely agree about adding infinite dependencies. @baitsguy since you have actually tried to construct the metadata, what do you think? I suppose if I were to vote I would say keep the UDF version and the Jinja version.

sycamore/transforms/augment_text.py
HenryL27: As I understand it, the jinja Environment object is essentially an interpreter that runs in python and parses jinja stuff. The SandboxedEnvironment is a subclass of that that can forbid certain operations and stuff. So it's still all in one process, but it's a rather heavyweight function call. That said, it ain't got nothin on openai.

It isn't completely secure, and we might want to add to the security it comes with. But I think it has APIs that make that at least doable. I haven't tried to mess with my filesystem with jinja template rendering, but I'll bet it's not impossible.

sycamore/transforms/augment_text.py
```diff
@@ -0,0 +1,162 @@
...
+        self._modules = {m.__name__: m for m in modules}
+
+    def augment_text(self, doc: Document) -> Optional[str]:
+        formatted = []
+        for s in self._sentences:
+            try:
+                evaluated = eval(s, self._modules, {"doc": doc})
+                formatted.append(evaluated)
+            except Exception as e:
+                print(f"augment_text: Document {doc.doc_id} failed in sentence {s} with error {e}")
```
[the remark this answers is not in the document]
HenryL27: we just use the `logging` package for sycamore/ray or is there something fancy?

sycamore/transforms/augment_text.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __init__(self, child: Node, text_augmentor: TextAugmentor, **kwargs):
+        super().__init__(child, **kwargs)
+        self._augmentor = text_augmentor
+
+    def execute(self) -> Dataset:
+        input_ds = self.child().execute()
+
+        def augment_text(doc: Document) -> Document:
+            doc.text_representation = self._augmentor.augment_text(doc)
```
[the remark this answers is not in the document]
HenryL27: So our embedder interface can already take a `pre_process_function` - maybe it would be worth adding a `__call__` method for augmentors so it can be used there? 
In general I feel like you want to add metadata to both the embedded text and the keyword-retrieved text, since a lot of the time the metadata may include weird words that the embedding model doesn't know about. But I agree, you may want different formatting for embeddings vs keywords.

I think I've convinced myself to add a `__call__` so you can use augmentors as embedding preprocessors

sycamore/transforms/augment_text.py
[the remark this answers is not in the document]
HenryL27: dropping the fstring augmentor so irrelevant

HenryL27:
btw I'm not sure that `TextAugmentor` is really the best name for this class/feature. Maybe `TextFormatter`? Idk, that doesn't feel quite right to me. Not sure. Any suggestions?

sycamore/docset.py
```diff
@@ -218,6 +219,35 @@ def spread_properties(self, props: list[str], **resource_args) -> "DocSet":
         plan = SpreadProperties(self.plan, props, **resource_args)
         return DocSet(self.context, plan)
 
+    def augment_text(self, augmentor: TextAugmentor, **resource_args) -> "DocSet":
+        """
+        Augments text_representation with external information.
```
bsowell, line 224: You might need to add a directive in the docs TOC to make sure this gets included.

sycamore/tests/unit/transforms/test_augment_text.py
```diff
@@ -0,0 +1,42 @@
...
+    doc = Document(
+        {
+            "doc_id": "doc_id",
+            "type": "pdf",
+            "text_representation": "text",
+            "properties": {"path": "/docs/foo.txt", "title": "bar"},
+        }
+    )
+
+    def test_udf_augmentation(self, mocker):
```
bsowell, line 17: super minor: You don't need the mocker here, do you? There isn't any mocking happening?

sycamore/transforms/augment_text.py
bsowell: Got it. As I understand it, this should be instantiated once per worker since you create it in the __init__ call. Is that right? I want to avoid creating a separate sandbox per document, but I don't think this is doing that.

bsowell (approved):
Couple of minor comments (mostly the docs one), but otherwise LGTM.

sycamore/tests/unit/transforms/test_augment_text.py
HenryL27: nope!

sycamore/docset.py
HenryL27: ack

sycamore/transforms/augment_text.py
HenryL27: yep, should be just one sandbox per worker. Now, we create the same template over and over - jinja doesn't want to pickle templates and that caused problems. Could maybe get around that by doing some sort of check at runtime and if template isn't there then and only then do you compile it, sorta like what the embedder does?

bsowell: What do you mean create the same template? Isn't the template just a string?

HenryL27: theres a jinja Template object (`self._env.from_string(source=self._template, globals=self._modules)`). I could store this and just call `Template.render` a whole bunch with different Documents. But it doesn't like to serialize

PR #217, opened 2024-01-24
Add some useful utility methods for DocSets.

bsowell:
This commit adds the `with_property`, `with_properties`, and `map_elements`. This are primarily syntactic sugar around map operations that I have found to be useful when writing complex scripts. The `with_property` and `with_properties` methods simplify adding properties to each Document in the in the DocSet. They are loosely modeled after similar methods on Spark Datasets.

The `map_elements` method applies a function to each element in each Document in the DocSet. This is a thin wrapper around the Document-level `map` operation, but helps reduce boilerplate.

### DIFF
--- a/sycamore/docset.py
+++ b/sycamore/docset.py
@@ -1,10 +1,11 @@
+from collections.abc import Mapping
 import logging
 import pprint
 import sys
 from typing import Callable, Optional, Any, Iterable, Type
 
 from sycamore import Context
-from sycamore.data import Document
+from sycamore.data import Document, Element
 from sycamore.functions.tokenizer import Tokenizer
 from sycamore.plan_nodes import Node, Transform
 from sycamore.transforms.augment_text import TextAugmentor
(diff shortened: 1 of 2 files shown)


sycamore/docset.py
```diff
@@ -529,6 +577,26 @@ def map_batch(
...
+        Applies the given mapping function to each element in the each Document in this DocsSet.
+
+        Args:
+            f: A Callable that takes an Element and returns an Element. Elements for which
+                f evaluates to None are dropped.
+        """
+
+        def process_doc(doc: Document) -> Document:
+            new_elements = []
+            for e in doc.elements:
```
baitsguy, line 591: Given the current data model I don't think we can parallelize/distribute this to a map call over elements right? Docs can have a large amount of elements and ops can get expensive

bsowell: Yeah, this doesn't solve the problem of a huge number of elements in a single document. At the moment I don't think we have a better mechanism for that than exploding and processing the elements as documents.

sycamore/docset.py
```diff
@@ -202,6 +203,53 @@ def partition(
...
+
+        def add_property_fn(doc: Document) -> Document:
+            new_props = doc.properties
+            new_props[name] = f(doc)
+            doc.properties = new_props
+            return doc
+
+        return self.map(add_property_fn, **resource_args)
+
+    def with_properties(self, property_map: Mapping[str, Callable[[Document], Any]], **resource_args) -> "DocSet":
```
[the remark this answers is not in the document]
bsowell: Yeah, that's fair. The API is borrowed from Spark, but I can change `with_property` to call `with_properties`

baitsguy (approved):

PR #218, opened 2024-01-24
Make Textract disabled by default

bohou-aryn:
For detail refer @jon's refactoring doc Sycamore Deployment - Containers https://www.notion.so/Sycamore-Deployment-Containers-91c4f7602b984e9991642f5cc8193ba9?pvs=4


eric-anderson (changes requested):
See comment in quickstart, I think we want to get rid of the ENABLE_TEXTRACT env var.

eric-anderson (approved):

PR #219, opened 2024-01-25
Create notebook file with default ingest script

bohou-aryn:
Take the default script being used for ingest and create a notebook file so customers can more easily see the script and make edits. For detail refer @jon's refactoring doc Sycamore Deployment - Containers https://www.notion.so/Sycamore-Deployment-Containers-91c4f7602b984e9991642f5cc8193ba9?pvs=4


eric-anderson (changes requested):
There was some confusion on this task. The goal wasn't "give a notebook with the entire ingest script in it" it was "give a notebook that lets people use the ingesting that the default script uses" So the notebook should be mostly the import_pdf and import_html functions but fixed so that the inputs can be chosen (rather than assuming the path).

Think "As a user I want to adjust the default sycamore importing script to experiment with different ways to process data in sycamore."

Eventually the goal would be that we could swap out the code someone writes into the container so that it can run automatically.

eric-anderson:
Pseudo-code of the cells for the notebook

# setup (import python modules)
# curl <some files from sortbenchmark.org> -- similar to this notebook
  # optionally could point people at the download directory /app/work/crawl_dir/downloads, but unless they run the crawler it wouldn't have anything
# define a slight variant of the process_pdf & process_html things; could just take a path root since that works.
  # maybe let people choose the index
# call process_pdf & process_html on the directory where the files were curled
# print link to demo ui

jonfritz:
Yes - Eric is right. Basically, have a notebook that when I click "run all the cells" it does what the default code in the Importer does. The core data processing code should be the same in each.

jonfritz:
Should we name the notebook "default-prep-script"?

notebooks/docker_local_import.ipynb
```diff
@@ -0,0 +1,426 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 39,
+   "id": "e57c0521-4eae-4323-b367-ea94a80938a6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def get_os_client_args():\n",
+    "    args = {\n",
+    "        \"hosts\": [{\"host\": \"localhost\", \"port\": 9200}],\n",
```
eric-anderson, line 165: s/localhost/opensearch/
and add # Change this to localhost if you are running outside of the container.

eric-anderson: Alternately, and more complexly, check to see if you're in the container with a stat of /app/work/AAA_SEE_README_FOR_PERSISTENT_DATA_DIRECTORIES; if it exists it's in the container, otherwise outside.

eric-anderson (commented):
Here's the notebook I had after simplifying it a bunch and testing it.  I think I made comments for all the changes, but am not sure.

[docker_local_import.ipynb.txt](https://github.com/aryn-ai/sycamore/files/14214679/docker_local_import.ipynb.txt)

eric-anderson:
I agree docker_local_import.ipynb is an odd name to me.

notebooks/default-prep-script.ipynb
```diff
@@ -0,0 +1,417 @@
...
+    "from sycamore.llms import OpenAI, OpenAIModels\n",
+    "from sycamore.transforms.embed import SentenceTransformerEmbedder\n",
+    "from sycamore.transforms.extract_entity import OpenAIEntityExtractor\n",
+    "from sycamore.transforms.merge_elements import GreedyTextElementMerger\n",
+    "from sycamore.transforms.partition import UnstructuredPdfPartitioner, HtmlPartitioner\n",
+    "\n",
+    "index = \"demoindex0\""
+   ]
+  },
+  {
```
eric-anderson, line 31: I think this cell is obsolete now also, it's only used to print out file paths not in the execution.

notebooks/default-prep-script.ipynb
```diff
@@ -0,0 +1,417 @@
...
+   "source": [
+    "%%bash\n",
+    "pwd\n",
+    "sudo mkdir -p /app/work/crawl_data/downloads/pdf\n",
+    "sudo mkdir -p /app/work/crawl_data/downloads/html\n",
+    "sudo curl https://sortbenchmark.org/ELSAR2022.pdf -o /app/work/crawl_data/downloads/pdf/elsar.pdf\n",
+    "sudo curl https://sortbenchmark.org -o /app/work/crawl_data/downloads/html/home.html"
+   ]
+  },
+  {
```
eric-anderson, line 89: I think this cell is unneeded, if you want to show the files, I'd just do:
! find /app/work/crawl_data/downloads -print
But I don't think that's helpful, the curl lines either download or fail.

notebooks/default-prep-script.ipynb
```diff
@@ -0,0 +1,417 @@
...
+    "    html_files = find_recursive(downloads, \"html\")\n",
+    "    return pdf_files + html_files"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "04cb55d4-3ad7-4003-bc09-6ea1b1ca4432",
+   "metadata": {},
+   "outputs": [
```
eric-anderson, line 59: Make sure to clear the output. You need to run:
poetry run pre-commit run nbstripout --all-files
or do a "reset kernel and clear all outputs" in Jupyter.

eric-anderson (approved):

PR #220, opened 2024-01-26
Metadata extraction updates

baitsguy:
1. Added a new notebook to extract metadata from ntsb samples
2. Handle inconsistent json responses from openai and guidance
3. Ability to customize max number of properties in schema


sycamore/utils/extract_json.py
```diff
@@ -1,11 +1,15 @@
...
-        raise ValueError("JSON block not found in LLM response")
+    try:
+        return json.loads(payload)
+    except (ValueError, TypeError, JSONDecodeError):
+        pattern = r"```json([\s\S]*?)```"
+        match = re.match(pattern, payload)
+        if match:
+            return json.loads(match.group(1))
+        else:
+            raise ValueError("JSON block not found in LLM response: " + str(payload))
```
HenryL27, line 15: What's the difference between these (besides the better error message)?

HenryL27 (approved):
LGTM, just one question

jonfritz (approved):
LGTM.

sycamore/utils/extract_json.py
baitsguy: we try to first parse the string as json directly, if that fails then we try to parse it as:
```json
{actual json}
```
which is what openai/guidance seems to be returning sometimes

PR #230, opened 2024-02-06
Query transform and NestedJsonReader

baitsguy:
1. Create a wrapper Document around OpenSearch queries/responses
2. Add `query` transform to send those requests
3. Added a `NestedJsonReader`, which can generate a list of Documents from a single JSON file. It requires a parser method to be passed in which would know about the specific JSON being read. This is a common format for evaluation datasets. 

Sample code is in the integration test, pasting here for ease:

```
query_executor = OpenSearchQueryExecutor(OS_CLIENT_ARGS)

query1 = OpenSearchQuery()
query1.query = {"query": {"match_all": {}}, "size": 1}
query1.index = self.INDEX

query2 = OpenSearchQuery()
query2.query = {"query": {"match_all": {}}, "size": 2}
query2.index = self.INDEX

queries = [query1, query2]

context = sycamore.init()
result = context.read.document(queries).query(query_executor=query_executor)
```


sycamore/transforms/query.py
```diff
@@ -0,0 +1,57 @@
...
+
+from opensearchpy import OpenSearch
+from ray.data import Dataset
+
+from sycamore.data import OpenSearchQueryResult, Element, OpenSearchQuery
+from sycamore.plan_nodes import Node, NonCPUUser, NonGPUUser, Transform
+from sycamore.utils.generate_ray_func import generate_map_function
+
+
+class QueryExecutor(ABC):
```
bsowell, line 12: Trying to understand when we would have another executor. Aren't the query objects kind of OpenSearch specific?

sycamore/transforms/query.py
```diff
@@ -0,0 +1,57 @@
...
+            "POST",
+            url=f"/{query['index']}/_search",
+            params=query.get("params", None),
+            headers=query.get("headers", None),
+            body=query["query"],
+        )
+
+        result = OpenSearchQueryResult()
+        result.result = os_result
+        result.hits = [Element(hit["_source"]) for hit in os_result["hits"]["hits"]]
```
HenryL27, line 39: maybe keep the scores? idk if that's necessary tbh but seems like something that would be nice to have in some cases (at the very least debugging)

HenryL27 (approved):

sycamore/scans/file_scan.py
```diff
@@ -226,3 +226,42 @@ def execute(self) -> Dataset:
...
+
+
+class NestedJsonScan(FileScan):
+    def __init__(
+        self,
+        paths: Union[str, list[str]],
+        *,
+        parallelism: Optional[int] = None,
+        filesystem: Optional[FileSystem] = None,
+        doc_extractor: Optional[Callable] = None,
```
bsowell, line 238: Is the only difference from the existing json scan the doc_extractor? Would it be easier to add the doc_extractor as an optional parameter to that scan rather than adding a new one?

sycamore/data/document.py
```diff
@@ -157,3 +157,110 @@ def from_row(row: dict[str, bytes]) -> "Document":
     def to_row(self) -> dict[str, bytes]:
         """Serialize this document into a row for use with Ray."""
         return {"doc": self.serialize()}
+
+
+class OpenSearchQuery(Document):
```
bsowell, line 162: I do wonder if we won't want to write a generic sort of wrapper that lets you make a document out of any basic sort of python object, but I'm okay deferring that.

sycamore/transforms/query.py
baitsguy: ^We store the entire result above so should be able to access everything. I'll verify

sycamore/transforms/query.py
baitsguy: I expect we'll want a simplified query interface (e.g. just send in a question instead of os specifics). I've also just been building all classes with the assumption of having a different db

sycamore/scans/file_scan.py
baitsguy: Yeah that'll actually simplify it, we do a flatmap but I believe the current json _to_document method should work fine with it?

sycamore/data/document.py
baitsguy: Yeah agree, we should discuss that

sycamore/transforms/query.py
baitsguy: Had a bug, fixed and verified. Tests with evaluation will cover it as well

bsowell (approved):

PR #241, opened 2024-02-14
Update get_started.md

jonfritz:
Lowercase "Docker" because it's a command.


jonfritz:
Added changes to the API reference section to add more top level items and split things apart.

docs/source/index.rst
```diff
@@ -147,4 +147,6 @@ More Resources
    :maxdepth: 2
    :hidden:
 
-   /APIs/index.rst
+   /APIs/data_preparation.rst
+   /APIs/conversation_memory.rst
+   /APIs/transforms.rst
```
bsowell, line 152: These look a little strange b/c you get the little triangle dropdown, but since they don't have any subpages, it doesn't do anything when you click on them. The links work if you click on the text, but it's a bit confusing. Let me see if there is a way to not render the dropdowns.

docs/source/APIs/transforms.rst
```diff
@@ -0,0 +1,24 @@
+.. _Ref-Transforms:
```
bsowell, line 1: This page isn't rendering correctly. Will need to look into why.

PR #242, opened 2024-02-15
Upgrade the openai Python library to 1.x and guidance to 0.1.x.

bsowell:
Both of these libraries made significant backwards-incompatible changes that must be done together, since the new version of Guidance depends on the new version of OpenAI, and the old version of Guidance will not work with the new version of OpenAI.

This commit includes roughly the following changes:

  1. Replace the `OpenAIClientParameters` class with the `OpenAIClientWrapper` class. This is reponsible for managing parameters and providing the appropriate arguments to either the `OpenAIClient` or the `AzureOpenAIClient`, as necessary. To provide some backwards compatibility I aliased `OpenAIClientParameters` to this class and continue to support some now-deprecated parameters.

  2. Update the OpenAIModels to make it clearer which are chat vs instruct models. Update the OpenAI LLM implementation to handle these parameters with some nod to compatibility. Instantiate the client lazily to avoid serialization issues. (An alternative is to implement the __reduce__ method to customize the pickle format, which I got working, but is more complicated/less explainable).

  3. Rewrite the prompts in default_prompts.py to use Guidance's new format.

  4. Update `_generate_using_openai` and `_generate_using_guidance` to use the new clients and syntax.

  5. Update the `summarize` and `extract_entity` transforms to use the new prompts.

  6. Update the `OpenAIEmbedder` class to use the new client wrapper and clients.

  7. Bunch of small test fixes to get everything working.

There is still work to be done before this is ready to go. In particular, while I have added a few tests, we definitely need some additional integration tests covering some of the OpenAI (and particularly Azure) parameters. I also feel the backwards compatibility logic made some of this overly complex and I would like another stab at cleaning things up a bit.


sycamore/llms/openai.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+@dataclass
+class OpenAIModel:
+    name: str
+    is_chat: bool = False
+
+
+class OpenAIModels(Enum):
+    TEXT_DAVINCI = OpenAIModel(name="text-davinci-003", is_chat=True)
```
baitsguy, line 29: this is deprecated now I think?

sycamore/llms/openai.py
```diff
@@ -0,0 +1,227 @@
...
+    ):
+        if isinstance(model_name, OpenAIModels):
+            self.model = model_name.value
+        elif isinstance(model_name, OpenAIModel):
+            self.model = model_name
+        elif isinstance(model_name, str):
+            self.model = OpenAIModels.from_name(model_name).value
+
+        if self.model.name == OpenAIModels.TEXT_DAVINCI.value.name:
+            print("text-davinci-003 is deprecated. Falling back to gpt-3.5-turbo-instruct")
```
baitsguy, line 168: side note: we should probably use a logger everywhere

sycamore/transforms/summarize.py
```diff
@@ -60,10 +61,10 @@ def summarize(self, document: Document) -> Document:
         return document
 
     def _summarize_text_element(self, element: Element) -> Element:
-        if self._llm.is_chat_mode:
-            prompt = TEXT_SUMMARIZER_GUIDANCE_PROMPT_CHAT
+        if self._llm.is_chat_mode():
+            prompt: GuidancePrompt = TextSummarizerGuidancePromptChat()
```
baitsguy, line 65: is it too complex if we implemented each prompt as a class with get_prompt_, get_chat_prompt methods? Might be easier to compose prompts and use in places like these

baitsguy (commented):
Thanks for the painful change, looks good functionally

sycamore/llms/openai.py
bsowell: Sure, will update.

sycamore/llms/openai.py
bsowell: It is, but we added fallback logic, and I didn't want to break that.

sycamore/transforms/summarize.py
bsowell: That's a good question. Let me play around it a bit. Totally on board if we can find a way to unify them. I also wonder if there is some way to extract basic prompts from specific LLMs. I know that won't work in all cases, but ideally if we wanted to try Anthropic or Mixtral for entity extraction, for instance, we wouldn't have to start from scratch.

sycamore/llms/openai.py
```diff
@@ -0,0 +1,252 @@
...
+        if self.client_type == OpenAIClientType.OPENAI:
+            return OpenAIClient(
+                api_key=self.api_key,
+                organization=self.organization,
+                base_url=self.base_url,
+                max_retries=self.max_retries,
+                **self.extra_kwargs,
+            )
+        elif self.client_type == OpenAIClientType.AZURE:
+            if self.azure_endpoint is None:
```
baitsguy, line 115: shouldn't this just be in the constructor with other validations?

sycamore/llms/openai.py
```diff
@@ -0,0 +1,252 @@
...
+                max_retries=self.max_retries,
+                **self.extra_kwargs,
+            )
+
+        else:
+            raise ValueError(f"Invalid client_type {self.client_type}")
+
+
+# Allow rough backwards compatibility
+OpenAIClientParameters = OpenAIClientWrapper
```
baitsguy, line 175: Didn't know you can do this, neat

sycamore/llms/openai.py
```diff
@@ -0,0 +1,252 @@
...
+        client_wrapper: Optional[OpenAIClientWrapper] = None,
+        params: Optional[OpenAIClientParameters] = None,
+        **kwargs,
+    ):
+        if isinstance(model_name, OpenAIModels):
+            self.model = model_name.value
+        elif isinstance(model_name, OpenAIModel):
+            self.model = model_name
+        elif isinstance(model_name, str):
+            self.model = OpenAIModels.from_name(model_name).value
```
baitsguy, line 192: do we need to fail here if doesn't match?

sycamore/llms/openai.py
```diff
@@ -0,0 +1,252 @@
...
+
+        kwargs = {"client_wrapper": self.client_wrapper, "model_name": self.model_name}
+
+        return (deserializer, (kwargs,))
+
+    def is_chat_mode(self):
+        return self.model.is_chat
+
+    def generate(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> Any:
+        if llm_kwargs is not None:
```
baitsguy, line 232: this method always confused me, can we have an explicit parameter that says use guidance vs not? do we always need llm_kwargs to make a direct open ai call?

sycamore/llms/prompts/default_prompts.py
```diff
@@ -1,129 +1,128 @@
...
+
+    def execute(self, model: GuidanceModel, **kwargs) -> str:
+        if isinstance(model, Chat):
+            return self._execute_chat(model, **kwargs)
+        else:
+            return self._execute_instruct(model, **kwargs)
+
+    def _execute_chat(self, model, **kwargs) -> str:
+        with system():
+            lm = model + self.system.format(**kwargs)
```
baitsguy, line 35: adding an object and string?

sycamore/transforms/embed.py
```diff
@@ -3,16 +3,19 @@
...
 import ray
 from ray.data import ActorPoolStrategy, Dataset
 from sentence_transformers import SentenceTransformer
-from tenacity import retry, stop_after_attempt, wait_random, retry_if_exception_type
 
 from sycamore.data import Document
-from sycamore.llms.llms import OpenAIClientParameters
+from sycamore.llms import OpenAIClientParameters
+
+# from sycamore.llms.llms import AzureOpenAI, OpenAIClientParameters
```
baitsguy, line 17: remove?

baitsguy (approved):
Thanks for the changes, looks like a lot of hard work. Had a few questions for clarifications, none are blocking

sycamore/llms/openai.py
bsowell: yeah, let me change that. Not quite sure why I didn't put it there to start.

sycamore/llms/openai.py
bsowell: Good catch.

sycamore/llms/openai.py
bsowell: Yes, I agree it's weird. Let me play around with that in a separate PR to avoid too much churn.

sycamore/llms/prompts/default_prompts.py
bsowell: Sorry, what do you mean by this?

sycamore/llms/prompts/default_prompts.py
```diff
@@ -1,129 +1,128 @@
...
-    {{#user~}}
-    Write a summary of the following. Use only the information provided.
-    Include as many key details as possible. Do not make up answer. Only return the summary as part of your answer."
-    {{query}}
-    {{~/user}}
 
-    {{#assistant~}}
-    {{gen "summary"}}
-    {{~/assistant}}
+class EntityExtractorFewShotGuidancePrompt(SimpleGuidancePrompt):
```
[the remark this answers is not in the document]
bsowell: I think they may get formatted as part of `query`, though I agree its confusing. I kind of want to take a stab at rewriting the extract_entity transform, but will do that as a separate PR.

sycamore/transforms/embed.py
bsowell: Ack. Thanks.

PR #244, opened 2024-02-17
Fix examples to use https for links to UI

eric-anderson:
Also add missing default-prep-script.ipynb to Dockerfile examples.


alexaryn (approved):
LGTM

PR #249, opened 2024-02-21
Upgrade opensearch to 2.12

HenryL27:
1. upgrades opensearch to 2.12
  a. Lets us turn of the feature flags
  b. except that the rag pipeline feature flag is a little bugged, so we have to do this flick operation
2. stores the 2.11 stuff in opensearch/2.11/
  a. not sure if this is necessary but felt like it might be useful to keep a copy around per version.

will want to follow the merging of https://github.com/aryn-ai/demo-ui/pull/18

### DIFF
--- a/opensearch/2.11/Dockerfile
+++ b/opensearch/2.11/Dockerfile
@@ -1,0 +1,27 @@
+# Repo name: arynai/sycamore-opensearch
+
+ARG TAG=2.11.0
+FROM opensearchproject/opensearch:$TAG
+USER 0
+RUN yum install -y jq && yum clean all
+USER 1000
+COPY opensearch/opensearch.yml /usr/share/opensearch/config/
+COPY opensearch/sycamore-opensearch.sh .
+ENV discovery.type=single-node
+ENV DISABLE_SECURITY_PLUGIN=true
+ENV DISABLE_INSTALL_DEMO_CONFIG=true
+
+ARG GIT_BRANCH="main"
+ARG GIT_COMMIT="unknown"
+ARG GIT_DIFF="unknown"
+
+ENV GIT_BRANCH=${GIT_BRANCH}
+ENV GIT_COMMIT=${GIT_COMMIT}
+ENV GIT_DIFF=${GIT_DIFF}
+
+LABEL org.opencontainers.image.authors="opensource@aryn.ai"
+LABEL git_branch=${GIT_BRANCH}
+LABEL git_commit=${GIT_COMMIT}
+LABEL git_diff=${GIT_DIFF}
+
+CMD ./sycamore-opensearch.sh
--- a/opensearch/sycamore-opensearch.sh
+++ b/opensearch/sycamore-opensearch.sh
@@ -31,6 +31,8 @@ main() {
         setup_security
     fi
 
+    flick_rag_feature
+
     PERSISTENT_ENV="${ARYN_STATUSDIR}/persistent_env"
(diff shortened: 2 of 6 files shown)


opensearch/2.11/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Repo name: arynai/sycamore-opensearch
+
+ARG TAG=2.11.0
```
austintlee, line 3: Isn't `2.11.1` the latest 2.11 release?

opensearch/sycamore-opensearch.sh
```diff
@@ -226,6 +230,78 @@ END
...
+  "function_name": "TEXT_SIMILARITY",
+  "model_group_id": "${MODEL_GROUP_ID}",
+  "model_content_hash_value": "04157d66d847d08b3d2b51ad36cf0e1fb82afadb8086212a1d2bac2b7d6fe08a",
+  "model_config": {
+    "model_type": "roberta",
+    "embedding_dimension": 1,
+    "framework_type": "huggingface_transformers",
+    "all_config": ${all_config}
+  },
+  "url": "https://aryn-public.s3.amazonaws.com/models/BAAI/bge-reranker-base-quantized-2.zip"
```
austintlee, line 288: Are we not using the one hosted by OpenSearch (assuming they will host it)?

opensearch/sycamore-opensearch.sh
```diff
@@ -422,9 +498,34 @@ sp_create_non_rag_pipeline() {
 END
 }
 
+flick_rag_feature() {
+    # This exists because the rag_pipeline_feature_enabled setting
+    # only recognizes when it changes. Since it starts in the 'on'
+    # position now, we have to turn it off and then on again
+    _curl_json -X PUT "${BASE_URL}/_cluster/settings" \
```
austintlee, line 505: I should probably open an issue to fix this in ml-commons?

opensearch/sycamore-opensearch.sh
```diff
@@ -226,6 +230,78 @@ END
...
+  "output_past": true,
+  "pad_token_id": 1,
+  "position_embedding_type": "absolute",
+  "torch_dtype": "float32",
+  "transformers_version": "4.33.3",
+  "type_vocab_size": 1,
+  "use_cache": true,
+  "vocab_size": 250002
+}
+END
```
austintlee, line 267: How often do you think we will change this model?  I assume most of these parameters are specific to the choice of the model?

opensearch/2.11/Dockerfile
HenryL27: I guess. This is just a copy of what was previously in the quickstart

opensearch/sycamore-opensearch.sh
HenryL27: They're just hosting MiniLM-L-6-v2 and MiniLM-L-12-v2. I tried to get Dhrubo to take this one but he didn't want to. This one should be better tho. But we should run 'em all through our benchmarks

opensearch/sycamore-opensearch.sh
HenryL27: if you would be so kind. Sorry, I should have caught this when I turned the feature flag on

opensearch/sycamore-opensearch.sh
HenryL27: These params are def specific to the choice of model. On the other hand, they don't matter whatsoever. I've run tests with `"all_config": "nobody cares"`

PR #251, opened 2024-02-23
Add debug facility to sycamore-opensearch.sh entrypoint script.

alexaryn:
If $DEBUG is set to 1, debug messages will be emitted.
If $NOEXIT is set to 1, the script will sleep instead of exiting, keeping the container alive.


HenryL27:
this just gives us a debug flag, doesn't fix all our problems, yes?

HenryL27 (approved):

PR #256, opened 2024-02-23
By default, disable SSL for Jupyter, to avoid browser cert complaints.

alexaryn:
Avoid complaints from browsers about self-signed certificates.
Set SSL=1 for HTTPS
Set SSL=0 for HTTP
Docker build defaults to SSL=0

### DIFF
--- a/jupyter/Dockerfile
+++ b/jupyter/Dockerfile
@@ -18,7 +18,7 @@ RUN mkdir -p /app/work/docker_volume /app/work/bind_dir /app/work/examples
 RUN touch /app/work/AAA_SEE_README_FOR_PERSISTENT_DATA_DIRECTORIES
 COPY jupyter/run-jupyter.sh ./
 COPY jupyter/README.md ./work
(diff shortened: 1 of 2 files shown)


jupyter/Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN mkdir -p /app/work/docker_volume /app/work/bind_dir /app/work/examples
 RUN touch /app/work/AAA_SEE_README_FOR_PERSISTENT_DATA_DIRECTORIES
 COPY jupyter/run-jupyter.sh ./
 COPY jupyter/README.md ./work
-COPY notebooks/jupyter_dev_example.ipynb notebooks/default-prep-script.ipynb ./work/examples
+COPY notebooks/jupyter_dev_example.ipynb notebooks/default-prep-script.ipynb ./work/examples/
```
alexaryn, line 21: The trailing slash is required.  Was broken in a previous commit.

PR #258, opened 2024-02-23
Improve recall metrics

baitsguy:
This change let's you handle input datapoints where there are multiple page numbers. It returns 2 types of page_recall metrics, 1 where all pages are retrieved, and 1 where there is partial success.

### DIFF
--- a/sycamore/evaluation/datasets.py
+++ b/sycamore/evaluation/datasets.py
@@ -1,15 +1,19 @@
-from typing import Union, Optional, Callable
+from typing import Union, Optional, Callable, Any
 
 from datasets import IterableDataset
 from ray.data import Dataset, from_huggingface
 
+from sycamore.evaluation import EvaluationDataPoint
 from sycamore import DocSet, Context
 from sycamore.scans import MaterializedScan
 
 
 class HuggingFaceScan(MaterializedScan):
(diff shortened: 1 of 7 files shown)


sycamore/evaluation/datasets.py
```diff
@@ -1,30 +1,23 @@
...
-from sycamore.evaluation import EvaluationDataPoint
-from sycamore.scans import MaterializedScan
 from sycamore import DocSet, Context
+from sycamore.scans import MaterializedScan
 
 
 class HuggingFaceScan(MaterializedScan):
-    def __init__(self, dataset: Union[Dataset, IterableDataset], field_mapping: dict[str, str], **resource_args):
+    def __init__(
+        self, dataset: Union[Dataset, IterableDataset], doc_extractor: Optional[Callable] = None, **resource_args
```
HenryL27, line 12: Can I get a type for doc_extractors. Also what happens if its none?

HenryL27 (approved):
one potential issue, otherwise lgtm

sycamore/evaluation/datasets.py
baitsguy: I'll change it to required for now, and specify types for doc_extractor (just a function)

PR #260, opened 2024-02-26
Increase default model deployment stability

HenryL27:
This handles a number of errors we were previously failing on
- when using both torch script and onnx models, funky pytorch internal methods get into a bad state and crash. Only use ONNX.
- when restarting the node, models are often redeployed automatically. Check that they are not deployed before trying to deploy them again (in sp_setup_transient)
- sometimes the deploy task fails due to ml memory circuit breaker tripping. Detect this, wait a couple seconds for GC to run, and then try again
- sometimes the deploy task fails due to duplicate deploy task. Detect this and watch the task we should be watching. If there is no running task, try to deploy again.

Note: We still don't handle the case where the memory circuit breaker trips during the register step

Another note: There's some cleaning up to do here but if y'all could check this out and tell me how it breaks that would be good

### DIFF
--- a/opensearch/sycamore-opensearch.sh
+++ b/opensearch/sycamore-opensearch.sh
@@ -403,7 +403,7 @@ deploy_model() {
     # Create deploy task
     local deploy_model_log_file="${ARYN_STATUSDIR}/curl.better_deploy_model_task.${name}"
     spawn_deploy_model_task "${name}" "${model_id}"
-    local deploy_task_id="$(jq -r '.task_id' "${deploy_model_log_file}")"
+    local deploy_task_id=$(jq -r '.task_id' "${deploy_model_log_file}")
 
     # Cycle on the task status
     debug "Wait for deploy task to finish"
@@ -413,7 +413,8 @@ deploy_model() {
     local max_reps=60
     for i in $(seq "${max_reps}"); do
         get_deploy_status_and_act_on_it "${deploy_task_id}" "${name}" "${model_id}" "${i}" "${max_reps}" && return 0
-        deploy_task_id="$(jq -r '.task_id' "${deploy_model_log_file}")"
+        deploy_task_id=$(jq -r '.task_id' "${deploy_model_log_file}")
+        sleep 1
     done
     die "Out of time to deploy model ${name}"
 }
@@ -432,7 +433,6 @@ get_deploy_status_and_act_on_it() {
     # Case 1: RUNNING / state not found. Task is still running so wait.
     if [[ "${status}" == 'null' || "${status}" == 'RUNNING' || "${status}" == 'CREATED' ]]; then
         info "Waiting for ${name} to deploy... ${i}/${max_reps}"
-        sleep 1
         return 1
     # Case 2: COMPLETED. Task is completed, so exit
     elif [[ "${status}" == 'COMPLETED' ]]; then
@@ -456,10 +456,10 @@ handle_deploy_error() {
     local deploy_status_file="${ARYN_STATUSDIR}/curl.deploy_task_status.${name}"
     # handle error
     debug "Deploy task failed for ${name}"
-    local error="$(jq -r '.error' "${deploy_status_file}")"
+    local error=$(jq -r '.error' "${deploy_status_file}")
     debug "${error}"
-    local worker_node="$(jq -r '.worker_node[0]' "${deploy_status_file}")"
-    local error_message="$(jq -r ".\"${worker_node}\"" <<< ${error})"
+    local worker_node=$(jq -r '.worker_node[0]' "${deploy_status_file}")
+    local error_message=$(jq -r ".\"${worker_node}\"" <<< ${error})
     debug "${error_message}"
     info "Deploy task failed for ${name}: ${error_message}"
     # Memory Circuit Breaker error: wait 20 seconds and the try to deploy again
@@ -498,7 +498,6 @@ END
             model_is_deployed "${model_id}" && return 0
             info "${model_id} failed to deploy. Try again after 1 sec"
             spawn_deploy_model_task "${name}" "${model_id}"
-            sleep 1
         else
             deploy_task_id="$(jq -r '.hits.hits[0]._id' "${deploy_task_search_file}")"
             echo "{\"task_id\":\"${deploy_task_id}\"}" > ${deploy_model_log_file}
@@ -507,7 +506,6 @@ END
     # OrtEnvironment (thread pool) issue: try again and wait a sec
     elif [[ "${error_message}" = *OrtEnvironment* ]]; then
         spawn_deploy_model_task "${name}" "${model_id}"
-        sleep 1
     else
         die "Unknown error message: ${error_message}. Failing"
     fi
@@ -672,9 +670,12 @@ setup_transient() {
 model_is_deployed() {
     local model_id="$1"
     debug "checking whether ${model_id} is deployed"
-    local model_state="$(_curl "${BASE_URL}/_plugins/_ml/models/${model_id}" | jq -r '.model_state')"
-    debug "${model_state}"
+    local model_info=$(_curl "${BASE_URL}/_plugins/_ml/models/${model_id}")
+    [[ -z ${model_info} ]] && return 1
+    debug "" "${model_info}"
+    local model_state=$(jq -r '.model_state' <<< ${model_info})
     [[ -n ${model_state} && ${model_state} = DEPLOYED ]] && return 0
+    return 1
 }
 
 create_certificates() {
--- a/opensearch/sycamore-opensearch.sh
+++ b/opensearch/sycamore-opensearch.sh
@@ -220,7 +220,7 @@ END
     EMBEDDING_TASK_ID="${id}"
     echo "EMBEDDING_TASK_ID='${EMBEDDING_TASK_ID}'" >>"${PERSISTENT_ENV_TMP}"
 
-    deploy_model "" "${EMBEDDING_TASK_ID}" "embedding"
+    deploy_model_better "" "${EMBEDDING_TASK_ID}" "embedding"
     EMBEDDING_MODEL_ID="${MODEL_ID}"
     echo "EMBEDDING_MODEL_ID=${EMBEDDING_MODEL_ID}" >>"${PERSISTENT_ENV_TMP}"
 }
@@ -292,7 +292,7 @@ END
     GET_TASK_ID="${RERANKING_TASK_ID}"
     wait_or_die get_task "reranking model to register" 240
 
-    deploy_model "" "${RERANKING_TASK_ID}" "reranking"
+    deploy_model_better "" "${RERANKING_TASK_ID}" "reranking"
     RERANKING_MODEL_ID="${MODEL_ID}"
     echo "RERANKING_MODEL_ID=${RERANKING_MODEL_ID}" >>"${PERSISTENT_ENV_TMP}"
 }
@@ -301,7 +301,7 @@ sp_setup_openai_model() {
     [[ -z "${OPENAI_API_KEY}" ]] && die "No OPENAI_API_KEY set"
     wait_or_die sp_create_openai_connector 'creation of OpenAI connector' 60
     sp_register_openai_model
-    deploy_model "" "${OPENAI_TASK_ID}" "OpenAI"
+    deploy_model_better "" "${OPENAI_TASK_ID}" "OpenAI"
     OPENAI_MODEL_ID="${MODEL_ID}"
     echo "OPENAI_MODEL_ID=${OPENAI_MODEL_ID}" >>"${PERSISTENT_ENV_TMP}"
 }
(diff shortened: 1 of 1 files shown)


opensearch/sycamore-opensearch.sh
```diff
@@ -411,6 +375,123 @@ END
     echo "OPENAI_TASK_ID=${OPENAI_TASK_ID}" >>"${PERSISTENT_ENV_TMP}"
 }
 
+deploy_model_better() {
```
alexaryn, line 378: This function is very long and would benefit from breaking it up into smaller semantically focused functions.

alexaryn (changes requested):
Requesting changes due to the long, complicated function.  The bigger question, however, is "Does it work?".  I can't tell by reading it.

alexaryn:
By the way, it seems to deploy successfully for me on my laptop.

opensearch/sycamore-opensearch.sh
```diff
@@ -411,6 +375,123 @@ END
...
+            # handle error
+            debug "Deploy task failed for ${name}"
+            local error="$(jq -r '.error' "${deploy_status_file}")"
+            debug "${error}"
+            local worker_node="$(jq -r '.worker_node[0]' "${deploy_status_file}")"
+            local error_message="$(echo "${error}" | jq -r ".${worker_node}")"
+            debug "${error_message}"
+            echo "Deploy task failed for ${name}: ${error_message}"
+            # Memory Circuit Breaker error: wait 20 seconds and the try to deploy again
+            if [[ "${error_message}" = "Memory Circuit Breaker is open, please check your resources!" ]]; then
```
eric-anderson, line 428: this is fragile unless opensearch guarantees error messages won't change.  I'd suggest
```
if [[ "${error_message}" = *Memory*Circuit* ]]; then
```
which will tolerate all the messages that are similar.

opensearch/sycamore-opensearch.sh
```diff
@@ -411,6 +375,123 @@ END
...
+                local wait_time=20
+                echo "Waiting for ${wait_time}s while GC runs before retrying deploy"
+                for j in $(seq "${wait_time}"); do
+                    echo -n "."
```
eric-anderson, line 432: This works poorly with docker compose, it forces newlines.
But since you're not checking to see if it's fixed along the way, you might as well just do:
sleep 20.

Is there a reason to not do a shorter sleep and retry from the top? If you set max reps to 60 and sleep 5 here, then it would wait 300s in the face of circuit breaker issues.  The advantage of the shorter sleep is less waiting to get started.

opensearch/sycamore-opensearch.sh
```diff
@@ -411,6 +375,123 @@ END
...
+        # Case 1: RUNNING / state not found. Task is still running so wait.
+        if [[ "${status}" == 'null' || "${status}" == 'RUNNING' || "${status}" == 'CREATED' ]]; then
+            echo "Waiting for ${name} to deploy... ${i}/${max_reps}"
+            sleep 1
+        # Case 2: COMPLETED. Task is completed, so exit
+        elif [[ "${status}" == 'COMPLETED' ]]; then
+            echo "Deployed ${name} successfully"
+            MODEL_ID="${model_id}"
+            return 0
+        # Case 3: FAILED. Handle Memory Circuit Breaker error by waiting a bit but otherwise fail out
```
eric-anderson, line 417:
```
# Case 3: FAILED. Handle some error cases, fail in unrecognized ones.
```

opensearch/sycamore-opensearch.sh
```diff
@@ -411,6 +375,123 @@ END
...
+        }
+    },
+    "sort": [{"create_time": {"order": "DESC"}}]
+}
+END
+                if [[ $(jq -r '.hits.total.value' "${deploy_task_search_file}") ]]; then
+                    debug "No running deploy task for ${model_id}. => check whether it's deployed"
+                    model_is_deployed "${model_id}" && return 0
+                    echo "${model_id} failed to deploy. Try again after 1 sec"
+                    sleep 1
```
eric-anderson, line 467: What is this sleep doing?  I would think it should be after the spawn if anything, but right now the system is just stuck so you're waiting when no progress can be made.

eric-anderson (changes requested):

opensearch/sycamore-opensearch.sh
HenryL27: otherwise I can spin creating deploy tasks that fail due to duplication with 0 delay

HenryL27: but yeah, bump to post-spawn

opensearch/sycamore-opensearch.sh
```diff
@@ -423,9 +386,136 @@ deploy_model() {
     if [[ -z "${model_id}" ]]; then
         wait_task "${task_id}" "$name model to become ready"
         model_id="${GET_TASK_MODEL_ID}"
-        echo "Fetched $name model id $model_id for task_id $task_id"
+        echo "Fetched $name model id $model_id for task_id $task_id" >&2
```
eric-anderson, line 389: I'll recommend changing all of these to 
status "Fetched $name model id $model_id for task_id $task_id"
and writing a 
```
status() {
  echo "$@" >&2
}
```
function.
Requiring people to remember to put >&2 at the end of each line is very error-prone.
```
grep echo sycamore-opensearch.sh | grep -v '>>' | grep -v '>'
```
will find all the ones that were missed.

opensearch/sycamore-opensearch.sh
```diff
@@ -423,9 +386,136 @@ deploy_model() {
...
+            sleep 1
+            continue
+        # Case 2: COMPLETED. Task is completed, so exit
+        elif [[ "${status}" == 'COMPLETED' ]]; then
+            echo "Deployed ${name} successfully" >&2
+            MODEL_ID="${model_id}"
+            return 0
+        # Case 3: FAILED. Handle some error cases, fail in unrecognized ones.
+        elif [[ "${status}" == 'FAILED' ]]; then
+            handle_deploy_error "${model_id}" "${name}" && return 0
```
eric-anderson, line 421: I think you want `&& return 1` Near as I can tell when handle_deploy_error fails it's an error.

opensearch/sycamore-opensearch.sh
```diff
@@ -423,9 +386,136 @@ deploy_model() {
...
+    local model_id="$2"
+    debug "Create deploy task for model: ${name}"
+    local deploy_model_log_file="${ARYN_STATUSDIR}/curl.better_deploy_model_task.${name}"
+    _curl -X POST "${BASE_URL}/_plugins/_ml/models/${model_id}/_deploy" \
+          -o "${deploy_model_log_file}"
+    debug "" "${deploy_model_log_file}"
+}
+
+
+deploy_model_old() {
```
eric-anderson, line 506: Can we nuke this function now?

opensearch/sycamore-opensearch.sh
```diff
@@ -570,15 +660,26 @@ setup_transient() {
     _curl "${BASE_URL}/_cluster/settings" \
     | grep -Fq aryn_deploy_complete && die "aryn_deploy_complete already set"
 
-    deploy_model "${EMBEDDING_MODEL_ID}" "${EMBEDDING_TASK_ID}" "embedding"
-    deploy_model "${OPENAI_MODEL_ID}" "${OPENAI_TASK_ID}" "OpenAI"
-    deploy_model "${RERANKING_MODEL_ID}" "${RERANKING_TASK_ID}" "reranking"
+    model_is_deployed "${EMBEDDING_MODEL_ID}" ||
+        deploy_model "${EMBEDDING_MODEL_ID}" "${EMBEDDING_TASK_ID}" "embedding"
```
eric-anderson, line 664: I think we should we modify deploy_model to start with
```
if model_is_deployed "$1"; then
  status "Model $1 for $3 already deployed"
  return 0
fi
```
Then these duplicated lines simplify down.

eric-anderson (approved):

opensearch/sycamore-opensearch.sh
HenryL27: the logic I was going for was: when handle_deploy_error succeeds, the model is deployed already. When it's an unrecoverable error, handle_deploy_error dies. When handle_deploy_error fails, then it's a handled error and we can try again

opensearch/sycamore-opensearch.sh
```diff
@@ -423,15 +390,138 @@ deploy_model() {
...
+    local max_reps="$5"
+    local deploy_model_log_file="${ARYN_STATUSDIR}/curl.better_deploy_model_task.${name}"
+
+    _curl "${BASE_URL}/_plugins/_ml/tasks/${deploy_task_id}" -o "${deploy_status_file}"
+    local status="$(jq -r '.state' "${deploy_status_file}")"
+    debug "${status}"
+    # Case 1: RUNNING / state not found. Task is still running so wait.
+    if [[ "${status}" == 'null' || "${status}" == 'RUNNING' || "${status}" == 'CREATED' ]]; then
+        info "Waiting for ${name} to deploy... ${i}/${max_reps}"
+        sleep 1
```
alexaryn, line 435: Can we move the sleep to L416 above?  That would put the higher-level control all in one place.

opensearch/sycamore-opensearch.sh
```diff
@@ -423,15 +390,138 @@ deploy_model() {
...
+    local model_id="$1"
+    local name="$2"
+    local deploy_model_log_file="${ARYN_STATUSDIR}/curl.better_deploy_model_task.${name}"
+    local deploy_status_file="${ARYN_STATUSDIR}/curl.deploy_task_status.${name}"
+    # handle error
+    debug "Deploy task failed for ${name}"
+    local error="$(jq -r '.error' "${deploy_status_file}")"
+    debug "${error}"
+    local worker_node="$(jq -r '.worker_node[0]' "${deploy_status_file}")"
+    local error_message="$(jq -r ".\"${worker_node}\"" <<< ${error})"
```
alexaryn, line 462: nit: can just have $(...)

opensearch/sycamore-opensearch.sh
```diff
@@ -423,15 +390,138 @@ deploy_model() {
...
+    info "Deploy task failed for ${name}: ${error_message}"
+    # Memory Circuit Breaker error: wait 20 seconds and the try to deploy again
+    if [[ "${error_message}" = *Memory*Circuit*Breaker* ]]; then
+        local wait_time=5
+        info "Waiting for ${wait_time}s while GC runs before retrying deploy"
+        sleep ${wait_time}
+        spawn_deploy_model_task "${name}" "${model_id}"
+    # Duplicate Deploy Task error: set task id I'm watching to the RUNNING task
+    # - if no running task, then either the model is deployed or we try again
+    elif [[ "${error_message}" = "Duplicate deploy model task" ]]; then
```
alexaryn, line 473: It's too bad we have to rely on specific text of error messages so much.

opensearch/sycamore-opensearch.sh
```diff
@@ -423,15 +390,138 @@ deploy_model() {
...
+    },
+    "sort": [{"create_time": {"order": "DESC"}}]
+}
+END
+        if [[ $(jq -r '.hits.total.value' "${deploy_task_search_file}") ]]; then
+            debug "No running deploy task for ${model_id}. => check whether it's deployed"
+            model_is_deployed "${model_id}" && return 0
+            info "${model_id} failed to deploy. Try again after 1 sec"
+            spawn_deploy_model_task "${name}" "${model_id}"
+            sleep 1
```
alexaryn, line 501: It looks like we generally sleep when we return 1 and the caller generally sleeps.  I know there are cases where we don't need to sleep, but they seem rare.  It may be more robust and cleaner to just have the only sleep up at L416 in the for loop.

opensearch/sycamore-opensearch.sh
```diff
@@ -579,6 +669,14 @@ setup_transient() {
     '{"transient":{"cluster":{"metadata":{"aryn_deploy_complete":1}}}}'
 }
 
+model_is_deployed() {
+    local model_id="$1"
+    debug "checking whether ${model_id} is deployed"
+    local model_state="$(_curl "${BASE_URL}/_plugins/_ml/models/${model_id}" | jq -r '.model_state')"
```
alexaryn, line 675: May want to split this out so we can debug-print the curl output.
```
  local model_info=$(_curl ...)
  [[ -z ${model_info} ]] && return 1
  debug "" "${model_info}"
  local model_state=$(jq -r '.model_state' <<< ${model_info})
```
Can probably drop the other debug then.

opensearch/sycamore-opensearch.sh
```diff
@@ -579,6 +669,14 @@ setup_transient() {
...
+    debug "${model_state}"
+    [[ -n ${model_state} && ${model_state} = DEPLOYED ]] && return 0
+}
+
```
alexaryn, line 679: Seems like we should end with an explicit `return 1`

alexaryn (approved):
Looks much better to me.  I trust you to address the remaining items.

PR #265, opened 2024-02-28
Modify supplement_text for integrating text from pdfminer

bohou-aryn:


baitsguy (approved):
Can you add the description in the PR summary

PR #269, opened 2024-03-05
Bump opensearch ssl startup wait time to 30 tries.

eric-anderson:
It takes 12/15 on my laptop semi-regularly, and we want to have a bit more headroom for people on slower machines.


alexaryn (approved):
Yay!

PR #275, opened 2024-03-06
Fix notebooks -- proper protocol, truncate output.

eric-anderson:
* Pass SSL env var in via docker so people could enable SSL.
* Update jupyter dev example to truncate content and limit output -- running the notebook was getting tons of output.
* Update jupyter dev example to print the right URL.
* Update show operation in docset to not show binary in elements and truncate content in elements if we would do that in the main document. Slightly refactor to avoid duplicated code.

### DIFF
--- a/notebooks/jupyter_dev_example.ipynb
+++ b/notebooks/jupyter_dev_example.ipynb
@@ -3,6 +3,17 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "id": "1867d47c-b124-4ae2-93c1-390d9c8578c7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%load_ext autoreload\n",
+    "%autoreload 2"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "id": "e1666a1c-8c22-4643-8e15-a9a9f863893f",
    "metadata": {},
    "outputs": [],
@@ -206,7 +217,7 @@
    "outputs": [],
    "source": [
     "merged_docset = entity_docset.merge(GreedyTextElementMerger(tokenizer=HuggingFaceTokenizer(\"sentence-transformers/all-MiniLM-L6-v2\"), max_tokens=512))\n",
-    "merged_docset.show(show_binary = False)"
+    "merged_docset.show(show_binary = False, show_elements = False)"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


notebooks/jupyter_dev_example.ipynb
```diff
@@ -1,5 +1,16 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1867d47c-b124-4ae2-93c1-390d9c8578c7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%load_ext autoreload\n",
+    "%autoreload 2"
+   ]
+  },
```
HenryL27, line 13: what does this do? do I care?

bsowell (approved):
Can't speak to the SSL compose change, but the notebook/library changes look good.

notebooks/jupyter_dev_example.ipynb
```diff
@@ -285,8 +296,19 @@
...
+    "print(f\"Visit {protocol}://localhost:3000 and use the {index} index to query these results in the UI\")"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a45cd0e8-caad-45c8-a7db-57483a006d4b",
+   "metadata": {},
+   "outputs": [],
+   "source": []
```
HenryL27, line 311: empty cell?

sycamore/docset.py
```diff
@@ -90,16 +97,23 @@ def show(
...
+            if show_elements and "elements" in document.data:
+                if not show_binary:
+                    for i, e in enumerate(document.data["elements"]):
+                        if e.get("binary_representation") is not None:
+                            binary_length = len(e["binary_representation"])
+                            e["binary_representation"] = f"<{binary_length} bytes>".encode("utf-8")
+                if truncate_content:
+                    for i, e in enumerate(document.data["elements"]):
+                        if e.get("text_representation") is not None:
+                            e["text_representation"] = _truncate(e["text_representation"])
```
HenryL27, line 115: I guess this works bc docsets are copied?

HenryL27 (approved):

notebooks/jupyter_dev_example.ipynb
eric-anderson: It means that if you change the sycamore library it will automatically be reloaded in the jupyter notebook. Doesn't hurt anything if you don't. Was critical for making the show fix.

notebooks/jupyter_dev_example.ipynb
eric-anderson: Removed.

sycamore/docset.py
eric-anderson: Per discussion, yes, it's a little weird but it's the same as other code.

PR #276, opened 2024-03-07
 More SSL/container fixes.

eric-anderson:
* In order to disable SSL by default, the SSL variable has to be set to 0 all other values enable
  SSL.

* Make the notebooks consistent with this and print the proper URL depending on whether we are
  using SSL.

* Copy over the opensearch host auto-guessing logic from default-prep-script.ipynb


.env
```diff
@@ -10,5 +10,6 @@ UI_PORT=3000
 OPENSEARCH_PORT=9200
 RAY_CONSOLE_PORT=8265
 JUPYTER_PORT=8888
+SSL=0
```
alexaryn, line 13: What is the top-level .env file and what does it do?  Seems like a dangerous default setting for SSL.

docs/source/APIs/gen
```diff
@@ -31,7 +31,7 @@ def doFile(name, dir, ent):
         top = ast.parse(fp.read())
 
     ary = []
-    base = ent[ : -3]
+    base = ent[:-3]
```
alexaryn, line 34: I always disagree with black on this one.  Colon is a binary operator.  There should be spaces around it.  And also, I didn't think this file was under auto/enforced formatting.  GitHub doesn't seem to think it is.

docs/source/APIs/gen
```diff
@@ -42,9 +42,7 @@ def doFile(name, dir, ent):
             line = "=" * len(title)
             fp.write(f"{title}\n{line}\n\n")
             for sym in sorted(ary):
-                fp.write(f".. autoclass:: {sym}\n" +
-                          "   :members:\n" +
-                          "   :show-inheritance:\n")
+                fp.write(f".. autoclass:: {sym}\n" + "   :members:\n" + "   :show-inheritance:\n")
```
alexaryn, line 45: Does black really thing this is better?  It could at least concatenate the strings.

notebooks/default-prep-script.ipynb
```diff
@@ -284,7 +284,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.6"
+   "version": "3.11.2"
```
alexaryn, line 287: Downgrade?

notebooks/jupyter_dev_example.ipynb
```diff
@@ -252,8 +252,16 @@
    "source": [
     "index = \"local_development_example_index_withentity\" # You can change this to something else if you'd like\n",
     "\n",
-    "os_client_args = {\n",
-    "        \"hosts\": [{\"host\": \"localhost\", \"port\": 9200}],\n",
+    "if os.path.exists(\"/app/work/bind_dir\"):\n",
```
alexaryn, line 255: I think the canonical way to check is to look for the file /.dockerenv but any of these methods can be fooled.

notebooks/jupyter_dev_example.ipynb
```diff
@@ -319,7 +329,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.6"
+   "version": "3.11.2"
```
alexaryn, line 332: Is this setting a minimum version?

alexaryn (approved):
Seems OK, but I have some questions.

.env
eric-anderson: It sets the default environment when you run a docker compose.  If we don't do this, then people will get https if they don't do it manually, and Jon said we had to http.  I'm assuming other stuff won't read this.

docs/source/APIs/gen
eric-anderson: pre-commit does this:
```
        poetry run pre-commit run --all-files
```

notebooks/default-prep-script.ipynb
eric-anderson: it's just what jupyter does.  Per discussion leaving alone.

PR #280, opened 2024-03-08
build ml-commons locally with correct dependencies

HenryL27:
ml-commons figures out its onnxruntime dependency by looking at the operating system at build time, which doesn't quite do the correct thing, which in turn makes onnx fail on arm.
this pr fixes that, though we probably need to make it a less hacky fix

### DIFF
--- a/opensearch/Dockerfile
+++ b/opensearch/Dockerfile
@@ -9,6 +9,8 @@ RUN git clone https://github.com/opensearch-project/ml-commons.git
 WORKDIR /build_ml/ml-commons
 RUN git checkout -b 2.12 origin/2.12
 RUN sed -i 's#if (os.macOsX) {#if (System.getProperty("os.arch") == "aarch64") {#' ml-algorithms/build.gradle
+RUN sed -i 's#throw new OpenSearchException("GenerativeQAResponseProcessor failed in precessing response");#log.error(e); throw new OpenSearchException(e);#'\
+    search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
 RUN ./gradlew assemble
 
 FROM opensearchproject/opensearch:$TAG as install_ml_commons
@@ -16,6 +18,7 @@ FROM opensearchproject/opensearch:$TAG as install_ml_commons
 USER opensearch
 COPY --chown=opensearch:opensearch --from=build_ml_commons \
     /build_ml/ml-commons/plugin/build/distributions/opensearch-ml-2.12.0.0-SNAPSHOT.zip /tmp/
+# opensearch-skills extends ml-commons, so we need to remove it before we can mess with ml commons.
 RUN /usr/share/opensearch/bin/opensearch-plugin remove opensearch-skills
 RUN /usr/share/opensearch/bin/opensearch-plugin remove opensearch-ml
 RUN /usr/share/opensearch/bin/opensearch-plugin install --batch file:///tmp/opensearch-ml-2.12.0.0-SNAPSHOT.zip \
--- a/opensearch/Dockerfile
+++ b/opensearch/Dockerfile
@@ -1,7 +1,30 @@
 # Repo name: arynai/sycamore-opensearch
 
 ARG TAG=2.12.0
-FROM opensearchproject/opensearch:$TAG
+
+FROM gradle:jdk17 as build_ml_commons
+
+WORKDIR /build_ml
+RUN git clone https://github.com/opensearch-project/ml-commons.git
+WORKDIR /build_ml/ml-commons
+RUN git checkout -b 2.12 origin/2.12
+RUN sed -i 's#if (os.macOsX) {#if (true) {#' ml-algorithms/build.gradle
+RUN sed -i 's#throw new OpenSearchException("GenerativeQAResponseProcessor failed in precessing response");#log.error(e); throw new OpenSearchException(e);#'\
+    search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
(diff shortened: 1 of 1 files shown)


opensearch/Dockerfile
```diff
@@ -1,7 +1,30 @@
...
 ARG TAG=2.12.0
-FROM opensearchproject/opensearch:$TAG
+
+FROM gradle:jdk17 as build_ml_commons
+
+WORKDIR /build_ml
+RUN git clone https://github.com/opensearch-project/ml-commons.git
+WORKDIR /build_ml/ml-commons
+RUN git checkout -b 2.12 origin/2.12
+RUN sed -i 's#if (os.macOsX) {#if (true) {#' ml-algorithms/build.gradle
```
eric-anderson, line 11: I don't think we want this. If we're amd we want the _gpu one.
```
def arch = System.getProperty("os.arch")
if (arch == "aarch64") { ... }
```
might be the right thing.
https://github.com/apereo/cas/blob/bbfb3a9b4e95c17a45a72bd6e199343790755c17/build.gradle#L143

opensearch/Dockerfile
```diff
@@ -1,7 +1,30 @@
...
+RUN sed -i 's#throw new OpenSearchException("GenerativeQAResponseProcessor failed in precessing response");#log.error(e); throw new OpenSearchException(e);#'\
+    search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
+RUN ./gradlew assemble -DoperatingSystem=macOsX
```
eric-anderson, line 14: Remove the -D or make it conditional?

opensearch/Dockerfile
```diff
@@ -1,7 +1,30 @@
...
+FROM opensearchproject/opensearch:$TAG as install_ml_commons
+
+USER opensearch
+COPY --chown=opensearch:opensearch --from=build_ml_commons \
+    /build_ml/ml-commons/plugin/build/distributions/opensearch-ml-2.12.0.0-SNAPSHOT.zip /tmp/
+RUN /usr/share/opensearch/bin/opensearch-plugin remove opensearch-skills
+RUN /usr/share/opensearch/bin/opensearch-plugin remove opensearch-ml
+RUN /usr/share/opensearch/bin/opensearch-plugin install --batch file:///tmp/opensearch-ml-2.12.0.0-SNAPSHOT.zip \
+    https://repo1.maven.org/maven2/org/opensearch/plugin/opensearch-skills/2.12.0.0/opensearch-skills-2.12.0.0.zip
+RUN rm /tmp/opensearch-ml-2.12.0.0-SNAPSHOT.zip
```
eric-anderson, line 25: I'd remove this rm, near as I can tell from the Docker stuff, it won't do anything. The only way to not include files in layers is to create and remove them in the same layer.

opensearch/Dockerfile
```diff
@@ -1,7 +1,30 @@
...
+RUN sed -i 's#throw new OpenSearchException("GenerativeQAResponseProcessor failed in precessing response");#log.error(e); throw new OpenSearchException(e);#'\
+    search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
+RUN ./gradlew assemble -DoperatingSystem=macOsX
+
+FROM opensearchproject/opensearch:$TAG as install_ml_commons
+
+USER opensearch
+COPY --chown=opensearch:opensearch --from=build_ml_commons \
+    /build_ml/ml-commons/plugin/build/distributions/opensearch-ml-2.12.0.0-SNAPSHOT.zip /tmp/
+RUN /usr/share/opensearch/bin/opensearch-plugin remove opensearch-skills
```
eric-anderson, line 21: Why are we removing opensearch-skills and then re-installing it?

HenryL27: it extends ml commons, so the opensearch-plugin tool won't let me remove ml-commons without first removing skills

eric-anderson: optional: add a comment to explain.

opensearch/Dockerfile
```diff
@@ -1,7 +1,27 @@
...
-FROM opensearchproject/opensearch:$TAG
+
+FROM gradle:jdk17 as build_ml_commons
+
+WORKDIR /build_ml
+RUN git clone https://github.com/opensearch-project/ml-commons.git
+WORKDIR /build_ml/ml-commons
+RUN git checkout -b 2.12 origin/2.12
+RUN sed -i 's#if (os.macOsX) {#if (System.getProperty("os.arch") == "aarch64") {#' ml-algorithms/build.gradle
+RUN ./gradlew assemble
```
eric-anderson, line 12: If we're not doing the other sed, then for amd64 the plugin is unchanged.  Given that, I'd recommend:
```
RUN /mkdir /tmp/release
RUN [[ $(uname -m) == arm64 ]] && cp /build_ml/ml-commons/plugin/build/distributions/opensearch-ml-2.12.0.0-SNAPSHOT.zip /tmp/release
# uname -m is a guess, might be wrong

# then below
COPY --chown=opensearch:opensearch --from=build_ml_commons /tmp/release /tmp
COPY maybe-update-mlcommons.sh /tmp
```
with maybe-update-mlcommons.sh:
#!/bin/sh
if [[ ! -f /tmp/opensearch-ml-2.12.0.0-SNAPSHOT.zip ]]; then
   echo "No opensearch plugin to install"
   exit 0
fi
... rest of the install steps
```

HenryL27: looking like the other sed is valuable, so we'll use it. Even though it makes me very incomfortable

eric-anderson (approved):

PR #284, opened 2024-03-12
build and install remote processor plugin in opensearch dockerfile

HenryL27:
this builds and adds the RPS plugin to the opensearch dockerfile so it can use the RPS image once that gets published.
We may also want to add to the default pipeline definitions, but that can wait for @alexaryn to figure out what's going on with NDD stuff

### DIFF
--- a/opensearch/Dockerfile
+++ b/opensearch/Dockerfile
@@ -1,13 +1,10 @@
 # Repo name: arynai/sycamore-opensearch
 
-ARG TAG=2.12.0
-
 FROM gradle:jdk17 AS build_ml_commons
 
 WORKDIR /build_ml
-RUN git clone https://github.com/opensearch-project/ml-commons.git
+RUN git clone -b 2.12 --depth 1 https://github.com/opensearch-project/ml-commons.git
 WORKDIR /build_ml/ml-commons
-RUN git checkout -b 2.12 origin/2.12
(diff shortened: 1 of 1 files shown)


opensearch/Dockerfile
```diff
@@ -13,7 +13,18 @@ RUN sed -i 's#throw new OpenSearchException("GenerativeQAResponseProcessor faile
     search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
 RUN ./gradlew assemble
 
-FROM opensearchproject/opensearch:$TAG as install_ml_commons
+FROM gradle:jdk17 AS build_rps_plugin
+
+WORKDIR /build
+RUN git clone https://github.com/aryn-ai/opensearch-remote-processor.git
```
eric-anderson, line 19: `git clone -b 2.x --depth 1 https://github.com/aryn-ai/opensearch-remote-processor.git`

We don't care about history. Same fix above for ml-commons.git.  

Probably also remove the TAG variable at the top and just inline it. It's no longer safe with the 2.12 checkout for ml-commons. Alternately make a bunch of variables and apply them throughout, but since we don't test for other combinations it doesn't seem useful to support them.

eric-anderson (approved):

PR #288, opened 2024-03-14
Implement CachedTextractTableExtractor

bohou-aryn:
CachedTextractTableExtractor caches textractor table extraction result into s3 for reuse.

### DIFF
--- a/sycamore/transforms/extract_table.py
+++ b/sycamore/transforms/extract_table.py
@@ -1,6 +1,12 @@
 from abc import abstractmethod, ABC
+from collections import OrderedDict
+import io
 from typing import Optional
 
+import boto3
+from botocore.exceptions import ClientError
+import json
+import PyPDF2
 from textractor import Textractor
 from textractor.data.constants import TextractFeatures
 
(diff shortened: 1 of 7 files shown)


sycamore/transforms/partition.py
```diff
@@ -359,7 +359,7 @@ def __init__(self, model_name_or_path, threshold: float = 0.4):
         pass
 
     def partition(self, document: Document) -> Document:
-        from sycamore.transforms import SycamorePDFPartitioner
+        from sycamore.transforms.detr_partitioner import SycamorePDFPartitioner
```
bsowell, line 362: Just confirming that just importing this doesn't pull in any dependencies or anything that are platform specific? (e.g. cuda stuff).

sycamore/transforms/extract_table.py
```diff
@@ -108,3 +114,152 @@ def extract_tables(self, document: Document) -> Document:
...
+     1. if table pages not exist, leave document unchanged
+     2. if there is a cache hit for current doc, if true, return the cached extracted tables from S3
+     3. if a cache miss happens, clip pages which contain tables and run table extraction using textractor,
+      also update cache accordingly.
+
+    Cached table is in Json format as below:
+    {
+      "tables": [
+        {
+          "page_number": xx,
```
bsowell, line 133: Does textract definitely not support multi-page tables?

sycamore/transforms/extract_table.py
```diff
@@ -108,3 +114,152 @@ def extract_tables(self, document: Document) -> Document:
...
+        """Put table into S3"""
+        parts = self._s3_cache_location.replace("s3://", "").split("/", 1)
+        bucket = parts[0]
+        key = "/".join([parts[1], cache_id]) if len(parts) == 2 else cache_id
+        json_str = json.dumps(table, indent=2)
+        self._s3.put_object(Body=json_str, Bucket=bucket, Key=key)
+
+    def extract_tables(self, document: Document) -> Document:
+        # a list holding 1 based page number
+        table_pages = [element.properties["page_number"] for element in document.elements if element.type == "Table"]
```
bsowell, line 210: Do we want to have a option/fallback to the current behavior where we call for the entire doc? People who don't/can't run the DETR model (e.g. on the mac) might not have annotated labels, and caching would still be useful.

sycamore/transforms/extract_table.py
```diff
@@ -108,3 +114,152 @@ def extract_tables(self, document: Document) -> Document:
...
+    TextractTableExtractor with S3 based cache support
+
+    For each document, CachedTextractTableExtractor follows a series of steps to do table extraction
+     1. if table pages not exist, leave document unchanged
+     2. if there is a cache hit for current doc, if true, return the cached extracted tables from S3
+     3. if a cache miss happens, clip pages which contain tables and run table extraction using textractor,
+      also update cache accordingly.
+
+    Cached table is in Json format as below:
+    {
```
bsowell, line 130: So everything is in a single file? Do you have a sense of how far this will scale? For datasets with thousands of big tables, wouldn't downloading/updating the cache eventually become a bottleneck. I don't want to boil the ocean now, but curious what your thoughts are.

sycamore/transforms/partition.py
bohou-aryn: no.

sycamore/transforms/extract_table.py
bohou-aryn: not sure,  just populate whatever textrator returns, and it's int.

sycamore/transforms/extract_table.py
bohou-aryn: not sure, how big it scales on disk, but seems today textractor just put all the result in memory, I don't have a lot of concern on the file size for now.

sycamore/transforms/extract_table.py
bohou-aryn: good point, would expose a knob for this.

baitsguy (approved):

bsowell (approved):

PR #296, opened 2024-03-19
Refactor sycamore to apps, lib

HenryL27:
this PR touches like every file in the repository and also makes new files. yay. The main gist is:
- sycamore -> lib/sycamore
- crawler -> apps/crawler
- opensearch -> apps/opensearch
- jupyter -> apps/jupyter
- importer -> apps/importer

Some of these will get new pyproject.tomls. I'll expand on this more once I better grok how that works, but basically, instead of being a single poetry project, sycamore becomes several (sycamore-crawler, sycamore-ai, etc) and then at the repo root there's a pyproject.toml that depends on all of them called sycamore-monorepo

I think merging in RPS and DemoUI (->lib and apps, ->apps) will wait until this step is ready


HenryL27:
don't worry about the force-pushes, I accidentally rebased the wrong direction

HenryL27:
here's what the tree looks like after this PR
```
tree -I __pycache__ -I node_modules -I "*.py|*.ipynb|*.sh|*.yml|*.yaml|*.ttf|*.md|*.json|*.html|*.pptx|*.pdf|*.rst|*.jpg|*.png|*.svg|*.ico|*.quickstart" .

sycamore
├── LICENSE
├── apps
│   ├── crawler
│   │   ├── crawler
│   │   │   ├── http
│   │   │   │   ├── Dockerfile
│   │   │   │   ├── crawler
│   │   │   │   │   └── spiders
│   │   │   │   ├── scrapy.cfg
│   │   │   │   └── tests
│   │   │   └── s3
│   │   │       ├── Dockerfile
│   │   │       ├── crawler
│   │   │       └── tests
│   │   │           └── unit
│   │   ├── poetry.lock
│   │   └── pyproject.toml
│   ├── importer
│   │   ├── Dockerfile
│   │   ├── importer
│   │   │   └── docker
│   │   ├── poetry.lock
│   │   └── pyproject.toml
│   ├── jupyter
│   │   ├── Dockerfile
│   │   ├── bind_dir
│   │   │   └── BIND_MOUNT_BETWEEN_DOCKER_AND_HOST
│   │   └── sudoers
│   └── opensearch
│       ├── 2.11
│       │   └── Dockerfile
│       └── Dockerfile
├── docs
│   ├── Makefile
│   ├── make.bat
│   └── source
│       ├── APIs
│       │   ├── conversation_memory
│       │   ├── data_preparation
│       │   ├── gen
│       │   └── transforms
│       ├── conversation_memory
│       │   └── imgs
│       ├── data_ingestion_and_preparation
│       │   ├── imgs
│       │   │   └── test
│       │   └── transforms
│       ├── images
│       ├── querying_data
│       │   └── imgs
│       ├── tutorials
│       └── welcome_to_sycamore
│           └── imgs
│               └── test
├── examples
├── lib
│   └── sycamore
│       ├── poetry.lock
│       ├── pyproject.toml
│       └── sycamore
│           ├── data
│           ├── evaluation
│           │   └── metrics
│           ├── functions
│           ├── llms
│           │   └── prompts
│           ├── rules
│           ├── scans
│           ├── tests
│           │   ├── integration
│           │   │   ├── evaluation
│           │   │   ├── llms
│           │   │   ├── query
│           │   │   ├── textractor
│           │   │   └── transforms
│           │   ├── pytest.ini
│           │   ├── resources
│           │   │   └── data
│           │   │       ├── htmls
│           │   │       ├── json
│           │   │       ├── nested_json
│           │   │       ├── pdfs
│           │   │       └── pptx
│           │   └── unit
│           │       ├── data
│           │       ├── evaluation
│           │       ├── functions
│           │       ├── llms
│           │       ├── scans
│           │       └── transforms
│           ├── transforms
│           ├── utils
│           └── writers
├── notebooks
├── poetry.lock
└── pyproject.toml
```

lib/sycamore/pyproject.toml
```diff
@@ -0,0 +1,102 @@
...
+aiohttp = "^3.9.2"
+
+# Dependencies for building docs. Defined as an extra
+# so they can be installed using pip on RTD.
+furo = { version = "^2023.9.10", optional = true }
+myst-parser = { version = "^2.0.0", optional = true }
+sphinx = { version = "^7.2.5", optional = true }
+rouge = "^1.0.1"
+
+[tool.poetry.group.crawler_s3.dependencies]
```
bsowell, line 52: Can these go away since the crawler has its own pyproject.toml now?

.github/workflows/testing.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Install sycamore
         run: poetry install
       - name: Run tests
-        run: poetry run pytest sycamore/tests/unit/
+        run: poetry run pytest lib/sycamore/sycamore/tests/unit/
```
bsowell, line 41: The tests pass, so this must work, but just for my own comprehension I'm trying to make sure I understand the difference between this and something like `cd lib/sycamore && poetry run pytest sycamore/tests/unit/`. If I understand right, this will run in the context of the root pyproject.toml, so it will also have the crawler dependencies, etc?

lib/sycamore/sycamore/README.md
bsowell, line 1: I think we sort of rewrote README this to reflect the full stack. Should we still have it at the root? We need some kind of README at the root, and I can't quite tell from the diff what's there.

lib/sycamore/README.md
```diff
@@ -0,0 +1,63 @@
+![SycamoreLogoFinal.svg](https://raw.githubusercontent.com/aryn-ai/sycamore/main/docs/source/images/sycamore_logo.svg)
+
+[![PyPI](https://img.shields.io/pypi/v/sycamore-ai)](https://pypi.org/project/sycamore-ai/)
+[![PyPI - Python Version](https://img.shields.io/pypi/pyversions/sycamore-ai)](https://pypi.org/project/sycamore-ai/)
+[![Slack](https://img.shields.io/badge/slack-sycamore-brightgreen.svg?logo=slack)](https://join.slack.com/t/sycamore-ulj8912/shared_invite/zt-23sv0yhgy-MywV5dkVQ~F98Aoejo48Jg)
+[![Docs](https://readthedocs.org/projects/sycamore/badge/?version=stable)](https://sycamore.readthedocs.io/en/stable/?badge=stable)
```
bsowell, line 6: Do the docs still work? I'm wondering if the autodoc stuff that creates API docs from docstrings will still pick up the library. At the very least, I suspect the paths need to be updated in this gen script: https://github.com/aryn-ai/sycamore/blob/main/docs/source/APIs/gen

.github/workflows/testing.yml
HenryL27: yep, that's right. Will look to see if I can get it to run with just the dataprep lib context

bsowell: I don't necessarily know that that is required. I'm trying to think in terms of virtualenvs. Will I primarily want to operate in a single virtualenv that has everything, or switch virtualenvs based on sub-project. I guess probably the former unless we get incompatible dependencies (which we really should try to avoid).

lib/sycamore/sycamore/README.md
HenryL27: I haven't edited any READMEs; this is just a copy. We probably do need to fix them tho

lib/sycamore/README.md
HenryL27: That's a very good question.

.github/workflows/testing.yml
HenryL27: well, on the positive side, I did this and immediately found a bug where a dependency (pypdf2) was added to the root pyproject but wasn't in the lib/sycamore pyproject - and since the lib/sycamore pyproject is what will be used to build the PyPI release, we want to catch that

lib/sycamore/README.md
HenryL27: surprisingly it seems like running a `make html` correctly generated all the docs... updating paths to sycamore lib anyway

lib/sycamore/sycamore/README.md
HenryL27: Oh I think I understand this comment better. I copied the root readme to lib/sycamore/, so it's in both places. lib/sycamore/sycamore/README talks about execution and scans and stuff

bsowell (approved):

PR #298, opened 2024-03-20
Make sure filetype property is already set.

eric-anderson:
Set it to binary_format if it isn't already set.


HenryL27 (changes requested):
requesting changes here to block merging until refactor PR is pushed in case of an eager fixer or something

HenryL27 (approved):
unlocking. might wanna address my comment anyway

sycamore/scans/file_scan.py
```diff
@@ -128,6 +128,8 @@ def _to_document(self, dict: dict[str, Any]) -> dict[str, bytes]:
         if self._is_s3_scheme():
             dict["path"] = "s3://" + dict["path"]
         properties.update({"path": dict["path"]})
+        if "filetype" not in properties and self._binary_format is not None:
+            properties["filetype"] = self._binary_format
```
alexaryn, line 132: To match what was observed on the PDF side, the value would have to be 'text/html'.  We do want to be consistent in using MIME.  To do that without being hacky may be a bit harder.

alexaryn (changes requested):
I don't think this is high-priority to fix the HTML viewer bug, but we do want to have complete properties for all document types.  Inevitably, this may involve manifests.

sycamore/scans/file_scan.py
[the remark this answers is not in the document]
eric-anderson: The existing tests were looking for filetype:
```
ycamore/tests/unit/transforms/test_merge_elements.py:                    "properties": {"filetype": "text/plain", "page_number": 1},
sycamore/tests/unit/transforms/test_merge_elements.py:                    "properties": {"filetype": "text/plain", "page_number": 2},
sycamore/tests/unit/transforms/test_merge_elements.py:            "filetype": "text/plain",
sycamore/tests/unit/transforms/test_partition.py:            "metadata": {"filename": "Bert.pdf", "filetype": "application/pdf", "page_number": 1},
sycamore/tests/unit/transforms/test_partition.py:            "filetype": "application/pdf",
sycamore/tests/unit/transforms/test_spread_properties.py:                    "properties": {"filetype": "text/plain", "page_number": 1},
sycamore/tests/unit/transforms/test_spread_properties.py:                    "properties": {"filetype": "text/plain", "page_number": 2},
sycamore/tests/unit/transforms/test_spread_properties.py:            assert elem.properties["filetype"] == "text/plain"
sycamore/tests/unit/transforms/test_spread_properties.py:                assert elem.properties["filetype"] == "text/plain"
```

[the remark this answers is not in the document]
eric-anderson: I fixed the docstring.

eric-anderson: Made it consistent with MIME by translating the three formats we handle in the partitioner.

sycamore/scans/file_scan.py
```diff
@@ -129,13 +133,24 @@ def _to_document(self, dict: dict[str, Any]) -> dict[str, bytes]:
...
+        if self._binary_format == "html" or self._binary_format == "htm":
+            return "text/html"
+        if self._binary_format == "pdf":
+            return "application/pdf"
+        if self._binary_format == "pptx":
+            return "application/vnd.openxmlformats-officedocument.presentationml.presentation"
+        ret = f"application/{self._binary_format}"
+        logger.warning(f"Unrecognized extenstion {self._binary_format}; using {ret}")
+        return ret
+
```
alexaryn, line 153: It's unfortunate that we have to reinvent this wheel.  Can we use Python's mimetypes.guess_type()?  We should at least add `.ppt` and both `.doc` and `.docx`. I've seen `.doc` in the sort benchmark crawl.  It would be good to just make this table-driven.  Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types I'd suggest also adding csv, tsv (text/tab-separated-values), rtf, xls, xlsx, and xml.  There's reasonable likelihood we could ingest those.

alexaryn (approved):
Approving, as the current code solves the problem, but see comments.

sycamore/scans/file_scan.py
eric-anderson: I changed it to support mimetypes.guess_type. 
I'd previously only handled stuff we partitioned.

PR #304, opened 2024-03-22
Address HTML viewer bug when doing sycamore_crawler_http_sort_all

alexaryn:


baitsguy (approved):
We can update/rename proxyhtml (..)when it can do more

PR #307, opened 2024-03-25
Add some documentation for NDD: Sketcher at ingestion time.

alexaryn:


docs/source/data_ingestion_and_preparation/transforms/sketcher.md
```diff
@@ -0,0 +1,12 @@
+## Sketcher
```
jonfritz, line 1: Should this be called "sketch.md" - this seem to follow the pattern with the other transforms?

docs/source/data_ingestion_and_preparation/transforms/sketcher.md
```diff
@@ -0,0 +1,12 @@
...
+The Sketcher transform annotates each Document with a sketch that can be used to identify near-duplicate documents.  This process is the prerequisite for later removing or collapsing near-duplicate documents.  Currently, the sketch consists of a set of hash values called `shingles`.  These are relatively inexpensive to calculate and can safely be a default part of any ingestion pipeline.  Using `Sketcher` in a Sycamore data prep pipeline is relatively easy:
```
jonfritz, line 2: Is the sketch stored as metadata with each Document?

jonfritz (commented):
Two small questions for consideration.

docs/source/data_ingestion_and_preparation/transforms/sketcher.md
alexaryn: Good question.  The access via DocSet is with a method called `sketch` (keeping with the verb forms), but the transform is called `Sketcher`.  It's not clear to me what's right, but I'll rename if you like.

docs/source/data_ingestion_and_preparation/transforms/sketcher.md
alexaryn: Yes, it's metadata, but it's not a property.  Is it clear enough to say "annotates each Document"?  Seems like maybe not.

docs/source/data_ingestion_and_preparation/transforms/sketcher.md
jonfritz: I think most customers will use this in the context of a docset, not the transform directly. Feels like we should then bias towards "sketch" in a section that is talking about how to use DocSet, IMO. The other sections also follow this pattern, it seems.

docs/source/data_ingestion_and_preparation/transforms/sketcher.md
jonfritz: It's clear!

jonfritz (approved):
LGTM!

PR #312, opened 2024-03-27
Add sketch() after explode() in all our default pipelines.

alexaryn:
This should be our new default.  Creating sketches is cheap and useful.  I have spot-checked these changes.  The importer works.  The default-prep-script works.  Also, jupyter_dev_example works.


HenryL27 (approved):
Not sure whether this is _all_ of the places - I havent gone through all of the places or counted them or anything - but this looks like it's at least most of them...
You probly haven't run all these scripts but that's probly fine. 
LGTM once you fix the conflict

jonfritz (approved):
LGTM! Thanks for making these updates. I cant' confirm this is everywhere, but checks the boxes on places I know we'd need it.

PR #315, opened 2024-03-27
Add support for OCR in the Sycamore partitioner.

bsowell:
Setting use_ocr=True will run OCR on each bounding box identified by the partitioner to determine the correct text. Requires tesseract to be installed.


bohou-aryn (approved):
Looks good.

PR #318, opened 2024-03-27
Setup query-time NDD: pre-create RPS processors, add to pipelines

alexaryn:
Perhaps you can test this in your environment.

### DIFF
--- a/apps/opensearch/setup_models.py
+++ b/apps/opensearch/setup_models.py
@@ -534,11 +534,19 @@ def create_pipelines(openai_id):
                     "combination": {"technique": "arithmetic_mean", "parameters": {"weights": [0.111, 0.889]}},
                 }
             }
-        ]
+        ],
+        "response_processors": [
+            {
+                "remote_processor": {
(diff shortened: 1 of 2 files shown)


alexaryn:
I managed to build new containers and I've verified that queries work with the changes to the OpenSearch pipeline and the RPS pipelines.

apps/opensearch/setup_models.py
```diff
@@ -534,11 +534,19 @@ def create_pipelines(openai_id):
                     "combination": {"technique": "arithmetic_mean", "parameters": {"weights": [0.111, 0.889]}},
                 }
             }
-        ]
+        ],
+        "response_processors": [
+            {
+                "remote_processor": {
+                    "endpoint": "rps:2796/RemoteProcessorService/ProcessResponse",
```
HenryL27, line 541: you can also specify this with a https:// prefix but default is ssl. (http://blah also works if you run rps in insecure mode - ie no certs)

HenryL27 (approved):
lgtm

PR #319, opened 2024-03-28
Upgrade to Ray 2.10.0.

bsowell:
There were a few deprecations and changes in the datasource API.

### DIFF
--- a/lib/sycamore/pyproject.toml
+++ b/lib/sycamore/pyproject.toml
@@ -17,7 +17,7 @@ python = ">=3.9,<3.12"
 opensearch-py = "^2.3.1"
 pandas = "2.1.1"
 pdf2image = "^1.16.3"
-ray = { extras = ["default"], version = ">=2.9.0,<2.10.0" }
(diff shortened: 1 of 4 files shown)


lib/sycamore/pyproject.toml
```diff
@@ -17,7 +17,7 @@ python = ">=3.9,<3.12"
 opensearch-py = "^2.3.1"
 pandas = "2.1.1"
 pdf2image = "^1.16.3"
-ray = { extras = ["default"], version = ">=2.9.0,<2.10.0" }
+ray = { extras = ["default"], version = "2.10.0" }
```
HenryL27, line 20: is it worth careting this or will that just be another headach when they break another thing in the next version?

alexaryn (approved):
I don't see anything concerning, but I'm not really familiar.

lib/sycamore/pyproject.toml
bsowell: I thought about that, but each of the last few ray upgrades have caused issues, so I guess I figured we would want to control the upgrades.

PR #322, opened 2024-03-28
Fix importer to check for user, apply similar fix to crawlers

eric-anderson:
Then tons of minor cleanup to make pre-commit pass again.


apps/importer/importer/docker/docker_local_import.py
```diff
@@ -57,6 +57,20 @@ def main():
         print("Assuming execution is in container, using adjusted host")
         running_in_container = True
 
+    # check for user change in Mar 2024; code removable after Mar 2025 when
```
alexaryn, line 60: Optimism!

alexaryn (approved):
LGTM

PR #323, opened 2024-03-29
Upgrade guidance to 0.1.13.

bsowell:
Guidance classifies models as either Chat, Instruction, or Completion. Between versions 0.1.10 and 0.1.13, they reclasified the gpt-3.5-instruct model from instruction to completion. Previously we only handling chat and instruction models, so this commit adds support for completion models as well.


lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -48,6 +50,10 @@ def _execute_instruct(self, model, **kwargs) -> str:
         lm += gen(name=self.var_name)
         return lm[self.var_name]
 
+    def _execute_completion(self, model, **kwargs) -> str:
```
baitsguy, line 53: we should probably add tests at some point but this is fine for now since it'll break integrations and be detected

baitsguy (approved):

PR #327, opened 2024-03-31
Added -anon switch to S3 crawler for public buckets.

alexaryn:
This is needed for the NDD blog post very soon.  It allows the crawling of public S3 buckets without AWS credentials.

### DIFF
--- a/apps/crawler/crawler/s3/crawler/s3_crawler.py
+++ b/apps/crawler/crawler/s3/crawler/s3_crawler.py
@@ -4,7 +4,14 @@ from typing import Any
 
 import boto3
 import botocore.client
+from botocore import UNSIGNED
+from botocore.config import Config
 import sys
+
+
+def usage(val: int) -> None:
+    print("Usage: poetry run python s3_crawler.py bucket_name prefix_value")
+    sys.exit(val)
 
 
 class S3Crawler:
@@ -12,11 +19,13 @@ class S3Crawler:
         self,
         bucket_location: str,
         prefix: str,
+        anon: bool,
         boto_session_args: list[Any] = [],
         boto_session_kwargs: dict[str, Any] = {},
     ):
         self._bucket_location = bucket_location
         self._prefix = prefix
+        self._anon = anon
         self._file_storage_location = "./.data/.s3/downloads"
         self._s3_client = self._get_s3_client(boto_session_args, boto_session_kwargs)
         if os.path.exists("/.dockerenv"):
@@ -88,7 +97,11 @@ class S3Crawler:
         self, boto_session_args: list[Any], boto_session_kwargs: dict[str, Any]
     ) -> botocore.client.BaseClient:
         session = boto3.session.Session(*boto_session_args, **boto_session_kwargs)
-        return session.client("s3")
+        if self._anon:
+            cfg = Config(signature_version=UNSIGNED)
+            return session.client("s3", config=cfg)
+        else:
+            return session.client("s3")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


apps/crawler/crawler/s3/crawler/s3_crawler.py
```diff
@@ -84,30 +105,44 @@ def _is_new_object(self, object_metadata: dict[str, Any], file_path: str) -> boo
...
-    if len(sys.argv) > 3 or (len(sys.argv) > 1 and sys.argv[1] == "-h"):
-        print("Usage : poetry run python s3_crawler.py bucket_name prefix_value")
-    else:
-        if len(sys.argv) == 1:
-            bucket = "aryn-public"
-            prefix = "sort-benchmark"
-        elif len(sys.argv) == 2:
-            bucket = sys.argv[1]
-            prefix = ""
+    anon = False
```
eric-anderson, line 121: Lets use argparse. It's part of the standard library.  Rather than rolling our own (even if it's simple)
https://docs.python.org/3/library/argparse.html

apps/crawler/crawler/s3/tests/unit/test_s3_crawler.py
```diff
@@ -34,7 +34,7 @@ def mock_download_func(bucket, key, file):
         download.side_effect = mock_download_func
 
     def test_s3_crawler_first_crawl(self, mocker, tmp_path: Path):
-        s3_crawler = S3Crawler("bucket", "prefix")
+        s3_crawler = S3Crawler("bucket", "prefix", False)
```
eric-anderson, line 37: I'd expect you to have to change something in the mocking to make sure that you either require credentials or not.

apps/crawler/crawler/s3/crawler/s3_crawler.py
```diff
@@ -28,7 +38,18 @@ def __init__(
...
     def crawl(self) -> None:
-        self._find_and_download_new_objects()
+        try:
+            self._s3_client = self._get_s3_client()
+            self._find_and_download_new_objects()
+            return
+        except NoCredentialsError:
+            if self._anon:
+                raise
+            # Retry anonymously...
```
eric-anderson, line 48: logging.Warning("Automatically retrying in anonymous mode")

apps/crawler/crawler/s3/crawler/s3_crawler.py
```diff
@@ -4,21 +4,31 @@
...
         self,
         bucket_location: str,
         prefix: str,
+        anon: bool,
         boto_session_args: list[Any] = [],
         boto_session_kwargs: dict[str, Any] = {},
     ):
         self._bucket_location = bucket_location
         self._prefix = prefix
+        self._anon = anon
```
eric-anderson, line 29: Auto detect that we want anon mode if AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY are empty.

apps/crawler/crawler/s3/crawler/s3_crawler.py
alexaryn: This file doesn't use `logging` anywhere else.  Is it OK to `print`?

apps/crawler/crawler/s3/crawler/s3_crawler.py
alexaryn: It's not that simple.  Boto3 will look in lots of filesystem places for credentials.

apps/crawler/crawler/s3/crawler/s3_crawler.py
alexaryn: argparse it is.  Click looked like a more extensive job.

apps/crawler/crawler/s3/tests/unit/test_s3_crawler.py
alexaryn: I wasn't going to worry about this right now.  It's still testing what it was meant to test.

eric-anderson: `# TODO: Add a test to verify that this works in anonymous mode. Probably needs to be an integration test to be useful`

eric-anderson (approved):

apps/crawler/crawler/s3/crawler/s3_crawler.py
```diff
@@ -28,7 +39,18 @@ def __init__(
                 )
 
     def crawl(self) -> None:
-        self._find_and_download_new_objects()
+        try:
+            self._s3_client = self._get_s3_client()
```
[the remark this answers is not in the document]
alexaryn: I was mostly going for the smallest change, given it's not my code and we're in a hurry.

PR #328, opened 2024-04-01
add docs for RPS

HenryL27:
There are a couple missing links still

### DIFF
--- a/docs/source/querying_data/remote_processors.md
+++ b/docs/source/querying_data/remote_processors.md
@@ -23,7 +23,7 @@ These processors run in a microservice in the sycamore ecosystem called RPS. The
 
 You can add a remote processor to a search pipeline using standard OpenSearch search pipeline creation syntax:
 
-```json
+```lang-http
 PUT /_search/pipeline/pipeline-with-remote-dedup
 {
   "response_processors": [
@@ -62,7 +62,7 @@ docker container start sycamore-rps-1
 
 And then create the search pipeline like so
 
-```json
+```lang-http
 PUT /_search/pipeline/pipeline-with-dedup-for-debugging
 {
   "response_processors": [
--- a/docs/source/querying_data/remote_processors.md
+++ b/docs/source/querying_data/remote_processors.md
@@ -1,18 +1,15 @@
 # Remote Search Processors
 
-<aside>
-👉 This page assumes you’re familiar with OpenSearch’s concept of a search pipeline and search processor
-
-</aside>
+> 👉 This page assumes you’re familiar with OpenSearch’s concept of a search pipeline and search processor
 
 Sycamore exposes a number of search processors in addition to the ones built in to OpenSearch. We accomplish this with a search processor in OpenSearch called ‘remote-processor’, which makes network calls to a service hosting the sycamore search processors.
 
 These search processors include:
 
 - `debug`: prints the search response to stdout. Useful for debugging.
-- `dedup`: works in conjunction with the `Sketcher` ingest transform to deduplicate search results at query-time. See [NDD Stuff] for more details.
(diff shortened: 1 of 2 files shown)


docs/source/querying_data/remote_processors.md
```diff
@@ -0,0 +1,77 @@
...
+    }
+  ]
+}
+```
+
+The `processor_name` should point to the top-level name of the processor you want to call.
+
+### Configuring RPS
+
+You can change the config file for RPS. Simply declare a processor, which processors are part of it, and their individual parameters. So, maybe I want a processor that dedupes and prints the search response before and after so I can compare them manually. I can declare this configuration in a new `my-pipelines.yml`:
```
jonfritz, line 44: I don't know if we want people doing this right now. I might remove it - unless there is a clear use case.

docs/source/querying_data/remote_processors.md
```diff
@@ -0,0 +1,77 @@
+# Remote Search Processors
+
+> 👉 This page assumes you’re familiar with OpenSearch’s concept of a search pipeline and search processor
+
+Sycamore exposes a number of search processors in addition to the ones built in to OpenSearch. We accomplish this with a search processor in OpenSearch called ‘remote-processor’, which makes network calls to a service hosting the sycamore search processors.
+
+These search processors include:
+
+- `debug`: prints the search response to stdout. Useful for debugging.
+- `dedup`: works in conjunction with the `Sketcher` ingest transform to deduplicate search results at query-time. See the [sketch](../data_ingestion_and_preparation/transforms/sketch.md) transform for more details.
```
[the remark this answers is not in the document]
HenryL27: can bump it. was going with alphabetical scheme but can go by importance scheme. I couldn't find anywhere to point to other than sketch. alex confirmed this

docs/source/querying_data/remote_processors.md
HenryL27: Can I put in a 

> **Warning:** This is super experimental. Don't do this unless you really need to.

? Also happy to delete

docs/source/querying_data/remote_processors.md
```diff
@@ -0,0 +1,83 @@
+# Remote Search Processors
+
+> 👉 This page assumes you’re familiar with OpenSearch’s concept of a search pipeline and search processor
```
jonfritz, line 3: This is not a friendly customer experience. Instead, write a note guiding somewhere were to go to learn.

docs/source/querying_data/remote_processors.md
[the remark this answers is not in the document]
jonfritz: Yes - it's because Alex hasn't written it yet. That's where all the info is. Can you please add this in today once Alex's docs for query time NDD are written? We want users going to that page, not sketcher.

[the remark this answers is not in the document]
jonfritz: Alphabetical is fine, I guess.

docs/source/querying_data/remote_processors.md
jonfritz: If there isn't a use case we know, I'd leave this out. Our docs are already complex, let's try to keep it simple. I'd keep this text on your machine, in case we need to add it back. Unless - do you have a use case we think customers would do here?

jonfritz (commented):
A few more comments. Thx for the iteration.

docs/source/querying_data/remote_processors.md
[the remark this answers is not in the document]
HenryL27: will re-point with a new target. Actually will make alex re-point this.

docs/source/querying_data/remote_processors.md
HenryL27: k, will remove.

jonfritz (approved):
LGTM, pending a nicer note at the top. Thx!

PR #330, opened 2024-04-01
Expand NDD doc into separate file.

alexaryn:


docs/source/querying_data/dedup.md
```diff
@@ -0,0 +1,66 @@
...
+          threshold: 0.45
+- dedup08:
+    processors:
+      - dedup-response:
+          threshold: 0.55
+```
+
+As can be seen, there's one parameter, `threshold`, which is a distance metric that controls how different two documents can be before they're considered not to be duplicates.  A 0.0 means documents must be practically identical.  A 1.0 means all documents are considered to be the same.  Due to limitations in passing parameters through the system, we've pre-created most of the potentially useful setups.
+
+The current implementation of NDD uses "shingles" which consist of 16 hash values.  The `dedup02` preset allows two of those hashes to differ; 0.15 is just a smidge higher than 2 divided by 16.
```
HenryL27, line 46: might want to add a bit that explains it's looking for a strict inequality, not <=

docs/source/querying_data/dedup.md
```diff
@@ -0,0 +1,66 @@
+# De-Duplicating Query Results
+
+As mentioned in [Remote Search Processors](remote_processors.md), the Sycamore near-duplicate detection (NDD) facility can be used to drop duplicates from query results.  This is implemented as a remote response processor called `dedup-response`.  It's configured in `pipelines.yml` like so:
```
HenryL27, line 3: nit: "remote response processor" -> "remote search response processor"

docs/source/querying_data/dedup.md
```diff
@@ -0,0 +1,66 @@
...
+    processors:
+      - dedup-response:
+          threshold: 0.45
+- dedup08:
+    processors:
+      - dedup-response:
+          threshold: 0.55
+```
+
+As can be seen, there's one parameter, `threshold`, which is a distance metric that controls how different two documents can be before they're considered not to be duplicates.  A 0.0 means documents must be practically identical.  A 1.0 means all documents are considered to be the same.  Due to limitations in passing parameters through the system, we've pre-created most of the potentially useful setups.
```
HenryL27, line 44: I feel like some of the explanation in the remote_processors doc could be brought in here

HenryL27 (approved):
lgtm. a few things to think about, idk

PR #331, opened 2024-04-01
Add .profile to container so that we get poetry python not container python

eric-anderson:
Without this, terminals need to do poetry run python to get things like pytesseract


alexaryn (approved):
Seems fine, even if the games part is odd.

PR #333, opened 2024-04-02
Cache entire Amazon Textract response

baitsguy:
Store the entire DocumentAnalysis json response from Textract instead of our parsed Table objects. This lets you use the textractor interface to access fields we hadn't previously used in a typed way (e.g. individual row elements).

I validated this by running the following on a single doc:
1. regular TextractTableExtractor
2. CachedTextractTableExtractor for the first time
3. CachedTextractTableExtractor for the second time

We can likely add integration tests validating that caching is working but I don't want to combine it here.

### DIFF
--- a/lib/sycamore/sycamore/transforms/extract_table.py
+++ b/lib/sycamore/sycamore/transforms/extract_table.py
@@ -63,7 +63,7 @@ class TextractTableExtractor(TableExtractor):
         self._kms_key_id: str = kms_key_id
         self._s3_upload_root: str = s3_upload_root
 
-    def get_textract_result(self, document: Document) -> list[Table]:
+    def get_textract_result(self, document: Document) -> TextractorDocument:
 
         extractor = Textractor(self._profile_name, self._region_name, self._kms_key_id)
         path = document.properties["path"]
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/transforms/extract_table.py
```diff
@@ -107,40 +112,24 @@ def bbox_to_coord(bbox):
...
-        ...
-      ]
-    }
-    bbox format is in [left/width, top/height, right/width, bottom/height]
+    Extends TextractTableExtractor with S3 based cache support for raw Textract results.
+
+    CachedTextractTableExtractor overrides the 'get_textract_result' method by doing the following:
+     1. if cache hit for current document, get from cache and return, otherwise continue
+     2. if run_full_textract is enabled, call textractor on the whole document and go to step 4
+     3. else clip pages which contain tables and run table extraction using textractor
```
bohou-aryn, line 131: is this intentional, I see we still do the clip pages

bohou-aryn (approved):

lib/sycamore/sycamore/transforms/extract_table.py
baitsguy: Yeah that's point 3?

PR #337, opened 2024-04-04
Added ability to write JSONL block files.

alexaryn:


lib/sycamore/sycamore/writers/file_writer.py
```diff
@@ -100,6 +101,18 @@ def elements_to_bytes(doc: Document) -> bytes:
     return out.getvalue().encode("utf-8")
 
 
+def document_to_bytes(doc: Document) -> bytes:
+    """
+    Returns a UTF-8 encoded json string of the document.  Adds newline.
+    Beware this will try to interpret binary_representation as UTF-8.
```
bsowell, line 107: Is this ever what we want? Should we just skip the binary_representation entirely?

lib/sycamore/sycamore/writers/file_writer.py
```diff
@@ -176,3 +231,19 @@ def __init__(
...
+        self,
+        path: str,
+        filesystem: Optional[FileSystem] = None,
+    ) -> None:
+        super().__init__(path, filesystem=filesystem, filename_provider=BlockFilenameProvider())
+
+    def write_block_to_file(self, block: BlockAccessor, file: NativeFile) -> None:
+        for row in block.iter_rows(True):  # type: ignore[var-annotated]
+            doc = Document.from_row(row)
+            del doc.binary_representation  # Doesn't make sense in JSON
```
bsowell, line 247: Oh I see. So the comment above was just to warn other callers that might want to use document_to_bytes?

bsowell (commented):
Generally looks good. Might be nice to have a few tests here: https://github.com/aryn-ai/sycamore/blob/main/lib/sycamore/sycamore/tests/unit/test_writer.py#L112. Or even something simple that verifies that we can read the output we write.

lib/sycamore/sycamore/writers/file_writer.py
alexaryn: Looking at JSONEncodeWithUserDict, it doesn't seem possible to know what "bytes" object we're looking at.  I figured it was safest not to mess with that.

lib/sycamore/sycamore/writers/file_writer.py
alexaryn: Pretty much.  This particular bytes object was the only problem I ran into.  I have no idea what other values are possible.  In many cases, they may be UTF-8 strings.  It seems safer to drop this known field than to blanket-drop all bytes objects.  Time will tell.

alexaryn:
Yes, there aren't any tests yet.  In manual testing, ctx.read.json() seems to work fine on output from this.

bsowell (approved):
LGTM. Thanks for adding the tests.

PR #339, opened 2024-04-05
Initial Container ITs

HenryL27:
This adds tests that :
- ingest data using a crawler image
- run a bunch of queries over the resulting indices

Run them locally by starting a fresh docker compose and then doing `poetry run pytest apps/integration`


apps/integration/integration/containers/preexisting.py
```diff
@@ -0,0 +1,63 @@
...
+        "rps": ("localhost", 2796),
+        "demo-ui": ("localhost", 3000),
+        "jupyter": ("localhost", 8888),
+    }
+
+
+@pytest.fixture(scope="session")
+def container_handles():
+    docker_client = docker.from_env()
+    jupyter = docker_client.containers.list(filters={"label": "com.docker.compose.service=jupyter"})[0]
```
pparmar30, line 33: Just wondering if we can do containers.get("name")

apps/integration/integration/containers/preexisting.py
```diff
@@ -0,0 +1,63 @@
...
+    rps = docker_client.containers.list(filters={"label": "com.docker.compose.service=rps"})[0]
+    importer = docker_client.containers.list(filters={"label": "com.docker.compose.service=importer"})[0]
+    return {"jupyter": jupyter, "opensearch": opensearch, "demo_ui": demo_ui, "rps": rps, "importer": importer}
+
+
+@pytest.fixture(scope="session")
+def opensearch_client(container_urls):
+    host, port = container_urls["opensearch"]
+    urlstr = f"https://{host}:{port}"
+    while True:
```
pparmar30, line 45: We should time out here

apps/integration/integration/containers/preexisting.py
```diff
@@ -0,0 +1,63 @@
+import pytest
```
pparmar30, line 1: nit : assuming you are just setting things up here, file name can be fixtures or setup?

apps/integration/integration/containers/preexisting.py
HenryL27: my main worry doing that is that docker compose likes to add a bunch of extra stuff to the name.
so the service 'jupyter' is named 'sycamore-jupyter-1'; if there are replicas I guess they'll get other numbers... This seemed safer

apps/integration/integration/containers/preexisting.py
HenryL27: well, I'm gonna have another file that starts the containers. These are fixtures that help me deal with running containers. so will rename to running.py

apps/integration/integration/queries/options.py
```diff
@@ -0,0 +1,49 @@
...
+            return True
+        return False
+
+
+class BooleanOption(Option):
+    def __init__(self, name):
+        super().__init__(name, [False, True])
+
+
+class OptionSet(Iterable):
```
pparmar30, line 41: nit : Any way to specify this is an Iterable of type Option?

apps/integration/integration/queries/queries.py
```diff
@@ -0,0 +1,183 @@
...
+from dataclasses import dataclass
+
+
+class RagMode(Enum):
+    OFF = 0
+    ONE_SHOT = 1
+    CONVERSATIONAL = 2
+
+
+DEFAULT_OPTIONS = OptionSet(
```
pparmar30, line 18: Does this create a total of 48 tests?

apps/integration/integration/queries/options.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def get_name(self):
+        return self._name
+
+    def advance(self):
+        self._current_pos += 1
+        if self._current_pos == len(self._values):
+            self._current_pos = 0
+            return True
+        return False
```
pparmar30, line 29: Nit : After starring at it for a while, I understand how it is being used but ideally, Would you not want to return True if an option can advance?

apps/integration/integration/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+
+QUERY_FIXTURE_NAME = "os_query"
+
+
+def pytest_generate_tests(metafunc):
+    if QUERY_FIXTURE_NAME in metafunc.fixturenames:
+        metafunc.parametrize(QUERY_FIXTURE_NAME, list(QueryConfigGenerator(DEFAULT_OPTIONS)))
+
+
+__all__ = [
```
pparmar30, line 17: Are you planning to use these elsewhere and hence importing?

apps/integration/integration/tests/test_basic.py
```diff
@@ -0,0 +1,32 @@
...
+    pipeline_name = "it-pipeline"
+    logger.info(pipeline)
+    opensearch_client.transport.perform_request(method="PUT", url=f"/_search/pipeline/{pipeline_name}", body=pipeline)
+    logger.info(query)
+    search_response = opensearch_client.search(
+        index=ingested_index.name,
+        params={"search_pipeline": pipeline_name, "_source_excludes": "embedding"},
+        body=query,
+    )
+    logger.info(search_response)
```
pparmar30, line 35: Do we want to assert on any behavior here?

apps/integration/integration/containers/preexisting.py
```diff
@@ -0,0 +1,63 @@
...
+from pathlib import Path
+from testcontainers.compose.compose import DockerCompose
+from typing import List
+
+
+def docker_compose(services: List[str] = []):
+    sycamore_root = Path(__file__).parent.parent.parent.parent.parent
+    print(sycamore_root)
+    if len(services) > 0:
+        return DockerCompose(sycamore_root, services=services)
```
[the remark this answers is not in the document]
HenryL27: yep. and if services is empty, it's the default set

apps/integration/integration/ingests/crawler.py
```diff
@@ -0,0 +1,67 @@
...
+                for file in imported_files:
+                    print(f"imported {file}")
+                    files.remove(file)
+                print(f"remaining files: {files}")
+            if len(files) == 0:
+                print("finished importing!")
+                break
+            if log.startswith("No changes") and len(files) != num_files:
+                raise RuntimeError(
+                    "Importer thinks there are no more files to ingest, but there are more files to ingest"
```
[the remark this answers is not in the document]
HenryL27: Yeah, I'm pretty nervous about this whole deal. We could look at the crawler data volume I guess... that would also entail clearing it out before every run, which is probably worth doing anyway. Waiting for the ingestion to finish is harder, since the importer doesn't exit when it finishes. I guess we could maybe do some sort of opensearch query to get the set of filenames in the index, although I'm not sure that means all the chunks are in?

apps/integration/integration/queries/options.py
HenryL27: True means the next thing can advance. Think of the return value as a carry bit in a counter, cuz that's essentially what it represents

apps/integration/integration/tests/test_basic.py
HenryL27: probably? I mean the main thing is that the query doesn't throw an error itself.

apps/integration/integration/ingests/__init__.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+from integration.ingests.crawler import HttpCrawlerIndex
+
+
+@pytest.fixture(scope="session", params=["crawler-http-one", "crawler-http-all"])
+def ingested_index(request, opensearch_client, container_handles):
```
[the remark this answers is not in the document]
HenryL27: agree. moved it to ingests/index.py, and moved IndexInfo to ingests/index_info.py (in danger of a circ. dependency)

apps/integration/integration/ingests/crawler.py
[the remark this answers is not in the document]
HenryL27: looking into it further, it seems like there isn't a way to look at the data in the volume, at least on macos, because docker runs in a VM and gets like a drive partition that I can't access from outside (in my tests). We could do a bunch of `container.exec("ls")` calls, but this seems not much better. I'll also note that eric has plans to change how the importer does its detection type stuff so I think all of this code may have to get chucked regardless of whether I do something more robust with it. So I'm tempted to leave it as is.
You could also do something like repeatedly query opensearch until the index size stops changing, but I don't trust opensearch to update reliably enough for that to work

pparmar30 (approved):

apps/integration/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.poetry]
+name = "integration"
+version = "0.1.0"
+description = "Integration Testing for Sycamore"
+authors = ["Your Name <you@example.com>"]
```
bsowell, line 5: nit: authors = ["aryn.ai <opensource@aryn.ai>"]

bsowell (approved):
I don't see any showstoppers here. It's pretty complex, which is I guess is just managing the OpenSearch stuff, but we will have to keep an eye on how stable the tests are. 

Thanks for doing this!

apps/integration/integration/queries/opensearch.py
```diff
@@ -0,0 +1,67 @@
+from opensearchpy import OpenSearch
+
+
+class OpenSearchHelper:
```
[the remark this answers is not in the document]
HenryL27: agree. something like a `lib/sycamore-stack-utils` may be in order

PR #341, opened 2024-04-09
Fix bug in updating a single property and most workarounds for the bug.

eric-anderson:
Also nbstripout on the notebook.

### DIFF
--- a/notebooks/ndd_example.ipynb
+++ b/notebooks/ndd_example.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "ed9dd0a2-6b12-436f-b6c3-02e0072115e3",
+   "id": "0",
    "metadata": {},
    "source": [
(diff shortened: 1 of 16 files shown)


notebooks/ndd_example.ipynb
bsowell, line 1: Alex links to this one from the blog post. I am a fan of nbstripout in general, but I can see some value in having outputs here. I guess the downside is that stuff like execution time changes every time you run it? I'm okay either way. (I asked Alex and he said that he slightly preferred to keep the output there, but didn't feel too strongly either).

bsowell (approved):
Thanks for doing this. As far as I can tell the changes look good.

notebooks/ndd_example.ipynb
eric-anderson: I updated the pre-commit to skip the ndd_example.ipynb file.

bsowell (approved):

PR #342, opened 2024-04-09
Set RPS default version to follow VERSION again

HenryL27:
We had set the default to 'latest_rc' because up until recently there wasn't a stable tag. Now there is, so we can follow VERSION again


alexaryn (approved):
Woo hoo!

PR #344, opened 2024-04-11
Better container integration testing automation

HenryL27:
- Jupyter testing
- Create compose stack as a fixture
- provide a tag to run tests against

### DIFF
--- a/apps/integration/integration/automation/runtests.sh
+++ b/apps/integration/integration/automation/runtests.sh
@@ -42,7 +42,7 @@ handle_outputs() {
 runtests() {
   docker system prune -f --volumes
   docker compose up reset
-  poetry run pytest apps/integration/ -p integration.conftest --noconftest --docker-tag latest_rc
+  poetry run pytest apps/integration/ -p integration.conftest \--noconftest --docker-tag latest_rc
   # this is a complicated command, so: ^                        ^            ^ test against containers tagged latest_rc
   #                                    |                     don't load conftest at pytest runtime; it's already loaded
   #                                     load conftest with plugins, to capture the custom command line arg --docker-tag
--- a/apps/integration/integration/automation/runtests.sh
+++ b/apps/integration/integration/automation/runtests.sh
@@ -42,7 +42,7 @@ handle_outputs() {
 runtests() {
   docker system prune -f --volumes
   docker compose up reset
-  poetry run pytest apps/integration/ -p integration.conftest \--noconftest --docker-tag latest_rc
+  poetry run pytest apps/integration/ -p integration.conftest --noconftest --docker-tag latest_rc
   # this is a complicated command, so: ^                        ^            ^ test against containers tagged latest_rc
   #                                    |                     don't load conftest at pytest runtime; it's already loaded
   #                                     load conftest with plugins, to capture the custom command line arg --docker-tag
--- a/apps/integration/integration/automation/runtests.sh
+++ b/apps/integration/integration/automation/runtests.sh
@@ -12,7 +12,7 @@ main() {
     echo "Changes detected. Running Tests" >&2
     poetry install
     build_containers
-    poetry run pytest apps/integration
+    runtests
     handle_outputs
   else
     echo "No changes detected. Skipping integration tests" >&2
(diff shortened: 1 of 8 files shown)


apps/integration/integration/automation/runtests.sh
```diff
@@ -39,5 +39,14 @@ handle_outputs() {
   echo "Yep, definitely handling test outputs. That's what this function does" >&2
 }
 
+runtests() {
+  docker system prune -f --volumes
+  docker compose up reset
+  poetry run pytest apps/integration/ -p integration.conftest --noconftest --docker-tag latest_rc
+  # this is a complicated command, so: ^                        ^            ^ test against containers tagged latest_rc
+  #                                    |                     don't load conftest at pytest runtime; it's already loaded
+  #                                     load conftest with plugins, to capture the custom command line arg --docker-tag
```
alexaryn, line 48: Nice hack.  Not sure the ASCII art translates to narrower screens.

apps/integration/integration/containers/stack.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import List
+
+
+def docker_compose(services: List[str] = []):
+    """
+    Get a docker compose object that controls some set of the sycamore stack
+    :param services: the list of services to control. [] -> all services
+    :return: the docker compose object
+    """
+    if len(services) > 0:
```
alexaryn, line 13: I'd just have `if not services:` so people can pass in None, too.  Reads better, too.

apps/integration/integration/containers/stack.py
```diff
@@ -0,0 +1,39 @@
...
+        return DockerCompose(SYCAMORE_ROOT, services=services)
+    else:
+        return DockerCompose(SYCAMORE_ROOT)
+
+
+def set_version_tag(tag: str):
+    with open(SYCAMORE_ROOT / ".env", "r") as f:
+        lines = f.readlines()
+    lines[0] = f"VERSION={tag}\n"
```
alexaryn, line 22: This seems remarkably delicate.  What if somebody puts a comment at the top of .env?

apps/integration/integration/ingests/index.py
```diff
@@ -1,11 +1,12 @@
 import pytest
 from integration.ingests.crawler import HttpCrawlerIndex
+from integration.ingests.jupyter import JupyterIndex
 
+INGEST_PROFILES = ["crawler-http-one", "crawler-http-all", "jupyter-default-prep", "jupyter-dev-example"]
```
alexaryn, line 5: Do we care about sycamore_demo?

apps/integration/integration/ingests/index.py
```diff
@@ -25,6 +26,13 @@ def ingested_index(opensearch_client, container_handles, ingest_profile):
...
             profile="sort-all", opensearch=opensearch_client, importer=container_handles["importer"]
         )
+    elif ingest_profile == "jupyter-default-prep":
+        index_ctx = JupyterIndex(
+            nb_name="default-prep-script.ipynb", opensearch=opensearch_client, jupyter=container_handles["jupyter"]
+        )
+    elif ingest_profile == "jupyter-dev-example":
+        index_ctx = JupyterIndex(
+            nb_name="jupyter_dev_example.ipynb", opensearch=opensearch_client, jupyter=container_handles["jupyter"]
+        )
```
alexaryn, line 36: Could this stuff be data-driven or based on patterns?  A bunch of if-then-else logic isn't ideal when the code is so similar.

apps/integration/integration/ingests/jupyter.py
```diff
@@ -0,0 +1,41 @@
...
+}
+
+
+class JupyterIndex:
+    def __init__(self, nb_name: str, opensearch: OpenSearch, jupyter: Container):
+        self._nb_name = nb_name
+        self._opensearch = opensearch
+        self._jupyter = jupyter
+        if nb_name not in JUPYTER_NB_INFO:
+            raise ValueError(f"Unrecognized notebook name: {nb_name}")
```
alexaryn, line 19: I'd expect to see parameter validation come first.

alexaryn (approved):
Some comments; no blockers.

apps/integration/integration/containers/stack.py
HenryL27: well, don't! but also yeah will fix

apps/integration/integration/ingests/index.py
HenryL27: If by that you mean `jupyter/sycamore_demo.ipynb`, then it isn't in the container, so no. If there's some other sycamore_demo then I don't know what it is

apps/integration/integration/ingests/index.py
HenryL27: Could do something like that, sure (although I think python's `match` statement is only a thing in 3.12, and you really want that for pattern matching stuff... pretty sure until 3.12 if elif is how you do that). I'm not sure we're ever going to have enough ingest profiles that it's worth doing anything more complicated than simply enumerating them. remember, each one of these is a full sycamore ingest job + 48 opensearch queries, so the cost of adding to this section is not in the complexity/duplicativeness of this section

apps/integration/integration/automation/runtests.sh
HenryL27: I'll let you take up the 80 versus 120 argument with someone else.

PR #345, opened 2024-04-11
Fix almost all the pyproject.toml and poetry.lock files to have consistent requirements on python dependencies.

eric-anderson:
demo-ui can't be fixed as it has conflicting requirements.
add script to perform the work of cross-checking.
manually resolve (and document why) all the special cases in the new lib/poetry-lock "library"


lib/poetry-lock/poetry-lock-all.sh
```diff
@@ -0,0 +1,31 @@
...
+for i in **/pyproject.toml; do
+    # Demo UI needs 0.28, rest of sycamore needs 1.x
+    [[ $i = *openai-proxy* ]] && continue
+    (
+        echo "--------------------- proecessing in $i"
+        cd $(dirname "$i")
+        poetry lock --no-update || return 1
+        poetry install |& tee /tmp/poetry-install.out || return 1
+        perl -ne 'print qq{$1 = "$2"\n} if /Downgrading (\S+) \((\S+) ->/o;' </tmp/poetry-install.out
+    )
```
HenryL27, line 30: Ok I think I understand that this is the 'lock everything' section but I have no idea what this is actually doing. also nit: proecessing -> processing

lib/poetry-lock/sycamore_poetry_lock/noop.py
```diff
@@ -0,0 +1 @@
+pass
```
HenryL27, line 1: nice

HenryL27 (approved):
Cool!

HenryL27:
Although won't this leave us no hope of reducing our container sizes?

eric-anderson:
This shouldn't affect the container sizes. When we do installs we don't install the dev dependencies.

PR #347, opened 2024-04-11
Update NDD notebook with JSON/PDF ingestion options.

alexaryn:


bsowell (approved):
LGTM as far as I can tell.

PR #349, opened 2024-04-12
Bind mount to convey SSL cert/key to Jupyter & UI

alexaryn:
To test, put two files in apps/jupyter/bind_dir:
- hostcert.pem
- hostkey.pem

Then, run like:
- VERSION=test SSL=1 docker compose up

Jupyter and the UI should come up with the specified certificate, probably still with a warning, as we're not using fully-qualified domain names yet.

### DIFF
--- a/apps/demo-ui/run-ui.sh
+++ b/apps/demo-ui/run-ui.sh
@@ -15,8 +15,7 @@ if [[ -f /.dockerenv ]]; then
     PROXYDIR=/home/pn/py-proxy
     UIDIR=/home/pn/js-ui
 else
-    cd $(dirname $0)
-    DIR=$(pwd)
+    DIR=$(cd "$(dirname "$0")"; pwd)
     PROXYDIR="${DIR}/openai-proxy"
     UIDIR="${DIR}/ui"
 fi
@@ -29,7 +28,7 @@ cd "${PROXYDIR}"
 : ${HOST:=localhost}
 SELFSIGNED="${HOST}_ss"
 SSLNAME="${SELFSIGNED}"
-ARYN_DIR=/etc/opt/aryn
+ARYN_ETC=/etc/opt/aryn
 
 if [[ (! -f ${SELFSIGNED}-key.pem) || (! -f ${SELFSIGNED}-cert.pem) ]]; then
     openssl req -batch -x509 -newkey rsa:4096 -days 10000 \
@@ -40,11 +39,11 @@ if [[ (! -f ${SELFSIGNED}-key.pem) || (! -f ${SELFSIGNED}-cert.pem) ]]; then
     echo "Created ${HOST} certificate"
 fi
 
-if [[ -f ${ARYN_DIR}/hostcert.pem && -f ${ARYN_DIR}/hostkey.pem ]]; then
-    cp -p "${ARYN_DIR}/hostcert.pem" "${HOST}-cert.pem"
-    cp -p "${ARYN_DIR}/hostkey.pem" "${HOST}-key.pem"
+if [[ -f ${ARYN_ETC}/hostcert.pem && -f ${ARYN_ETC}/hostkey.pem ]]; then
+    cp -p "${ARYN_ETC}/hostcert.pem" "${HOST}-cert.pem"
+    cp -p "${ARYN_ETC}/hostkey.pem" "${HOST}-key.pem"
     SSLNAME="${HOST}"
-    echo "Copied certificate from ${ARYN_DIR}"
+    echo "Copied certificate from ${ARYN_ETC}"
 fi
 
 if [[ ${SSL} == 0 ]]; then
--- a/apps/demo-ui/run-ui.sh
+++ b/apps/demo-ui/run-ui.sh
@@ -27,20 +27,33 @@ function cleanup {
 
 cd "${PROXYDIR}"
 : ${HOST:=localhost}
-if [[ (! -f ${HOST}-key.pem) || (! -f ${HOST}-cert.pem) ]]; then
+SELFSIGNED="${HOST}_ss"
+SSLNAME="${SELFSIGNED}"
+ARYN_DIR=/etc/opt/aryn
+
+if [[ (! -f ${SELFSIGNED}-key.pem) || (! -f ${SELFSIGNED}-cert.pem) ]]; then
     openssl req -batch -x509 -newkey rsa:4096 -days 10000 \
     -subj "/C=US/ST=California/O=Aryn.ai/CN=${HOST}" \
     -extensions v3_req -addext "subjectAltName=DNS:${HOST}" \
-    -noenc -keyout "${HOST}-key.pem" -out "${HOST}-cert.pem" 2> /dev/null
+    -noenc -keyout "${SELFSIGNED}-key.pem" \
+    -out "${SELFSIGNED}-cert.pem" 2> /dev/null
     echo "Created ${HOST} certificate"
 fi
+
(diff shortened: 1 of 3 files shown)


apps/demo-ui/run-ui.sh
```diff
@@ -27,20 +27,33 @@ function cleanup {
 
 cd "${PROXYDIR}"
 : ${HOST:=localhost}
-if [[ (! -f ${HOST}-key.pem) || (! -f ${HOST}-cert.pem) ]]; then
+SELFSIGNED="${HOST}_ss"
+SSLNAME="${SELFSIGNED}"
+ARYN_DIR=/etc/opt/aryn
```
eric-anderson, line 32: I'd call this ARYN_ETC 
(I would have thought of /opt/aryn as ARYN_DIR)

apps/demo-ui/run-ui.sh
```diff
@@ -59,9 +72,11 @@ cd "${UIDIR}"
 # Running the UI this way means that the html is unminified and hence easy to read
 # Since the UI is open source, there's little downside of doing this and it helps with
 # debugging.
+# See this link for why we're using RSA self-signed certificates for npm:
+# https://community.letsencrypt.org/t/ecdsa-certificates-not-supported-by-create-react-app/201387/9
 BROWSER=none PORT=3001 WDS_SOCKET_PORT=3000 HTTPS=true \
-SSL_CRT_FILE="${PROXYDIR}/${HOST}-cert.pem" \
-SSL_KEY_FILE="${PROXYDIR}/${HOST}-key.pem" \
+SSL_CRT_FILE="${PROXYDIR}/${SELFSIGNED}-cert.pem" \
```
eric-anderson, line 78: If we have a real key, why are we using selfsigned here?

apps/jupyter/run-jupyter.sh
```diff
@@ -25,6 +27,9 @@ rm /app/.local/share/jupyter/runtime/jpserver-*-open.html 2>/dev/null
 if [[ ${SSL} == 0 ]]; then
     echo "Jupyter not serving over SSL."
     SSLARG=
+elif [[ -f ${ARYN_DIR}/hostcert.pem && -f ${ARYN_DIR}/hostkey.pem ]]; then
+    echo "Using SSL certificate from ${ARYN_DIR}"
+    SSLARG="--certfile=\"${ARYN_DIR}/hostcert.pem\" --keyfile=\"${ARYN_DIR}/hostkey.pem\""
```
eric-anderson, line 32: I would do this as SSLARG=("--certfile=\"${ARYN_DIR}/hostcert.pem\"" "--keyfile=\"${ARYN_DIR}/hostkey.pem\"")
and then use it as "${SSLARG[@]}"
That guarantees it will work even if there are spaces in ARYN_DIR.
Near as I can tell as written it will split the spaces:
```
% A="--c=\"abc def ghi\" --d=\"a b c d e\""
% cat /tmp/foo.sh 
echo $#
echo 1:$1
echo 2:$2
echo 3:$3
echo 4:$4
% /tmp/foo.sh $A
8
1:--c="abc
2:def
3:ghi"
4:--d="a
```

apps/demo-ui/run-ui.sh
alexaryn: There's a comment on L75.  For some reason, npm doesn't support ECDSA.

apps/demo-ui/run-ui.sh
alexaryn: Sure.

apps/jupyter/run-jupyter.sh
```diff
@@ -22,18 +31,18 @@ EOF
 fi
 ln -snf "${JUPYTER_CONFIG_DOCKER}" $HOME/.jupyter
 
-rm /app/.local/share/jupyter/runtime/jpserver-*-open.html 2>/dev/null
+find "${RUNTIME_DIR}" -type f -name 'jpserver-*-open.html' -delete 2>/dev/null
```
eric-anderson, line 34: nit: If find doesn't report errors on no matching files, remove the 2>/dev/null

apps/jupyter/run-jupyter.sh
```diff
@@ -42,19 +51,20 @@ else
...
+    while [[ $(find "${RUNTIME_DIR}" -type f -name 'jpserver-*-open.html' 2>/dev/null | wc -l) = 0 ]]; do
         echo "Waiting for jpserver-*-open.html to appear"
         sleep 1
     done
-    FILE="$(ls /app/.local/share/jupyter/runtime/jpserver-*-open.html)"
-    if [[ $(echo "${FILE}" | wc -w) != 1 ]]; then
-        echo "ERROR: got '${FILE}' for jpserver files"
-        ls /app/.local/share/jupyter/runtime
-        echo "ERROR: Multiple jpsterver-*-open.html files"
+    FILE="$(find "${RUNTIME_DIR}" -type f -name 'jpserver-*-open.html')"
```
eric-anderson, line 62: why change these from ls to find?  I think both are correct in the given usage.

apps/jupyter/run-jupyter.sh
```diff
@@ -42,19 +51,20 @@ else
...
+    if [[ $(wc -l <<< ${FILE}) != 1 ]]; then
+        CONCAT=$(xargs -r <<< ${FILE})
+        echo "ERROR: got '${CONCAT}' for jpserver files"
+        ls "${RUNTIME_DIR}"
```
eric-anderson, line 66: find -ls
(to match with the uses of find elsewhere, which would also have issues if the files are in subdirectories while ls won't show them)

eric-anderson (approved):

apps/jupyter/run-jupyter.sh
alexaryn: It reports errors on directories it lacks permission to traverse.

apps/jupyter/run-jupyter.sh
alexaryn: The big issue is if the * expands to something containing spaces, but also for uniformity.

apps/jupyter/run-jupyter.sh
alexaryn: We do print out the matches in the line above.  We could just use ls -R if we don't expect too much output.

eric-anderson: Why not ```find "${RUNTIME_DIR}" -type f -ls```
I'm also ok with ls -R, but since you switched to find everywhere else it seems better to be consistent.

apps/jupyter/run-jupyter.sh
```diff
@@ -42,19 +51,20 @@ else
...
     done
-    FILE="$(ls /app/.local/share/jupyter/runtime/jpserver-*-open.html)"
-    if [[ $(echo "${FILE}" | wc -w) != 1 ]]; then
-        echo "ERROR: got '${FILE}' for jpserver files"
-        ls /app/.local/share/jupyter/runtime
-        echo "ERROR: Multiple jpsterver-*-open.html files"
+    FILE="$(find "${RUNTIME_DIR}" -type f -name 'jpserver-*-open.html')"
+    if [[ $(wc -l <<< ${FILE}) != 1 ]]; then
+        CONCAT=$(xargs -r <<< ${FILE})
+        echo "ERROR: got '${CONCAT}' for jpserver files"
```
eric-anderson, line 65: Why not
```
echo "ERROR: got '${FILE}' for jpserver files"
```

apps/jupyter/run-jupyter.sh
eric-anderson: I don't see how that's an issue. Agree on uniformity, if we're going to use find in one place we should use it everywhere.

PR #350, opened 2024-04-12
Add image building and tagged testing to CO

HenryL27:
now `runtests.sh` also builds images for a tag specified at runtime and pushes them to dockerhub, using a registry cache (also in dockerhub)

### DIFF
--- a/apps/integration/integration/automation/runtests.sh
+++ b/apps/integration/integration/automation/runtests.sh
@@ -16,6 +16,7 @@ main() {
     echo "Changes detected. Running Tests" >&2
     poetry install
     build_containers
+    exit 0
     runtests
     handle_outputs
   else
@@ -23,12 +24,17 @@ main() {
   fi
 }
 
+error() {
+  echo "ERROR: $@"
+  exit 1
+}
+
 checkout_main_if_new() {
   old_sha="$(git rev-parse HEAD)"
   git fetch origin main >&2
   new_sha="$(git rev-parse FETCH_HEAD)"
   if [[ "${old_sha}" != "${new_sha}" ]]; then
-    git pull origin main >&2
+    git pull --rebase origin main >&2
     return 0
   else
     return 1
@@ -62,7 +68,7 @@ runtests() {
 docker-build-hub() {
   local docker_file="$1"
   [[ -n "${docker_file}" ]] || error "missing ${docker_file}"
-  local repo_name="$(_docker-repo-name)"
+  local repo_name="$(_docker-repo-name "${docker_file}")"
   [[ -n "${repo_name}" ]] || error "empty repo name"
   shift
 
@@ -75,8 +81,10 @@ docker-build-hub() {
 }
 
 _docker-repo-name() {
+  local docker_file="$1"
+  echo "Finding repo name in: ${docker_file}" >&2
   local repo_name="$(grep '^# Repo name: ' "${docker_file}" | awk '{print $4}')"
-  if [[ $(echo "${repo_name}" | wc -w) != 1 ]]; then
+  if (( $(wc -w <<< ${repo_name}) != 1 )); then
     echo "Unable to find repo name in ${docker_file}" 1>&2
     exit 1
   fi
--- a/apps/integration/integration/automation/runtests.sh
+++ b/apps/integration/integration/automation/runtests.sh
@@ -1,10 +1,14 @@
 #!/bin/bash
 
+TAG="$1"
+[[ -z "${TAG}" ]] && TAG="latest_rc"
+
 main() {
   if [[ ! -d ".git" ]]; then
     echo "Error: please run this script from sycamore root!" >&2
     exit 1
   fi
+  echo "Building/testing tag ${TAG}" >&2
   echo "Get the newest git commits" >&2
   checkout_main_if_new
   local should_run=$?
@@ -33,6 +37,13 @@ checkout_main_if_new() {
 
 build_containers() {
   echo "Yep, definitely building containers. That's what this function does" >&2
+  docker-build-hub apps/crawler/crawler/http/Dockerfile
+  docker-build-hub apps/crawler/crawler/s3/Dockerfile
+  docker-build-hub apps/importer/Dockerfile.buildx
+  docker-build-hub apps/opensearch/Dockerfile
+  docker-build-hub apps/jupyter/Dockerfile.buildx --build-arg=TAG="${TAG}"
+  docker-build-hub apps/demo-ui/Dockerfile.buildx
+  docker-build-hub apps/remote-processor-service/Dockerfile.buildx
 }
 
 handle_outputs() {
@@ -42,11 +53,47 @@ handle_outputs() {
 runtests() {
   docker system prune -f --volumes
   docker compose up reset
-  poetry run pytest apps/integration/ -p integration.conftest --noconftest --docker-tag latest_rc
+  poetry run pytest apps/integration/ -p integration.conftest --noconftest --docker-tag "${TAG}"
   # this is a complicated command, so: ^                        ^            ^ test against containers tagged latest_rc
   #                                    |                     don't load conftest at pytest runtime; it's already loaded
   #                                     load conftest with plugins, to capture the custom command line arg --docker-tag
 }
 
+docker-build-hub() {
+  local docker_file="$1"
+  [[ -n "${docker_file}" ]] || error "missing ${docker_file}"
+  local repo_name="$(_docker-repo-name)"
+  [[ -n "${repo_name}" ]] || error "empty repo name"
+  shift
+
+  local platform=linux/amd64,linux/arm64
+  echo
+  echo "Building in sycamore and pushing to docker hub with repo name '${repo_name}'"
+  docker buildx build "$(_docker-build-args)" -t "${repo_name}:${TAG}" -f "${docker_file}" --platform ${platform} . \
+    || error "buildx failed"
+  echo "Successfully built in $dir using docker file $docker_file"
+}
+
+_docker-repo-name() {
+  local repo_name="$(grep '^# Repo name: ' "${docker_file}" | awk '{print $4}')"
+  if [[ $(echo "${repo_name}" | wc -w) != 1 ]]; then
+    echo "Unable to find repo name in ${docker_file}" 1>&2
+    exit 1
+  fi
+  echo "${repo_name}"
+}
+
+_docker-build-args() {
+  local branch="$(git status | head -1 | grep 'On branch ' | awk '{print $3}')"
+  local rev="$(git rev-parse --short HEAD)"
+  local date="$(git show -s --format=%ci HEAD | sed 's/ /_/g')"
+  local diff=unknown
+  if [[ $(git status | grep -c 'nothing to commit, working tree clean') = 1 ]]; then
+    diff=clean
+  else
+    diff="pending_changes_$(git diff HEAD | shasum | awk '{print $1}')"
+  fi
+  echo "--build-arg=GIT_BRANCH=${branch} --build-arg=GIT_COMMIT=${rev}--${date} --build-arg=GIT_DIFF=${diff}"
+}
 
 main
--- a/apps/integration/integration/automation/runtests.sh
+++ b/apps/integration/integration/automation/runtests.sh
@@ -16,7 +16,6 @@ main() {
     echo "Changes detected. Running Tests" >&2
     poetry install
     build_containers
-    exit 0
     runtests
     handle_outputs
   else
--- a/apps/integration/integration/automation/runtests.sh
+++ b/apps/integration/integration/automation/runtests.sh
@@ -75,8 +75,10 @@ docker-build-hub() {
   local platform=linux/amd64,linux/arm64
   echo
   echo "Building in sycamore and pushing to docker hub with repo name '${repo_name}'"
-  docker buildx build "$(_docker-build-args)" -t "${repo_name}:${TAG}" -f "${docker_file}" --platform ${platform} . \
-    || error "buildx failed"
+  docker buildx build "$(_docker-build-args)" -t "${repo_name}:${TAG}" -f "${docker_file}" --platform ${platform} \
+     --cache-to type=registry,ref="${repo_name}:build-cache",mode=max \
+     --cache-from type=registry,ref="${repo_name}:build-cache" \
+     "$@" --push . || error "buildx failed"
   echo "Successfully built in $dir using docker file $docker_file"
 }
 


apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
...
+     --cache-to type=registry,ref="${repo_name}:build-cache",mode=max \
+     --cache-from type=registry,ref="${repo_name}:build-cache" \
+     "$@" --push . || error "buildx failed"
+  echo "Successfully built in $dir using docker file $docker_file"
 }
 
+_docker-repo-name() {
+  local docker_file="$1"
+  echo "Finding repo name in: ${docker_file}" >&2
+  local repo_name="$(grep '^# Repo name: ' "${docker_file}" | awk '{print $4}')"
```
alexaryn, line 106: A more robust approach might be:
```
local repo_name=$(sed -nr -e 's/^#\s?Repo\s+name:\s*(\w+)/\1/ip' "${docker_file}" | head -n 1)
```
It would save you from multiple copies and variations in formatting.  It goes well with your check that the word count is one, below.  In assignment context, you don't need the quotes around `$(...)` command substitution.

apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
...
+  if (( $(wc -w <<< ${repo_name}) != 1 )); then
+    echo "Unable to find repo name in ${docker_file}" 1>&2
+    exit 1
+  fi
+  echo "${repo_name}"
+}
+
+_docker-build-args() {
+  local branch="$(git status | head -1 | grep 'On branch ' | awk '{print $3}')"
```
alexaryn, line 115: head -n1 is better (fewer warnings).  I'd also use grep -i just in case (get it?).

apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
...
+  local rev="$(git rev-parse --short HEAD)"
+  local date="$(git show -s --format=%ci HEAD | sed 's/ /_/g')"
```
alexaryn, line 117: It's preferred to have `sed -e <program>`.

apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
...
+  local diff=unknown
+  if [[ $(git status | grep -c 'nothing to commit, working tree clean') = 1 ]]; then
```
alexaryn, line 119: Maybe `grep -i` but even that seems too dependent on the exact wording of the message.  Also, you might be able to use the exit code of grep directly, like:
```
if git status | grep -iq 'nothing to commit'; then
```

apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
...
+    diff=clean
+  else
+    diff="pending_changes_$(git diff HEAD | shasum | awk '{print $1}')"
```
alexaryn, line 122: Is the SHA important?  If so, SHA-1 is considered obsolete; so, you'd want `shasum -a 256`, but they're longer.  For a shorter checksum that's OK, try `md5sum`.

apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
...
+  fi
+  echo "--build-arg=GIT_BRANCH=${branch} --build-arg=GIT_COMMIT=${rev}--${date} --build-arg=GIT_DIFF=${diff}"
```
alexaryn, line 124: Why double-delimiters between rev and date?

alexaryn (commented):
A bunch of comments, but overall like it.

apps/integration/integration/automation/runtests.sh
alexaryn: Actually, my pipe to `head -n 1` may be unnecessary if you're checking for a single word right after.

apps/integration/integration/automation/runtests.sh
```diff
@@ -1,30 +1,52 @@
 #!/bin/bash
 
+TAG="$1"
+[[ -z "${TAG}" ]] && TAG="latest_rc"
```
eric-anderson, line 4: I would do TAG="integration_tests" by default.

apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
...
+}
+
+docker-build-hub() {
+  local docker_file="$1"
+  [[ -n "${docker_file}" ]] || error "missing ${docker_file}"
+  local repo_name="$(_docker-repo-name "${docker_file}")"
+  [[ -n "${repo_name}" ]] || error "empty repo name"
+  shift
+
+  local platform=linux/amd64,linux/arm64
```
eric-anderson, line 93: Do we want this? I would think we would build/test for the local platform.

apps/integration/integration/automation/runtests.sh
eric-anderson: We want to check for the specific message to guarantee the tree is clean (no unexpected files also).

apps/integration/integration/automation/runtests.sh
eric-anderson: Because date has -'s in it; and -- makes it easier to read.

apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
...
   echo "Yep, definitely handling test outputs. That's what this function does" >&2
+  local passed_tests="$1"
+  mv test-output.log "${QUERY_LOGFILE}"
+  [[ ${passed_tests} = 0 ]] && touch "${RUNDIR}/passed"
+  [[ ${passed_tests} != 0 ]] && touch "${RUNDIR}/failed"
+  aws s3 cp -r "${RUNDIR}" "s3://sycamore-ci/${ARCH}"
 }
 
 runtests() {
   docker system prune -f --volumes
```
eric-anderson, line 77: This seems risky since if someone runs it it will prune other stuff they might want; It also means that we get rid of the old volumes; it's also not guaranteed to clean things up if it's still in use.

I'd instead suggest that we generate a unique run id (use the date stamp?). and do the docker compose -p <uid> up. If we tag the images in the same way, then if there's a problem someone could go back and debug against the exact set of images. This would imply we don't want to push to docker hub since it's a lot of stuff.
That guarantees you're getting

apps/integration/integration/automation/runtests.sh
```diff
@@ -33,20 +55,73 @@ checkout_main_if_new() {
 
 build_containers() {
   echo "Yep, definitely building containers. That's what this function does" >&2
+  docker-build-hub apps/crawler/crawler/http/Dockerfile
```
eric-anderson, line 58: Do we want to build and push to docker hub or only build to local?  This seems like it's going to generate a lot of transit data. Moreover, given the way you're doing the testing, if we want to test on arm it would be a separate build/test run and then the two builds would clash.

apps/integration/integration/automation/runtests.sh
eric-anderson: We don't need to be very robust here. We control the input file.

apps/integration/integration/automation/runtests.sh
```diff
@@ -1,30 +1,52 @@
...
   echo "Get the newest git commits" >&2
   checkout_main_if_new
   local should_run=$?
   if [[ $should_run ]]; then
     echo "Changes detected. Running Tests" >&2
-    poetry install
-    build_containers
-    runtests
-    handle_outputs
+    poetry install > "${POETRY_LOGFILE}" 2>&1
```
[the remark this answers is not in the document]
HenryL27: well, I always want to get to the `handle_outputs`, so I don't think I want to die. I think I can `&&` all these things together to get the correct behavior though.
```bash
poetry install > "${POETRY_LOGFILE}" 2>&1 \
&& build_containers > "${DOCKER_LOGFILE}" 2>&1 \
&& runtests > "${PYTEST_LOGFILE}" 2>&1
```

apps/integration/integration/automation/runtests.sh
```diff
@@ -1,30 +1,52 @@
...
+    build_containers > "${DOCKER_LOGFILE}" 2>&1
+    runtests > "${PYTEST_LOGFILE}" 2>&1
+    local passed_tests=$?
+    handle_outputs passed_tests
```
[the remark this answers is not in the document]
HenryL27: you know it! (exit codes don't do anything weird when they become variables, right?)

apps/integration/integration/automation/runtests.sh
```diff
@@ -1,30 +1,52 @@
...
+}
+
 checkout_main_if_new() {
   old_sha="$(git rev-parse HEAD)"
-  git fetch origin main >&2
+  git fetch origin main > "${GIT_LOGFILE}"
   new_sha="$(git rev-parse FETCH_HEAD)"
   if [[ "${old_sha}" != "${new_sha}" ]]; then
-    git pull origin main >&2
+    git pull --rebase origin main >> "${GIT_LOGFILE}"
```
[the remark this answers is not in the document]
HenryL27: since I dont want to die, does
`{ echo "Working tree not clean" > "${GIT_LOGFILE}" && return 1; }`
do the right thing after the ||?

apps/integration/integration/automation/runtests.sh
HenryL27: ...why?

apps/integration/integration/automation/runtests.sh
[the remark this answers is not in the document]
HenryL27: Hmm, I need it to be okay with untracked files, bc I create them in apps/integration/runs.
will make this a `grep -c -e 'nothing to commit, working tree clean'  -e 'nothing added to commit but untracked files present'`

apps/integration/integration/automation/runtests.sh
```diff
@@ -1,30 +1,52 @@
...
+    build_containers > "${DOCKER_LOGFILE}" 2>&1
+    runtests > "${PYTEST_LOGFILE}" 2>&1
+    local passed_tests=$?
+    handle_outputs passed_tests
   else
     echo "No changes detected. Skipping integration tests" >&2
   fi
 }
 
+error() {
```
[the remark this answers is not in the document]
HenryL27: leaving as error but changing so it doesn't exit. I basically never want this script to exit early

apps/integration/integration/automation/runtests.sh
alexaryn: The definition of clean is zero modified files and zero untracked files.  The future-proof way to do this is:
```
if [[ -z $(git status --porcelain) ]]; then
```

apps/integration/integration/automation/runtests.sh
```diff
@@ -1,30 +1,52 @@
...
   if [[ ! -d ".git" ]]; then
     echo "Error: please run this script from sycamore root!" >&2
     exit 1
   fi
+  mkdir -p "${RUNDIR}"
+  echo "Building/testing tag ${TAG}" >&2
   echo "Get the newest git commits" >&2
   checkout_main_if_new
   local should_run=$?
   if [[ $should_run ]]; then
```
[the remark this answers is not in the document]
alexaryn: You can ditch the square brackets entirely.
```
$ mytrue() {
>   return 0
> }
$ if mytrue; then
>   echo yes
> fi
yes
```

apps/integration/integration/automation/runtests.sh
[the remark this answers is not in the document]
alexaryn: Let's not rely on specific English-language messages.  How about running `git status --porcelain | grep -vF '??'` and using `-z` to check for no output.

apps/integration/integration/automation/runtests.sh
alexaryn: Personally, I'd find it convenient that reasonably up-to-date images are easily available to test with.  Having them in DockerHub seems like the lowest-friction way.

apps/integration/integration/automation/runtests.sh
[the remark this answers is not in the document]
alexaryn: I believe the number just carries across the assignment faithfully.  It's when used as a conditional that zero becomes true.

apps/integration/integration/automation/runtests.sh
alexaryn: `git branch --show-current` ?

apps/integration/integration/automation/runtests.sh
alexaryn: Mostly hygiene.  If the program happens to have something that might get interpreted as a command-line option, -e protects it.  Also, you can do multi-expression scripts with multiple -e <program> directives.

apps/integration/integration/automation/runtests.sh
alexaryn: There are better formats for date available, for instance ISO 8601 basic.

alexaryn:
Is it time to rename `runtests.sh` to something more reflective of it's current effects?  Also, if we drop the suffix, we'd be free to rewrite this in Python without identifying and changing its callers.

apps/integration/integration/automation/runtests.sh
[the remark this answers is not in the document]
HenryL27: alternatively I can just add apps/integration/runs to the gitignore and then take the simpler -z git status --procelain you gave me earlier

HenryL27:
you know I love rewriting bash into python!

alexaryn (commented):
Looking better.  I'll let Eric look, too.

I'm still not super-happy calling it runtests when it does so much more.  Would things improve if it were split into different scripts (or verb arguments) for the various steps: building, pushing, etc.

apps/integration/integration/automation/runtests
```diff
@@ -0,0 +1,151 @@
...
+  echo "ERROR: $@" >&2
+}
+
+checkout_main_if_new() {
+  old_sha="$(git rev-parse HEAD)"
+  git fetch origin main > "${GIT_LOGFILE}" 2>&1
+  new_sha="$(git rev-parse FETCH_HEAD)"
+  if [[ "${old_sha}" != "${new_sha}" ]]; then
+    [[ -z $(git status --porcelain) ]] \
+      || { echo "Working tree not clean" > "${GIT_LOGFILE}" && return 1; }
```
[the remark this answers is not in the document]
HenryL27: Not sure, I copied it from the stack overflow I found on grouping. Although I guess I could clean this up a lot by turning the -z into a -n and the || into && and ungrouping it. echo > file will never fail, right?

HenryL27:
I guess `git mv` and edit a file in one commit doesn't do the right thing w.r.t. comments. ugh, sorry

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
...
+handle_outputs() {
+  echo "Handling test outputs" >&2
+  local passed_tests="$1"
+  mv test-output.log "${QUERY_LOGFILE}"
+  if [[ ${passed_tests} = 0 ]]; then
+    touch "${RUNDIR}/passed"
+  else
+    touch "${RUNDIR}/failed"
+  fi
+  aws s3 cp -r "${RUNDIR}" "s3://sycamore-ci/${ARCH}"
```
eric-anderson, line 123: Having arch be the top level thing to me feels weird. I'd suggest top level is the git version and below that is arch since I'd expect amd64 & arm64 builds for some rev have more in common.

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
...
+main() {
+  [[ ! -d ".git" ]] && die "Please run this script from sycamore root!"
+  mkdir -p "${RUNDIR}"
+  echo "Building/testing tag ${TAG}" >&2
+  echo "Get the newest git commits" >&2
+  if checkout_main_if_new; then
+    echo "Changes detected. Running Tests" >&2
+    poetry install --no-root > "${POETRY_LOGFILE}" 2>&1 \
+        && { [[ $SKIP_BUILD ]] || build_images > "${DOCKER_LOGFILE}" 2>&1; } \
+        && { [[ $SKIP_TESTS ]] || runtests > "${PYTEST_LOGFILE}" 2>&1; }
```
eric-anderson, line 69: I'd do:
```
 ...
 && touch "${RUNDIR}/passed_tests"
[[ -f "${RUNDIR}/passed_tests" ]] || touch "${RUNDIR}/failed_tests"
```

If someone put an echo "done with tests" above the local passed_tests=$? line, it would make all the tests appear to pass in the current form.

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
...
+  && docker-push-hub apps/opensearch/Dockerfile \
+  && docker-push-hub apps/jupyter/Dockerfile.buildx \
+  && docker-push-hub apps/demo-ui/Dockerfile.buildx \
+  && docker-push-hub apps/remote-processor-service/Dockerfile.buildx \
+  && return 0
+  return 1
+}
+
+runtests() {
+  docker system prune -f --volumes
```
eric-anderson, line 140: This is a pretty big hammer; my guess is the purpose is to clean up integration test volumes, but it will also clean up build caches and other people's stuff.
I'd suggest:
```
docker volume rm integration_crawl_data integration_jupyter_data integration_opensearch_data
docker compose -p integration up reset
```

That will clean up integration specific volumes but leave everything else.
Optionally clean them after successful testing and verify there are no volumes named `*integration*`

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
...
+  echo
+  echo "Building in sycamore and pushing to docker hub with repo name '${repo_name}'"
+  docker buildx build "$(_docker-build-args)" -t "${repo_name}:${TAG}" -f "${docker_file}" \
+     --cache-to type=registry,ref="${repo_name}:build-cache",mode=max \
+     --cache-from type=registry,ref="${repo_name}:build-cache" \
+     "$@" --load . || { error "buildx failed" && return 1;}
+  echo "Successfully built using docker file $docker_file"
+}
+
+docker-push-hub() {
```
eric-anderson, line 166: If you run this on both arm & amd, what shows up in dockerhub?  I was building both architectures in a single go because I thought that was necessary for it to show up cleanly (may not matter for integration testing)

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
...
+  local docker_file="$1"
+  [[ -n "${docker_file}" ]] || { error "missing ${docker_file}" && return 1;}
```
eric-anderson, line 168: slightly safer to write `{ error "..."; return 1 }`
That way if error returns an error you don't accidentally continue.

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
...
+}
+
+_docker-repo-name() {
+  local docker_file="$1"
+  echo "Finding repo name in: ${docker_file}" >&2
+  local repo_name="$(grep '^# Repo name: ' "${docker_file}" | awk '{print $4}')"
+  if (( $(wc -w <<< ${repo_name}) != 1 )); then
+    echo "Unable to find repo name in ${docker_file}" 1>&2
+    exit 1
+  fi
```
eric-anderson, line 185: `[[ "${repo_name}" = *private* ]] && die "Private repo ${repo_name} disallowed"`

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
HenryL27: wdym?
The apps/integration/integration comes from the python project name being equal to the directory name - that pattern is all over this repo.

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
...
+  exit 1
+}
+
+checkout_main_if_new() {
+  old_sha="$(git rev-parse HEAD)"
+  git fetch origin main > "${GIT_LOGFILE}" 2>&1
+  new_sha="$(git rev-parse FETCH_HEAD)"
+  if [[ "${old_sha}" != "${new_sha}" ]]; then
+    [[ -z $(git status --porcelain) ]] \
+      || { echo "Working tree not clean" > "${GIT_LOGFILE}" && return 1; }
```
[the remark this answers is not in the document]
HenryL27: Ok yeah. My initial thought was that hiving it continue and skip all the stuff would be fine but I guess you probably do want this case to outright kill itself

apps/integration/integration/automation/integrate
HenryL27: There's not, as far as I can tell, a good way to get testcontainers to do project names...

apps/integration/integration/automation/integrate
HenryL27: do we expect this? but ok

apps/integration/integration/automation/integrate
```diff
@@ -0,0 +1,202 @@
+#!/bin/bash
+
+# Parse args
+SKIP_BUILD=0
```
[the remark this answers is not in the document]
HenryL27: I think it needs to be negative conditions. the logic I need is (if I want to be able to chain each step with &&'s) is:
```
if (action_condition) -> (action)
else                  -> true
```
with, e.g. `DO_BUILD`, I have 
```bash
{ [[ $DO_BUILD ]] && build; } \
&& { [[ $DO_TESTS ]] && tests; } \
&& etc...
```
but if `DO_BUILD` is false this breaks my pipeline. With negative conditions I can do
```bash
{ [[ $SKIP_BUILD ]] || build; } \
&& { [[ $SKIP_TESTS ]] || tests; } \
&& etc...
```
which has the intended selectivity behavior. To get the right behavior with positive variable I think I need to negate them in the conditions, so SKIP seems cleaner to me

PR #351, opened 2024-04-12
Fix malformed text from beautiful soup.

bohou-aryn:
Seems the problem of below

```
<div><span>hello</span></div><div><span>world</span></div> ->
"helloworld" instead of "hello" "world"
```
is due to no separator is added when concatenate string when getting text

### DIFF
--- a/lib/sycamore/sycamore/tests/unit/transforms/test_partition.py
+++ b/lib/sycamore/sycamore/tests/unit/transforms/test_partition.py
@@ -78,7 +78,7 @@ class TestPartition:
             (
                 HtmlPartitioner(),
                 TEST_DIR / "resources/data/htmls/wikipedia_binary_search.html",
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/tests/unit/transforms/test_partition.py
```diff
@@ -93,7 +93,7 @@ def test_html_partitioner(self, partitioner, read_local_binary, expected_partiti
             (
                 HtmlPartitioner(extract_tables=True),
                 TEST_DIR / "resources/data/htmls/wikipedia_binary_search.html",
-                77,
+                74,
```
HenryL27, line 96: I must be missing something but why does this make the number go down?

HenryL27 (approved):
lgtm

PR #353, opened 2024-04-18
Use real SSL certificate for OpenSearch HTTP.

alexaryn:
Also moved some global stuff out of main function and to top of file.  The authority.pem is the ISRG X1 root certificate, which Let's Encrypt relies on.  It can be identified via: openssl x509 -text -noout -in authority.pem


apps/opensearch/opensearch.yml
```diff
@@ -9,8 +9,8 @@ plugins.security.ssl.transport.pemcert_filepath: node-cert.pem
 plugins.security.ssl.transport.pemkey_filepath: node-key.pem
 plugins.security.ssl.transport.pemtrustedcas_filepath: cacert.pem
 plugins.security.ssl.http.enabled: true
-plugins.security.ssl.http.pemcert_filepath: node-cert.pem
-plugins.security.ssl.http.pemkey_filepath: node-key.pem
-plugins.security.ssl.http.pemtrustedcas_filepath: cacert.pem
+plugins.security.ssl.http.pemcert_filepath: http-cert.pem
```
pparmar30, line 12: why http and not https?

apps/opensearch/sycamore-opensearch.sh
```diff
@@ -320,14 +322,29 @@ create_certificates() {
         chmod 600 "data/${X}"
         ln -sfn "../data/${X}" "config/${X}"
     done
+
+    # 4. Use provided certificates for HTTP if possible
+    if [[ -f ${ARYN_ETC}/hostcert.pem && -f ${ARYN_ETC}/hostkey.pem ]]; then
```
pparmar30, line 327: Can we check for presence of provided cert before we do anything related to using self-sign certs?

apps/opensearch/Dockerfile
```diff
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1
```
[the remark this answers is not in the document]
alexaryn: It didn't work for me without it.  https://forums.docker.com/t/problem-with-add-git-ref-how-to-switch-to-labs-channel/139121

apps/opensearch/opensearch.yml
alexaryn: That's the naming OpenSearch uses.  I'm just matching it.

apps/opensearch/sycamore-opensearch.sh
alexaryn: No.  We need self-signed certs for node and admin.

pparmar30 (approved):

PR #357, opened 2024-04-22
Update data model for table structure recognition.

bsowell:
This PR introduces a logical Table class that can be used to represent the data in a document table in a format independent way. It is based on representations used by TableTransformers and Textract, among others. This commit just adds the basic data structure and some format conversions and tests. Actual table understanding models will be added in a follow-up PR.

This also includes some enhancements to the BoundingBox class so that it can be used with a variety of table models.

### DIFF
--- a/lib/sycamore/sycamore/data/table.py
+++ b/lib/sycamore/sycamore/data/table.py
@@ -166,7 +166,7 @@ class Table:
 
         return ET.tostring(table, encoding="unicode")
 
-    U = TypeVar("U", bound=Union[Image, ImageDraw.ImageDraw])
+    U = TypeVar("U", bound=Union[Image.Image, ImageDraw.ImageDraw])
 
     # TODO: This currently assumes that the bounding rectangles are on the same page.
     def draw(self, target: U) -> U:
@@ -175,7 +175,7 @@ class Table:
         if isinstance(target, ImageDraw.ImageDraw):
             canvas = target
             width, height = target.im.size
-        elif isinstance(target, Image):
+        elif isinstance(target, Image.Image):
             canvas = ImageDraw.Draw(target)
             width, height = target.size
 
--- a/lib/sycamore/sycamore/data/table.py
+++ b/lib/sycamore/sycamore/data/table.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict
 from dataclasses import dataclass, field
+import itertools
 from typing import Any, Optional
 import xml.etree.ElementTree as ET
 
@@ -22,10 +23,20 @@ class TableCell:
     # Model/format specific properties
     properties: dict[str, Any] = field(default_factory=dict)
 
+    def __post_init__(self):
+        # Validate that row and column spans are contiguous.
+        for a, b in itertools.pairwise(self.rows):
+            if a + 1 != b:
+                raise ValueError(f"Found non-contiguous rows in {self}.")
+
+        for a, b in itertools.pairwise(self.cols):
+            if a + 1 != b:
+                raise ValueError(f"Found non-contiguous cols in {self}.")
+
 
 class Table:
     def __init__(self, cells: list[TableCell], caption: Optional[str] = None):
-        self.cells: list[TableCell] = cells
+        self.cells: list[TableCell] = sorted(cells, key=lambda tc: (min(tc.rows), min(tc.cols)))
         self.caption = caption
         self.num_rows = max(max(c.rows) for c in self.cells) + 1
         self.num_cols = max(max(c.cols) for c in self.cells) + 1
@@ -38,18 +49,15 @@ class Table:
             return False
 
         # TODO: Yikes! this is expensive. Do the sorting in init or something.
-        if self.sorted_cells() != other.sorted_cells():
+        if self.cells != other.cells:
             return False
 
         return True
 
     def __hash__(self):
-        return hash((self.num_rows, self.num_cols, self.cells))
+        return hash((self.cells))
 
-    def sorted_cells(self):
-        return sorted(self.cells, key=lambda tc: (min(tc.rows), min(tc.cols)))
-
-    # This algorithm is modified from the TableTransformers code. The conversion to Pandas/CSV is
+    # This algorithm is modified from the TableTransformers code. The conversion to Pandas/CSV
     # is necessarily lossy since these formats requires tables to be square and don't support
     # cells spanning multiple rows/columns.
     #
@@ -129,7 +137,7 @@ class Table:
             caption_cell.text = self.caption
 
         # TODO: We should eventually put these in <thead> and <tbody> tags.
-        for cell in self.sorted_cells():
+        for cell in self.cells:
             cell_attribs = {}
 
             rowspan = len(cell.rows)
--- a/lib/sycamore/sycamore/data/table.py
+++ b/lib/sycamore/sycamore/data/table.py
@@ -1,13 +1,22 @@
 from collections import OrderedDict
 from dataclasses import dataclass, field
-import itertools
-from typing import Any, Optional
+from typing import Any, Optional, TypeVar, Union
 import xml.etree.ElementTree as ET
 
 from sycamore.data import BoundingBox
 from PIL import Image, ImageDraw
 import numpy as np
 from pandas import DataFrame
+
+
+# This is part of itertools in 3.10+.
+# Adding here to support 3.9
+def _pairwise(iterable):
+    iterator = iter(iterable)
+    a = next(iterator, None)
+    for b in iterator:
+        yield a, b
+        a = b
 
 
 # This data model is similar to that used by Textract and TableTransformers.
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+# This data model is similar to that used by Textract and TableTransformers.
+# It is intended to be a common intermediate representation for a variety of
+# table structure recognition models.
+@dataclass(frozen=True)
+class TableCell:
+    content: str
+    rows: list[int]
+    cols: list[int]
```
alexaryn, line 19: Why are these lists?  Is this for spanning?  Can a list end up disjointed like [1,2,5,6]?

lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,164 @@
...
+        self.num_cols = max(max(c.cols) for c in self.cells) + 1
+
+    def __eq__(self, other):
+        if type(other) is not type(self):
+            return False
+
+        if self.num_rows != other.num_rows or self.num_cols != other.num_cols:
+            return False
+
+        # TODO: Yikes! this is expensive. Do the sorting in init or something.
```
alexaryn, line 40: Yes!  Canonical representation seems like a good invariant.

lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,164 @@
...
+        if self.sorted_cells() != other.sorted_cells():
+            return False
+
+        return True
+
+    def __hash__(self):
+        return hash((self.num_rows, self.num_cols, self.cells))
```
alexaryn, line 47: Hash(cells) should be sufficient, no?  And only if the order is canonical, right?  Otherwise, there's a divergence between eq(x,y) and eq(hash(x),hash(y)).

lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def sorted_cells(self):
+        return sorted(self.cells, key=lambda tc: (min(tc.rows), min(tc.cols)))
+
+    # This algorithm is modified from the TableTransformers code. The conversion to Pandas/CSV is
+    # is necessarily lossy since these formats requires tables to be square and don't support
```
alexaryn, line 53: is is

lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,164 @@
...
+    # since they are not nativey supported in CSV. The standard approaches are (1) to duplicate
+    # the cell value in each row/column or (2) to put the value in the first row/column in the cell
+    # and empty values in the other. TableTransformers do the former, The textractor library supports
+    # both, but defaults to the latter. Since our primary use case is preparing context for LLMs,
+    # we speculate that duplication may create confusion, so we default to only displaying a cells
+    # content for the first row/column for which it is applicable. The exception is for header rows,
+    # where we duplicate values to each columnn to ensure that every column has a fully qualified header.
+    def to_pandas(self) -> DataFrame:
+        # Find all row nums containing cells marked as headers.
+        header_rows = sorted(set((row_num for cell in self.cells for row_num in cell.rows if cell.is_header)))
```
alexaryn, line 66: Does sorting put headers at the top?

lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,164 @@
...
+import numpy as np
+from pandas import DataFrame
+
+
+# This data model is similar to that used by Textract and TableTransformers.
+# It is intended to be a common intermediate representation for a variety of
+# table structure recognition models.
+@dataclass(frozen=True)
+class TableCell:
+    content: str
```
alexaryn, line 17: Any support for nested tables?

alexaryn (commented):
My main concern is about a canonical representation that makes eq() efficient and allows hashing to be valid.

lib/sycamore/sycamore/data/table.py
bsowell: It is for spanning. A single cell should not be disjoined. I can try to add some validation logic.

lib/sycamore/sycamore/data/table.py
bsowell: I can move sorting to init. I'm debating whether the table should be immutable, in which case that's fine. The only caveat is that the models do a bunch of manipulation, so it may impact how early I convert to this Table object vs how many copies are required/etc.

lib/sycamore/sycamore/data/table.py
bsowell: Yeah, you are right. I will update.

lib/sycamore/sycamore/data/table.py
bsowell: No, not necessarily. Technically (in html for instance) you can have headers throughout the table -- for example "section" headers for groups of cells. The goal here is to find the set of contiguous header rows at the beginning of the document. We squash these into a header for pandas/csv. Those formats don't support headers elsewhere in the table, so we ignore those for the purposes of conversion.

lib/sycamore/sycamore/data/table.py
bsowell: No, not right now. Honestly I didn't think about it because none of the detection models I know of support them. I would suggest we defer this until it comes up.

lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,187 @@
...
+                        for col in cell.cols:
+                            table_array[row, col] = ""
+
+                else:
+                    table_array[cell.rows[0], cell.cols[0]] = cell.content
+                    for row in cell.rows[1:]:
+                        for col in cell.cols[1:]:
+                            table_array[row, col] = ""
+
+        header = table_array[: max_header_prefix_row + 1, :]
```
bohou-aryn, line 120: does this handle for cases that left most column is also header, or do we want such case handled?

bsowell: Interesting. I did not handle this case. It looks like Pandas does support this, though I guess it won't impact the CSV representation. Happy to play around with it, though not sure it will matter in any of our initial use cases.

lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,187 @@
...
+    # This algorithm is modified from the TableTransformers code. The conversion to Pandas/CSV
+    # is necessarily lossy since these formats requires tables to be square and don't support
+    # cells spanning multiple rows/columns.
+    #
+    # One of the main decisions here is how to render cells that span multiple rows and columns,
+    # since they are not nativey supported in CSV. The standard approaches are (1) to duplicate
+    # the cell value in each row/column or (2) to put the value in the first row/column in the cell
+    # and empty values in the other. TableTransformers do the former, The textractor library supports
+    # both, but defaults to the latter. Since our primary use case is preparing context for LLMs,
+    # we speculate that duplication may create confusion, so we default to only displaying a cells
```
bohou-aryn, line 78: How the LLM impact our decision here? You mean duplicates cause problem? This matters with the format we send to LLM in either delimited csv format or html format, seems the later one has more info loss.
Besides, do you feel it make sense to provide both modes by taking some knob?

bsowell: I thought I convinced myself that duplicates might be more confusing – for example imagine you wanted to sum up the cells in a column, but you duplicated values – you would get the wrong answer. Not completely sure though. I was just trying to pick something, and this is at least consistent with what we get from textract today. 

I think you lose less structural information when you convert to html vs csv, but LLMs may be less able to deal with complex html. Not sure. I'd like to do a proper benchmark, eventually. 

Happy to add an option. My instinct is that very few people are going to want to tweak configuration parameters at this level, but if we see significantly different behavior it might be worth it. Are you okay if we defer the knob to a later commit?

bohou-aryn: sure.

bohou-aryn (approved):

lib/sycamore/sycamore/data/table.py
baitsguy: should we just call this text_representation as well? Unless it has a different meaning

lib/sycamore/sycamore/data/table.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __hash__(self):
+        return hash((self.cells))
+
+    # This algorithm is modified from the TableTransformers code. The conversion to Pandas/CSV
+    # is necessarily lossy since these formats requires tables to be square and don't support
+    # cells spanning multiple rows/columns.
+    #
+    # One of the main decisions here is how to render cells that span multiple rows and columns,
+    # since they are not nativey supported in CSV. The standard approaches are (1) to duplicate
```
baitsguy, line 74: do we need to stick to csv? Can we do json or something and only do csv style conversion when necessary?

baitsguy (approved):
Looks good, lots of code in here so it's probably worth just getting this in so we have something. I had a question about using pandas/csv/fixed structure objects vs json to store the contents

lib/sycamore/sycamore/data/table.py
bsowell: Debating this. In hindsight I wish we hadn't called it text_representation for Documents, and now I'm not sure whether to go for consistency or clarity.

lib/sycamore/sycamore/data/table.py
bsowell: So in this PR we aren't actually storing anything as CSV -- it's only provided as a target for conversion. The question I guess is what do we embed/store in OpenSearch. I'm open to suggestions for that. My impression was that LLMs/embedding models would probably do better with CSV/HTML since they are common ways to represent tables, but happy to try something else.

PR #358, opened 2024-04-24
Add TableStructureExtractor interface and TableTransformer impl.

bsowell:
This adds a new interface for extracting the structure of tables once they have been recognized. The initial implementation leverages the TableTransformer model from Microsoft.

Sample usage:

ds = context.read.binary(paths=["/Users/bsowell/Downloads/test.html.pdf"], binary_format="pdf")\
                 .partition(partitioner=SycamorePartitioner(extract_table_structure=True))\
                 .flat_map(split_and_convert_to_image)\
                 .map_batch(DrawBoxes, f_constructor_args=[font_path], f_constructor_kwargs={"draw_table_cells": True})\
                 .write.files("./new_output", filename_fn=image_page_filename)


lib/sycamore/sycamore/data/element.py
```diff
@@ -69,14 +76,17 @@ def __init__(
...
+        tokens: Optional[list[dict[str, Any]]] = None,
         **kwargs,
     ):
         super().__init__(element, **kwargs)
         self.data["type"] = "table"
-        self.data["properties"] = {}
         self.data["properties"]["title"] = title
         self.data["properties"]["columns"] = columns
         self.data["properties"]["rows"] = rows
+        self.data["table"] = table
```
bsowell, line 88: It is maybe a little bit awkward to have both the table property and the existing columns/rows properties. Eventually we should rewrite everything else in terms of table, but I don't think we want to modify existing code quite yet. Open to suggestions.

lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -0,0 +1,143 @@
...
+
+        element.table = table
+        return element
+
+
+DEFAULT_TABLE_STRUCTURE_EXTRACTOR = TableTransformerStructureExtractor()
+
+
+class ExtractTableStructure(Transform):
+    def __init__(self, child: Node, table_structure_extractor: TableStructureExtractor, **resource_args):
```
bohou-aryn, line 135: Just curious, do we need this transform independent of partition.

lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -0,0 +1,143 @@
...
+        return tokens
+
+    def extract(self, element: TableElement, doc_image: Image.Image) -> TableElement:
+        # We need a bounding box to be able to do anything.
+        if element.bbox is None:
+            return element
+
+        width, height = doc_image.size
+
+        structure_model = TableTransformerForObjectDetection.from_pretrained(self.model)
```
bohou-aryn, line 81: In this case, we are reloading the model each time this called for a table?

lib/sycamore/sycamore/functions/document.py
```diff
@@ -81,16 +82,17 @@ class DrawBoxes:
...
-    def __init__(self, font_path: str, default_color: str = "blue"):
+    def __init__(self, font_path: str, default_color: str = "blue", draw_table_cells: bool = True):
         self.font = ImageFont.truetype(font_path, 20)
         self.color_map = {
             "Title": "red",
             "NarrativeText": "blue",
             "UncategorizedText": "blue",
             "ListItem": "green",
-            "Table": "orange",
+            "table": "orange",
```
bohou-aryn, line 92: this intentional or typo?

lib/sycamore/sycamore/data/element.py
bsowell: What do you mean by this?

lib/sycamore/sycamore/transforms/table_structure/extract.py
bsowell: I don't think it's necessary. I just thought it might be useful if you wanted to run table extraction model but were using a different partitioner or something. I can remove it if you feel strongly.

lib/sycamore/sycamore/functions/document.py
bsowell: It turned out the TableElement had the type set to lowercase, so I changed it to lower case here. In hindsight I think I should make it upper case everywhere to match other types. I'll make that change.

lib/sycamore/sycamore/transforms/table_structure/extract.py
bsowell: Ah yes, you are right. I see how we could change it to load once per document, which I will do. Do you think there's a better way to do it?

lib/sycamore/sycamore/data/element.py
bsowell: My next step is to support converting HTML tables to the `Table` class. I started on that, but it's a bit tricky, so I deferred it to a later PR b/c it's not necessary for the blog post.

lib/sycamore/sycamore/functions/document.py
bsowell: Hmm. Actually maybe I was wrong that we standardize on upper-case. There are still a bunch of tests that appear to assume lowercase. Maybe I'll leave this for now, and we can see about cleaning this up holistically.

lib/sycamore/sycamore/data/element.py
bohou-aryn: sorry, a bit confused myself also. I thought you try to get rid of columns/rows, and use table instead? For html tables, do we have such use case? I thought it was more on into html, but not from html.

lib/sycamore/sycamore/transforms/table_structure/extract.py
bohou-aryn: We discuss a bit before on partitioner, I feel the best way would be put in actor class and only call once per worker.

lib/sycamore/sycamore/transforms/table_structure/extract.py
bohou-aryn: no strong opinion here.

lib/sycamore/sycamore/data/element.py
bsowell: One of my goals is to have a common representation of a table across different formats. That way, for example, when we come up with a good way to represent a table for embedding, we can use it independently of the source format. I think that ideally we would have people access rows and columns via the Table interface, rather than storing them separately for some formats in the Element.

lib/sycamore/sycamore/transforms/table_structure/extract.py
bsowell: I want to make sure I understand the proposal here. Are you suggesting we convert the partition transform to use a map class instead of a function for the callable? And then I guess the init for that would initialize the partitioners and any models. Is that right?

bohou-aryn: yes, that's what I mean. This would change the docset partition API, we pass extractor parameters instead of extractor instance. If you have concern on this, I'm good with document level model initialization.

bsowell: I don't have any major objections to this. I am wondering if we want to do it in this PR or do a followup. There is some urgency for this change b/c we want to get a blog post out, but I don't want to steamroll over your comments. I can sync with you today.

bohou-aryn (approved):

PR #362, opened 2024-04-30
Use chown in our copy commands to make sure all files are owned by app

eric-anderson:
Also add check to Dockerfile to make sure that remains true.


HenryL27 (approved):
well foo you!

PR #365, opened 2024-05-01
Add the ability to summarize images to partitioned docsets.

bsowell:
Adds an option to the SycamorePartitioner to prune and attach the actual image data to the underlying elements, and a separate transform to summarize image text using gpt-4-turbo.

Example usage:

    context = sycamore.init()
    doc = context.read.binary(paths=paths, binary_format="pdf")\
                     .partition(partitioner=SycamorePartitioner(extract_images=True))\
                     .transform(SummarizeImages)\
                     .show()


lib/sycamore/sycamore/utils/image_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        return image.tobytes()
+
+    iobuf = BytesIO()
+    image.save(iobuf, format=format)
+    return iobuf.getvalue()
+
+
+def base64url(image: Image.Image) -> str:
+    encoded_image = image_to_bytes(image, "PNG")
+    return f"data:image/png/;base64,{base64.b64encode(encoded_image).decode('utf-8')}"
```
bohou-aryn, line 37: small one, if this bound to openai prompt format, we could move this or give a specific name?

bohou-aryn (approved):

lib/sycamore/sycamore/utils/image_utils.py
bsowell: It's actually a standard URL format: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs, but I agree it's a bit weird and I don't know if we will use it anywhere else. I'll update the name to try to make it a bit clearer.

PR #366, opened 2024-05-01
Don't say localhost if it's not going to work.

alexaryn:


HenryL27 (approved):
is this the only nb that does this?

PR #369, opened 2024-05-02
Jupyter container improvements

eric-anderson:
* Add aws cli to jupyter container
* Remove URL from notebooks since localhost won't work in cloud service


jonfritz (approved):
LGTM

PR #370, opened 2024-05-02
Fix typos in README

hsm207:


HenryL27 (approved):
Thanks! Guess I missed these when moving stuff around

PR #377, opened 2024-05-06
Handle deserializing JSON documents when elements is None.

bsowell:
This was broken when we switched to using create_element.


HenryL27 (approved):
That would do it

PR #378, opened 2024-05-06
Bump sycamore version to 0.1.16

bsowell:


HenryL27 (approved):
You're not gonna make me rebuild the containers so they think they have 0.1.16 in them, are you?

PR #387, opened 2024-05-11
lineage 1/n: add support for metadata.

eric-anderson:
* Add new MetadataDocument subclass that stores metadata and raises errors if tried to be used as a document.
* Add BaseMap class that processes documents, passes through metadata, and automatically generate lineage links.

### DIFF
--- a/lib/sycamore/sycamore/data/document.py
+++ b/lib/sycamore/sycamore/data/document.py
@@ -1,5 +1,6 @@
 from collections import UserDict
 from typing import Any, Optional
+import uuid
 
 from sycamore.data import BoundingBox, Element
 from sycamore.data.element import create_element
@@ -16,6 +17,9 @@ class Document(UserDict):
             from pickle import loads
 
             document = loads(document)
+            if "metadata" in document:
+                raise ValueError("metadata must be unserialized with Document.unserialize not Document.__init__")
+
         super().__init__(document, **kwargs)
         if "properties" not in self.data:
             self.data["properties"] = {}
@@ -24,6 +28,9 @@ class Document(UserDict):
             self.data["elements"] = []
         else:
             self.data["elements"] = [create_element(**element) for element in self.data["elements"]]
+
+        if "lineage_id" not in self.data:
+            self.update_lineage_id()
 
     @property
     def doc_id(self) -> Optional[str]:
@@ -34,6 +41,15 @@ class Document(UserDict):
     def doc_id(self, value: str) -> None:
         """Set the unique identifier of the document."""
         self.data["doc_id"] = value
+
+    @property
+    def lineage_id(self) -> str:
+        """A unique identifier for the document in its lineage."""
+        return self.data["lineage_id"]
+
+    def update_lineage_id(self):
+        """Update the lineage ID with a new identifier"""
+        self.data["lineage_id"] = str(uuid.uuid4())
 
     @property
     def type(self) -> Optional[str]:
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/data/document.py
```diff
@@ -155,18 +171,104 @@ def deserialize(raw: bytes) -> "Document":
...
+
+    @property
+    def metadata(self) -> dict[str, Any]:
+        """Internal metadata babout A collection of system or customer defined metadata, for instance,
+        a PDF document might have title and author metadata."""
+        return self.data.get("metadata", {})
+
+    @metadata.setter
+    def metadata(self, metadata: dict[str, Any]):
+        """Set all the proprites for this document."""
```
bsowell, line 263: minor: proprites -> properties

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -0,0 +1,127 @@
...
+
+    return decorator
+
+
+class BaseMap(UnaryNode):
+    """
+    BaseMap abstracts away MetadataDocuments from all other transforms.
+    """
+
+    def __init__(
```
bsowell, line 30: Once this is used everywhere, does this replace all of the map functions in https://github.com/aryn-ai/sycamore/blob/main/lib/sycamore/sycamore/transforms/map.py?

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -0,0 +1,127 @@
...
+        args = _noneOr(self._args, tuple())
+        kwargs = _noneOr(self._kwargs, {})
+
+        @rename(name)
+        def ray_callable(ray_input: dict[str, np.ndarray]) -> dict[str, list]:
+            return BaseMap._process_ray(ray_input, name, lambda d: f(d, *args, **kwargs))
+
+        return ray_callable
+
+    def _map_class(self):
```
bsowell, line 76: One thing I'm trying to understand is whether this has any impact on when Ray will create an actor or a task? At the moment I believe we use an actor when you try to map a class (which I think is from [here](https://github.com/aryn-ai/sycamore/blob/13322f80ec3610063b1073b14b1d006f44d0e8c1/lib/sycamore/sycamore/transforms/map.py#L37)). I don't know that we have to do it exactly the way we do it now, but would be good to understand if there are changes.

bsowell (approved):
Mostly minor typos. One more substantive question that might be more relevant for the next PR. I don't need to see this one again unless you think there is something we should discuss.

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -0,0 +1,127 @@
...
+
+def rename(new_function_name: str):
+    def decorator(f):
+        f.__name__ = new_function_name
+        return f
+
+    return decorator
+
+
+class BaseMap(UnaryNode):
```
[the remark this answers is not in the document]
eric-anderson: I went with BaseMapTransform.

lib/sycamore/sycamore/data/document.py
```diff
@@ -155,18 +171,104 @@ def deserialize(raw: bytes) -> "Document":
...
+        raise ValueError("MetadataDocument does not have properties")
+
+    @properties.setter
+    def properties(self, properties: dict[str, Any]):
+        raise ValueError("MetadataDocument does not have properties")
+
+    @property
+    def metadata(self) -> dict[str, Any]:
+        """Internal metadata babout A collection of system or customer defined metadata, for instance,
+        a PDF document might have title and author metadata."""
```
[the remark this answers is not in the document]
eric-anderson: Fixed. Looks like I copied and pasted and just stopped filling it in.

lib/sycamore/sycamore/data/document.py
```diff
@@ -16,6 +17,9 @@ def __init__(self, document=None, /, **kwargs):
             from pickle import loads
 
             document = loads(document)
+            if "metadata" in document:
+                raise ValueError("metadata must be unserialized with Document.unserialize not Document.__init__")
```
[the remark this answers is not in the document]
eric-anderson: fixed

lib/sycamore/sycamore/data/document.py
eric-anderson: fixed.

lib/sycamore/sycamore/transforms/base.py
eric-anderson: As discussed, probably not. The map functions in map.py become thin wrappers around this class.

lib/sycamore/sycamore/transforms/base.py
eric-anderson: It is not intended to have an impact. In class mode it makes a single instance and calls it repeatedly (as verified by test). I think that's what the old code also does.

PR #388, opened 2024-05-13
Add TimeTrace and instrument major pieces of code.

alexaryn:


lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -75,13 +76,15 @@ def __init__(
         self._prompt_formatter = prompt_formatter
 
     def extract_entity(self, document: Document) -> Document:
+        tt = TimeTrace("OaExtract")
```
eric-anderson, line 79: "entityExtract"?  I don't know what 0a is.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -152,6 +156,9 @@ def merge_elements(self, document: Document) -> Document:
         if len(document.elements) < 1:
             return document
 
+        tt = TimeTrace("mergeMarked")
```
eric-anderson, line 159: wouldn't you want this earlier to capture the early return?  Also I think the right if test on line 156 is <= 1. There's also nothing to merge if there's only one element.

lib/sycamore/sycamore/utils/time_trace.py
```diff
@@ -0,0 +1,61 @@
...
+        self.name = name.encode()
+        self._setup()
+
+    def __enter__(self):
+        self.start()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.end()
+
+    def start(self):
```
eric-anderson, line 21: `# TODO: replace all of this logic with metadata once that gets in`

Normally I'd complain a bunch about how all this works, but it all becomes obsolete once metadata goes in, and we'll want metadata since that will also work on multi-node.  It's unclear to me how much of the instrumentation will remain once we have metadata. Almost all of it is whole operator instrumentation which we should do automagically.

lib/sycamore/sycamore/transforms/extract_entity.py
alexaryn: Oa = OpenAI

lib/sycamore/sycamore/transforms/merge_elements.py
alexaryn: It's probably academic if we include that 'if' that's rarely taken.  For the sake of visualization, there would be zero pixels.  I didn't change any surrounding logic, but your logic sounds good.

lib/sycamore/sycamore/utils/time_trace.py
alexaryn: I'm not convinced that whole-operator is the only thing we'll want.  It's a thing we should make easy, though.  Let me know what "metadata" means and how I can integrate with it.

lib/sycamore/sycamore/scans/file_scan.py
```diff
@@ -122,6 +123,8 @@ def __init__(
         self._filter_paths_by_extension = filter_paths_by_extension
 
     def _to_document(self, dict: dict[str, Any]) -> dict[str, bytes]:
+        tt = TimeTrace("readBinary")
```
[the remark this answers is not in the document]
alexaryn: There were two reasons I didn't use context-manager style everywhere.  First, I wanted to demonstrate both ways to use TimeTrace.  Second, I didn't want to generate lots of indentation-only diffs.  I'll look into the decorator idea.

lib/sycamore/sycamore/transforms/extract_entity.py
eric-anderson: I'd recommend changing this to entityExtract since I can't find anything in there that forces this to be openai (the variable is even named self._llm.  If you're going to insist on leaving in something about openai, then either a) openAIEntityExtract or a comment to explain.  I'm going to have trouble remembering this even with the explanation.

eric-anderson (approved):

lib/sycamore/sycamore/transforms/extract_entity.py
alexaryn: The class is called `OpenAIEntityExtractor`.  I think the label is fine; they should be short.

PR #390, opened 2024-05-13
Fix integration tests.

eric-anderson:
* Document class started using the elements parameter, test_embed was setting them in an incompatible way
* Add additional checking so that the error message is clearer.


HenryL27 (approved):
let's see if it works or if there are a bunch of other problems that were hiding behind this one

PR #395, opened 2024-05-15
Fix model deserialization error

bohou-aryn:
Looks like the driver/worker process model ser/de incompatible, move model instantiation inside worker process to get rid of the ser/de.
```
RuntimeError: Attempting to deserialize object on a CUDA device but
torch.cuda.is_available() is False. If you are running on a CPU-only
machine, please use torch.load with map_location=torch.device('cpu')
to map your storages to the CPU.
```


alexaryn (approved):
Works for me!

PR #396, opened 2024-05-15
Tools to look at TimeTrace output.

alexaryn:


apps/timetrace/ttcat
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/python3
```
baitsguy, line 1: no .py extension?

baitsguy (approved):

baitsguy (approved):

Makefile
```diff
@@ -0,0 +1,12 @@
+DIRS := apps
```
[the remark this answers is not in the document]
alexaryn: Makefiles are a lot simpler if they don't try to build things in other directories themselves.

apps/timetrace/ttcat
alexaryn: Since it's a script one might want in their path, I think people would rather type `ttcat` than `ttcat.py`.  And who knows?  We may soon convert it to some language other than Python.

PR #398, opened 2024-05-16
ensure unique uuids post explode via sequence numbers

HenryL27:
I managed to get a bunch of docs with non-unique uuids post-explode, e.g.
```
fd9b2074-130f-11ef-a237-3a049e7f7082
fd9b2164-130f-11ef-a237-3a049e7f7082
fd9b2196-130f-11ef-a237-3a049e7f7082
...
```
Not sure how that happened, but this aims to make sure that the uuid1 function always has a different input. If we overflow the counter then we've pigeonholed ourselves anyway


HenryL27:
oh wait those are different, nvm

bohou-aryn (approved):

PR #399, opened 2024-05-16
Convert map.py transforms over to base map.

eric-anderson:
* Remove some unused generate_* functions
* Adjust transforms to import the generate function from the original source rather than map since map no longer uses them. Ideally we'll remove all ray specific bits from the transforms when converting them over.

### DIFF
--- a/lib/sycamore/sycamore/transforms/base.py
+++ b/lib/sycamore/sycamore/transforms/base.py
@@ -37,14 +37,15 @@ class BaseMapTransform(UnaryNode):
         kwargs: Optional[dict[str, Any]] = None,
         constructor_args: Optional[Iterable[Any]] = None,
         constructor_kwargs: Optional[dict[str, Any]] = None,
+        # If we auto-generate lineage, then the conversion to BaseMap has to go in a single PR
+        # since everything needs to be updated to skip metadata. If we temporarily disable the
+        # lineage metadata, then we can do the conversion to BaseMap in separate PRs.
+        enable_auto_metadata=False,
         **resource_args,
     ):
         super().__init__(child, **resource_args)
         if name is None:
-            if isinstance(f, type):
-                name = type(f).__name__
-            else:
-                name = f.__name__
+            name = f.__name__
 
         self._f = f
         self._name = name
@@ -52,6 +53,7 @@ class BaseMapTransform(UnaryNode):
         self._kwargs = kwargs
         self._constructor_args = constructor_args
         self._constructor_kwargs = constructor_kwargs
+        self._enable_auto_metadata = enable_auto_metadata
 
     def execute(self) -> "Dataset":
         input_dataset = self.child().execute()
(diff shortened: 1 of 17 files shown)


lib/sycamore/sycamore/transforms/map.py
```diff
@@ -62,20 +63,39 @@ def custom_flat_mapping_function(document: Document) -> list[Document]:
...
+            class _Wrap(f):  # type: ignore[valid-type,misc]
+                def __init__(self, *args, **kwargs):
+                    super().__init__(*args, **kwargs)
+
+                def __call__(self, docs):
+                    s = super()
+                    ret = []
+                    for d in docs:
+                        ret.extend(s.__call__(d))
+                    return ret
```
HenryL27, line 81: potential one-liner alternatives:
```python
return sum((s.__call__(d) for d in docs), start=[])

return list(itertools.chain.from_iterable(s.__call__(d) for d in docs)))
```

lib/sycamore/sycamore/transforms/map.py
```diff
@@ -62,20 +63,39 @@ def custom_flat_mapping_function(document: Document) -> list[Document]:
...
+            return _Wrap
         else:
-            ray_callable = generate_flat_map_function(self._f)
-            return input_dataset.flat_map(ray_callable, **self.resource_args)
 
+            def _wrap(docs):
+                assert isinstance(docs, list)
+                ret = []
+                for d in docs:
+                    assert isinstance(d, Document)
```
HenryL27, line 90: should these asserts also be in the class part?

lib/sycamore/sycamore/transforms/map.py
HenryL27: nvm, seems like the manual loop is actually faster

lib/sycamore/sycamore/transforms/map.py
```diff
@@ -102,20 +122,12 @@ def __init__(
...
-    def execute(self) -> "Dataset":
-        input_dataset = self.child().execute()
-        if isinstance(self._f, type):
-            ray_callable = generate_map_batch_class(
-                self._f, self._f_args, self._f_kwargs, self._f_constructor_args, self._f_constructor_kwargs
-            )
-            return input_dataset.map_batches(ray_callable, compute=ActorPoolStrategy(size=1), **self.resource_args)
-        else:
-            ray_callable = generate_map_batch_function(self._f)
-            return input_dataset.map_batches(ray_callable, **self.resource_args)
```
HenryL27, line 121: where did this go?

HenryL27 (commented):
mostly nits, main question is where did the ray dataset calls end up? Maybe they got moved in the prev PR?

lib/sycamore/sycamore/transforms/map.py
eric-anderson: All the interaction with ray happens in BaseMapTransform.

HenryL27 (approved):

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -37,21 +37,23 @@ def __init__(
         kwargs: Optional[dict[str, Any]] = None,
         constructor_args: Optional[Iterable[Any]] = None,
         constructor_kwargs: Optional[dict[str, Any]] = None,
+        # If we auto-generate lineage, then the conversion to BaseMap has to go in a single PR
+        # since everything needs to be updated to skip metadata. If we temporarily disable the
+        # lineage metadata, then we can do the conversion to BaseMap in separate PRs.
+        enable_auto_metadata=False,
```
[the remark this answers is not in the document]
eric-anderson: Done.

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -61,15 +63,37 @@ def execute(self) -> "Dataset":
...
 
+    def _local_process(self, in_docs: list[Document]) -> list[Document]:
+        """Internal function for faster testing during the conversion to running on BaseMap.
+        If extended with metadata support, this could become more real."""
+        import copy
+
+        # transforms assume they can mutate docs in place; this works in ray because documents are serialized and
+        # deserialized between every stage.
+        docs = copy.deepcopy(in_docs)
+        if isinstance(self._f, type):
```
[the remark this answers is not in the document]
eric-anderson: Added comment.

lib/sycamore/sycamore/transforms/map.py
eric-anderson: Ack.

lib/sycamore/sycamore/transforms/map.py
eric-anderson: They were for helping me debug problems, but don't hurt to put in both branches, so I did.

lib/sycamore/sycamore/transforms/map.py
eric-anderson: The same as the code here?
https://github.com/aryn-ai/sycamore/blob/7c7d3ce4aa48bd1543c70af20a9056f87c52e252/lib/sycamore/sycamore/transforms/base.py#L59

HenryL27: yeah, that! where's that thing you linked at me right there ;)
I guess I was blinded by the flashy _process_ray method

PR #400, opened 2024-05-16
Add weaviate writer

HenryL27:
Add a writer for the weaviate vector db. Try with:
```bash
docker compose -f apps/weaviate/compose.yml up -d
poetry run python examples/simple_weaviate.py lib/sycamore/sycamore/tests/resources/data/pdfs/Transformer.py
poetry run python -c 'import weaviate; client=weaviate.connect_to_local(); client.collections.get("DemoCollection").query.near_text(query="What is the transformer architecture", limit=3, target_vector="embedding", return_metadata=["text_representation"])'
```
This integration doesn't throw errors very well... When documents fail to write, which will happen occasionally because weaviate auto-schematizes empty lists as empty lists of strings (as far as I can tell), causing it to reject new documents that have real non-string values in the lists, the failures are silent. I think it's something to do with the interaction of the weaviate client batch import threading model and ray. Equally likely is that I've just done a stupid.

I'd like to make the schematizing smarter so I don't have to rely on the auto stuff and I also don't have to rely on the user fully speccing out all the internal properties that cause problems - my debugging consisted of adding a whole bunch of print statements to the internals of the weaviate client to see what properties were failing and what errors I was getting. That'll probably mean more strongly typed properties.

Also sycamore doesn't at the moment support more than one embedding per record; we might want to add this.


lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -0,0 +1,120 @@
...
+        block = builder.build()
+        client = WeaviateClient(**self._client_params)
+        with client:
+            collection = client.collections.get(self._collection_name)
+            with collection.batch.dynamic() as batch:
+                objects = self._extract_weaviate_objects(block)
+                refs = []
+                for obj in objects:
+                    if "references" in obj:
+                        obj_refs = obj.pop("references")
```
alexaryn, line 66: This does two hash lookups in quick succession.  Better would be like:
```
obj_refs = obj.pop("references", None)
if obj_refs is not None:

File: lib/sycamore/sycamore/writers/weaviate_writer.py [hunk 2]
```diff
@@ -0,0 +1,120 @@
...
+                        for k, v in obj_refs.items():
+                            refs.append(DataReference(from_uuid=obj["uuid"], from_property=k, to_uuid=v))
```
alexaryn, line 68: Are these our sketchy UUIDs, or are they some WV8 kind of UUIDs?

lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -0,0 +1,120 @@
...
+                    batch.add_object(**obj)
+                # Flush the objects first so that references can know about them
+                batch.flush()
+                for ref in refs:
+                    batch.add_reference(from_uuid=ref.from_uuid, from_property=ref.from_property, to=ref.to_uuid)
+
+    @staticmethod
+    def _extract_weaviate_objects(block):
+        # Weaviate gets grumpy about NoneTypes so we remove them
+        # either by defaulting to null-ish values or dropping
```
alexaryn, line 78: I don't see defaulting, but I do see dropping.

lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -0,0 +1,120 @@
...
+                "text_representation": "",
+                "elements": [],
+                "embedding": None,
+                "parent_id": None,
+                "properties": {},
+                "bbox": [],
+                "shingles": [],
+            }
+            doc = Document.from_row(record)
+            uuid = doc.doc_id
```
alexaryn, line 100: I still dream of a world in which the ID is a URL.

lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -0,0 +1,120 @@
...
+                "properties": {
+                    "properties": _trim_nones_in_props(data.get("properties", default["properties"])),
+                    "type": data.get("type", default["type"]),
+                    "text_representation": data.get("text_representation", default["text_representation"]),
+                    "bbox": data.get("bbox", default["bbox"]),
+                    "shingles": data.get("shingles", default["shingles"]),
+                },
+            }
+            if "embedding" in data:
+                object["vector"] = {"embedding": data.get("embedding")}
```
alexaryn, line 114: Another double-lookup.

alexaryn (approved):
Seems pretty decent.  Some comments, but nothing on fire.

apps/weaviate/compose.yml
```diff
@@ -0,0 +1,32 @@
...
+version: "3.4"
+services:
+  weaviate:
+    command:
+      - --host
+      - 0.0.0.0
+      - --port
+      - "8080"
+      - --scheme
+      - http
```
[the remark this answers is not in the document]
HenryL27: maybe, but then I have to deal with certs and this is just a demo. if we swap out opensearch we'll want to think about this

examples/simple_weaviate.py
```diff
@@ -0,0 +1,83 @@
...
+    .partition(partitioner=SycamorePartitioner(extract_table_structure=True, extract_images=True))
+    .regex_replace(COALESCE_WHITESPACE)
+    .extract_entity(entity_extractor=OpenAIEntityExtractor("title", llm=davinci_llm, prompt_template=title_template))
+    .mark_bbox_preset(tokenizer=tokenizer)
+    .merge(merger=MarkedMerger())
+    .spread_properties(["path", "title"])
+    .split_elements(tokenizer=tokenizer, max_tokens=512)
+)
+
+ds = ds.explode().embed(embedder=SentenceTransformerEmbedder(model_name=model_name, batch_size=100)).sketch(window=17)
```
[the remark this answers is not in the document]
HenryL27: so it's easy to stick a `.show()` in the middle

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -214,7 +214,7 @@ def infer(self, image: Image.Image, threshold: float) -> list[Element]:
             element = create_element(
                 type=self.labels[label],
                 bbox=BoundingBox(box[0] / w, box[1] / h, box[2] / w, box[3] / h).coordinates,
-                properties={"score": score},
+                properties={"score": score.item()},
```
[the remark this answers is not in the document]
HenryL27: should be eminently resolvable

lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -0,0 +1,120 @@
...
+
+class WeaviateDatasink(Datasink):
+    def __init__(self, collection_name: str, client_params: dict, collection_config: Optional[dict]):
+        self._collection_name = collection_name
+        self._client_params = client_params
+        self._collection_config = collection_config
+
+    def on_write_start(self):
+        client = WeaviateClient(**self._client_params)
+        with client:
```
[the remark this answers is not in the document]
HenryL27: yep

lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -0,0 +1,120 @@
...
+                        "already exists, so ignoring provided config."
+                    )
+                else:
+                    client.collections.create(**self._collection_config)
+
+    def write(self, blocks: Iterable[Block], ctx: TaskContext) -> Any:
+        builder = DelegatingBlockBuilder()
+        for block in blocks:
+            builder.add_block(block)
+        block = builder.build()
```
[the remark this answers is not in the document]
HenryL27: I'm just following the opensearch writer here. I think the DBB.build is essentially a big concatenation op, like a StringBuilder

lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -0,0 +1,120 @@
...
+        client = WeaviateClient(**self._client_params)
+        with client:
+            collection = client.collections.get(self._collection_name)
```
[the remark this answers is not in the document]
HenryL27: this is a method in the weaviate client, not a python dictionary. client.collections is a namespace, not an object - client.collections.get() is what performs the database call. there's also, for example, client.collections.list_all()

lib/sycamore/sycamore/writers/weaviate_writer.py
HenryL27: ack

lib/sycamore/sycamore/writers/weaviate_writer.py
HenryL27: these are our sketchy uuids

lib/sycamore/sycamore/writers/weaviate_writer.py
HenryL27: defaulting: type = data.get("type", default["type"])

lib/sycamore/sycamore/writers/weaviate_writer.py
HenryL27: ack

lib/sycamore/sycamore/writer.py
```diff
@@ -78,6 +78,98 @@ def opensearch(
...
+                    .regex_replace(COALESCE_WHITESPACE)
+                    .extract_entity(entity_extractor=OpenAIEntityExtractor(
+                            "title", llm=davinci_llm, prompt_template=title_template))
+                    .mark_bbox_preset(tokenizer=tokenizer)
+                    .merge(merger=MarkedMerger())
+                    .spread_properties(["path", "title"])
+                    .split_elements(tokenizer=tokenizer, max_tokens=512)
+                    .explode()
+                    .embed(embedder=SentenceTransformerEmbedder(model_name=model_name, batch_size=100))
+                    .sketch(window=17)
```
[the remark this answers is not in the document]
HenryL27: given that the .sketch has implications for defining the collection schema ima leave it fleshed out

PR #405, opened 2024-05-18
Switch Filter and Enbed over to BaseMapTransform

eric-anderson:
* Add support for new "object called as class or function" mode.
* Fix bug where specifying num_gpus = 0 in resource args causes problems because ray thinks that means it needs a GPU. Turns out existing code wasn't passing in resource_args.


HenryL27 (approved):
lgtm

PR #407, opened 2024-05-20
Add timing for OpenSearch writer.

alexaryn:


HenryL27 (approved):
prob a dumb q but does this essentially time the entire pipeline execution since lazy + write is an end node?

alexaryn:
> prob a dumb q but does this essentially time the entire pipeline execution since lazy + write is an end node?

It does seem to time everything.  I guess I should try moving it around and see if I can get more specific timing.  If not, it still records CPU usage, which might be useful.

PR #408, opened 2024-05-20
Convert classes over to new *Map classes 3/n

eric-anderson:
* Remove now unused functions in generate_ray_func.py


HenryL27 (approved):
Beautiful, no notes

PR #410, opened 2024-05-21
Switch from uuid1() to uuid4() in explode.

bsowell:
uuid1().is_safe returns unknown on Mac, and it's posible that concurrent calls may produce collisions on the same host. This change just switches to uuid4 (random uuids) to reduce this risk.

### DIFF
--- a/lib/sycamore/sycamore/transforms/explode.py
+++ b/lib/sycamore/sycamore/transforms/explode.py
@@ -32,11 +32,9 @@ class Explode(SingleThreadUser, NonGPUUser, FlatMap):
 
         import uuid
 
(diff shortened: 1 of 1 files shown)


HenryL27 (approved):
`python -c 'import uuid; assert uuid.uuid4().is_safe == uuid.SafeUUID.safe'` also fails on my mac, but ok

eric-anderson (approved):
uuid1 safety is determined by whether getting the clock is safe.
https://github.com/python/cpython/blob/e870c852c0ea96fa4e4569e9c39c7ceb80ce858d/Lib/uuid.py#L670
The rest of the uuid's return Unknown.

PR #414, opened 2024-05-21
Update poetry lock files

eric-anderson:
* Upgrade to fix some  reported vulnerabilities on github
  * Make versions consistent between potery lock files again


HenryL27:
aww man this makes your added line count positive this week

bsowell (approved):
As far as I can tell this all looks fine. I confess I find the poetry-lock package a bit confusing. I may need you to explain it to me at some point, but that is not important to this PR.

PR #415, opened 2024-05-21
Fix OOM on CPU by reducing default batch size

MarkLindblad:


lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -394,7 +394,7 @@ def __init__(
         extract_images=False,
         device=None,
         model_server_endpoint=None,
-        batch_size: int = 10,
+        batch_size: int = 1,
```
alexaryn, line 397: There was also a default of 10 in detr_partitioner.py.  Do we want to change that, too?

alexaryn (approved):
Thanks.

lib/sycamore/sycamore/transforms/partition.py
MarkLindblad: Yes.

PR #416, opened 2024-05-22
Conversion to base map 5/n: Partition

eric-anderson:
* Add new Composite class that models multiple composite operations in a single node
* Use it to rewrite partition which is a sometime composite operation.

### DIFF
--- a/lib/sycamore/sycamore/transforms/base.py
+++ b/lib/sycamore/sycamore/transforms/base.py
@@ -193,3 +193,29 @@ class BaseMapTransform(UnaryNode):
         to_ids = [d.lineage_id for d in to_docs]
 
         return [MetadataDocument(lineage_links={"from_ids": from_ids, "to_ids": to_ids})]
+
+
+class Composite(UnaryNode):
+    def __init__(self, child: Node, base_args: list[dict], **resource_args):
+        super().__init__(child, **resource_args)
+        self.nodes = Composite.combine(child, base_args, **resource_args)
+
+    @staticmethod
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -501,23 +502,20 @@ class Partition(Transform):
...
-                **self.resource_args,
-            )
-        else:
-            dataset = input_dataset.map(generate_map_function(self._partitioner.partition))
-        if self._table_extractor:
-            dataset = dataset.map(generate_map_function(self._table_extractor.extract_tables))
-        return dataset
+        ops = []
+
+        if isinstance(partitioner, SycamorePartitioner):
```
HenryL27, line 507: I hope we can turn this check into a property or method or something but out of scope of this

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -193,3 +193,29 @@ def _update_lineage(cls, from_docs, to_docs):
         to_ids = [d.lineage_id for d in to_docs]
 
         return [MetadataDocument(lineage_links={"from_ids": from_ids, "to_ids": to_ids})]
+
+
+class Composite(UnaryNode):
```
HenryL27, line 198: maybe `CompositeTransform` or `CompositeNode`? idk, just feels a little funny next to `BaseMapTransform` but fine if you wanna keep it

HenryL27 (approved):
1 minor nit you don't have to pay attention to

lib/sycamore/sycamore/transforms/base.py
eric-anderson: done.

PR #419, opened 2024-05-22
TimeTrace: add RSS and improve usability

alexaryn:


HenryL27 (approved):
please let this not be some malicious thing I'm letting through

PR #421, opened 2024-05-23
remove empty lists from documents in weaviate writer

HenryL27:
Weaviate autoschema will parse empty lists as lists of strings and then refuse to ingest subsequent documents that prove this assumption wrong. But empty lists don't have any meaningful information in them, so we can simply drop them entirely from the record, so weaviate will not have problems. This means we have a much easier time specifying out the problematic fields ahead of time because there aren't any.

### DIFF
--- a/examples/simple_weaviate.py
+++ b/examples/simple_weaviate.py
@@ -1,8 +1,8 @@
 import sys
 
-from weaviate.classes.config import Property, ReferenceProperty
+from weaviate.classes.config import ReferenceProperty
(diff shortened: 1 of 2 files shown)


bsowell:
@HenryL27 Looks like there is an issue with the unit tests and mypy. possible this: `lib/sycamore/sycamore/writers/weaviate_writer.py:2: error: Module "types" has no attribute "NoneType"  [attr-defined]`.

bsowell (changes requested):

bsowell (approved):

PR #426, opened 2024-05-23
TimeTrace: Fallback to RUSAGE_SELF when RUSAGE_THREAD isn't present.

bsowell:
This is true on the Mac, for instance. Also, switch the ttviz Makefile to c++20 to be more accomodating to older compilers and add a README.


alexaryn (approved):
LGTM

PR #427, opened 2024-05-24
Switch extract_schema to BaseMap -- *Map 8/8

eric-anderson:
* Rewrite extract_schema so that the extraction runs under ray and via an actor rather than on the source machine.
* Added checking that the various functions and parameters are serializable. The error when it isn't is very confusing.  In particular mock objects are not serializable, so the check is skipped if the parent node is None since that's only valid in a local execution.
* Discovered that typechecking wasn't applying, and as a result we never typechecked the Class variation of the various functions. Turning on typechecking showed that it didn't work, so I swapped it to Any since we haven't had bugs there.
* Added -> None to some of the tests to force typechecking
* Remove the last generate_map_* function.


lib/sycamore/sycamore/llms/llms.py
```diff
@@ -13,3 +13,17 @@ def generate(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) ->
     @abstractmethod
     def is_chat_mode(self):
         pass
+
+
+class FakeLLM(LLM):
+    """Useful for tests where te fake LLM needs to run in a ray function because mocks are not serializable"""
```
HenryL27, line 19: te -> the

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -49,6 +49,21 @@ def __init__(
...
+            # give a much more sensible error message and give it before ray starts execution.
+            from ray.util import inspect_serializability
+            import io
+
+            log = io.StringIO()
+            (ok, s) = inspect_serializability(
+                [f, name, args, kwargs, constructor_args, constructor_kwargs], print_file=log
+            )
+            if not ok:
+                raise ValueError(f"Something isn't serializable {s}\nLog: {log.getvalue()}")
```
HenryL27, line 65: Any chance we can also stick the name of the failing node/map fn in this error message?

HenryL27 (approved):

PR #430, opened 2024-05-28
Prettier output for timetrace files when using ttviz & ttanal

alexaryn:


apps/timetrace/ttanal
```diff
@@ -32,6 +33,46 @@ def tt_reader(paths: list[str]):
...
+    ("y", 31556952),
+    (None, None),
+]
+
+mem_units = [
+    ("B", 1),
+    ("kB", 1024),
+    ("MB", 1048576),
+    ("GB", 1073741824),
+    ("TB", 1099511627776),
```
HenryL27, line 53: any chance of notating these numbers as `2**20, 2**30, 2**40` or at least sticking in underscores?

apps/timetrace/ttanal
```diff
@@ -32,6 +33,46 @@ def tt_reader(paths: list[str]):
...
+            s /= val
+            return f"{sign}{s:.3g}{unit}"
+
+def pretty_time(s: float) -> str:
+    return pretty_num(time_units, s)
+
+def pretty_mem(s: float) -> str:
+    return pretty_num(mem_units, s)
+
 class Foo:
```
HenryL27, line 76: I must be missing something here but why is this a class in the first place?

HenryL27 (approved):

apps/timetrace/ttanal
alexaryn: It's a habit, because globals are bad and encapsulation is good.  I forgot to take advantage.  Update coming momentarily.

PR #431, opened 2024-05-28
Pin setuptools so that we can run DETR model on GPU on Linux.

alexaryn:


HenryL27 (approved):
I feel like this should touch the lockfiles, no?

HenryL27:
You may need to run the poetry-lock-all script here

PR #434, opened 2024-05-29
Enable table extraction to use GPU.

alexaryn:


lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -81,8 +82,17 @@ def __init__(self, model: str = DEFAULT_TTAR_MODEL):
         """
 
         self.model = model
+        self.device = device
         self.structure_model = None
 
+    def _get_device(self) -> str:
```
bsowell, line 88: I might see if I can add `mps` as an option here, but doesn't make sense for you to add it since you would need a mac to test.

bsowell (approved):
Thanks for doing this!

PR #435, opened 2024-05-29
Pinecone connector

HenryL27:
Adds support for writing to [Pinecone](https://pinecone.io).
Use the `term_frequency` transform to enable basic keyword/hybrid search.
Parent Documents are not ingested, rather leaf documents have ids `parent_id#leaf_id`

Note: I somehow ended up in dependency hell... don't worry about the lockfiles (or do)


lib/sycamore/sycamore/llms/openai.py
```diff
@@ -151,18 +151,23 @@ def get_guidance_model(self, model) -> Model:
             # specific assumptions about how deployed models are named that don't work
             # well with Azure. This is the only way I was able to get it to work
             # reliably.
+            # p.s. mypy seems to get mad if cls is not defined as being either, hence
+            # the union expression
             if model.is_chat:
-                cls = AzureOpenAIChat
+                cls: Union[type[AzureOpenAIChat], type[AzureOpenAICompletion]] = AzureOpenAIChat
```
bsowell, line 157: Did this fail for you in Github or locally? Did you run mypy it from the root (sycamore-monorepo) or from the lib directory? I run mypy all the time (mostly in the library directory), and it's frustrating that these errors didn't surface for me or in any of the Github commits after this change.

bsowell: I guess one other thought -- did mypy get upgraded in your poetry lock?

lib/sycamore/sycamore/writers/pinecone_writer.py
```diff
@@ -0,0 +1,175 @@
...
+                elif v is not None:
+                    items.append(((str(k) if len(prefix) == 0 else f"{prefix}.{k}"), v))
+            return items
+
+        def _metadata_special_cases(data: dict) -> dict:
+            binary = data.get("binary_representation", None)
+            if binary:
+                b64binary = b64encode(binary)
+                strbinary = b64binary.decode("UTF-8")
+                data["binary_representation"] = strbinary
```
bsowell, line 131: Is this the right default? It seems like the user is very rarely going to actually want to write the binary_representation out to the vector store. Maybe we default to skipping it and have a property to enable?

lib/sycamore/sycamore/writers/pinecone_writer.py
```diff
@@ -0,0 +1,175 @@
...
+                    if isinstance(v, (list, tuple)) and all(isinstance(innerv, str) for innerv in v):
+                        items.append(((str(k) if len(prefix) == 0 else f"{prefix}.{k}"), v))
+                    else:
+                        inner_values = _flatten_metadata(v, prefix=(str(k) if len(prefix) == 0 else f"{prefix}.{k}"))
+                        items.extend([(innerk, innerv) for innerk, innerv in inner_values])
+                elif v is not None:
+                    items.append(((str(k) if len(prefix) == 0 else f"{prefix}.{k}"), v))
+            return items
+
+        def _metadata_special_cases(data: dict) -> dict:
```
bsowell, line 126: Is there a way to generalize this, so we don't have to list every field? Maybe handle things by type? Are there ever cases where we are going to want to handle two properties/fields of the same type differently? 

I also wonder if the Table object in table elements will cause trouble: https://github.com/aryn-ai/sycamore/blob/3914113955d87308f6f4c04afa9a59618357d401/lib/sycamore/sycamore/data/element.py#L121 ?

lib/sycamore/sycamore/writers/pinecone_writer.py
```diff
@@ -0,0 +1,175 @@
...
+            iterator = []  # type: ignore
+            if isinstance(data, dict):
+                iterator = data.items()  # type: ignore
+            if isinstance(data, (list, tuple)):
+                iterator = enumerate(data)  # type: ignore
+            items = []
+            for k, v in iterator:
+                if isinstance(v, (dict, list, tuple)):
+                    if isinstance(v, (list, tuple)) and all(isinstance(innerv, str) for innerv in v):
+                        items.append(((str(k) if len(prefix) == 0 else f"{prefix}.{k}"), v))
```
bsowell, line 118: Minor, but might be a bit more readable if you defined a helper for `(str(k) if len(prefix) == 0 else f"{prefix}.{k}")` since it shows up a few times.

bsowell (commented):
Initial comments. Let's talk about some of the type stuff, but otherwise nothing too major.

lib/poetry-lock/poetry-lock-all.sh
```diff
@@ -11,7 +11,7 @@ poetry config virtualenvs.in-project true --local
 poetry install
 [[ -d .venv ]] || fail "Did not get .venv directory after install"
 
-tomls=$(find . -name pyproject.toml)
+tomls=$(find . ! -path "**/site-packages/**/*" -name pyproject.toml)
```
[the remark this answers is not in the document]
HenryL27: lock-all wants to create the venvs inside the sycamore repo, and some packages have a pyproject.toml in them, so this excludes them from trying to be poetry-installed by lock-all (find supports boolean logic, ! is --not and putting two things next to each other is --and)

lib/sycamore/sycamore/writers/pinecone_writer.py
HenryL27: Are people gonna want to split document components across several dbs? e.g. vectors to pinecone, metadata to postgres, binary to s3?
Not worried about it now just curious if this is a common etl thing.

lib/sycamore/sycamore/llms/openai.py
HenryL27: (it passed local pre-commit mypy. but `poetry run mypy lib/sycamore/sycamore/llms/openai.py` failed)

lib/sycamore/sycamore/writers/pinecone_writer.py
```diff
@@ -0,0 +1,207 @@
...
+
+        objects = self._extract_pinecone_objects(block)
+        obj_it = iter(objects)
+        batch = list(itertools.islice(obj_it, self._batch_size))
+        async_results = []
+        while len(batch) > 0:
+            async_results.append(index.upsert(vectors=batch, namespace=self._namespace, async_req=True))
+            batch = list(itertools.islice(obj_it, self._batch_size))
+        for r in async_results:
+            r.result()
```
eric-anderson, line 114: Ideally check something in the result (I can't work out what we can usefully check); or at least `# force async completion, errors will be thrown at that point`

lib/sycamore/sycamore/writers/pinecone_writer.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def _extract_pinecone_objects(self, block: Block) -> Iterable[Vector]:
+        def _add_key_to_prefix(key, prefix=""):
+            if len(prefix) == 0:
+                return str(key)
+            else:
+                return f"{prefix}.{key}"
+
+        def _flatten_metadata(data: Union[dict, list, tuple], prefix="") -> Iterable[Tuple[Any, Any]]:
+            # Pinecone requires metadata to be flat (no nested objects)
```
eric-anderson, line 124: `except for lists of strings`?

lib/sycamore/sycamore/writers/pinecone_writer.py
```diff
@@ -0,0 +1,207 @@
...
+        batch = list(itertools.islice(obj_it, self._batch_size))
+        async_results = []
+        while len(batch) > 0:
+            async_results.append(index.upsert(vectors=batch, namespace=self._namespace, async_req=True))
+            batch = list(itertools.islice(obj_it, self._batch_size))
+        for r in async_results:
+            r.result()
+
+    def _extract_pinecone_objects(self, block: Block) -> Iterable[Vector]:
+        def _add_key_to_prefix(key, prefix=""):
```
eric-anderson, line 117: Suggest (prefix, key) since that's the order they end up in and you always specify both parameters.

lib/sycamore/sycamore/writers/pinecone_writer.py
```diff
@@ -0,0 +1,207 @@
...
+        def _flatten_metadata(data: Union[dict, list, tuple], prefix="") -> Iterable[Tuple[Any, Any]]:
+            # Pinecone requires metadata to be flat (no nested objects)
+            # so here's a traversal
+            iterator = []  # type: ignore
+            if isinstance(data, dict):
+                iterator = data.items()  # type: ignore
+            if isinstance(data, (list, tuple)):
+                iterator = enumerate(data)  # type: ignore
+            items = []
+            for k, v in iterator:
```
eric-anderson, line 132: Can you do an integration test where they keys are things like "\0" "a b c", "a\nb\r\n"? The pinecone docs say that keys are strings, but all the examples are /^[a-z_]+$/; we're going to have all sorts of weirdness as a result of flattening.

lib/sycamore/sycamore/writers/pinecone_writer.py
eric-anderson: I'd suggest adding a .to_simple_type() method to the metadata special cases; then the logic for conversion goes becomes part of the type rather than part of the writer.  Unclear what to do in cases where it isn't a simple type, probably a warning and skip the field is the best you can do. 

The other two writers are currently just taking whatever is in properties and passing it through. I have no idea what that will do with complex types.

lib/sycamore/sycamore/writers/pinecone_writer.py
eric-anderson: It looks to me like the other two writers are going to drop the binary representation, so I'd do the same here.

lib/sycamore/sycamore/writers/pinecone_writer.py
```diff
@@ -0,0 +1,207 @@
...
+            # store it as csv.
+            table = data.pop("table", None)
+            if table:
+                data["table"] = table.to_csv()
+            return data
+
+        def _record_to_object(record):
+            doc = Document.from_row(record)
+            data = doc.data
+            # Pull out pinecone top-level properties and leave
```
eric-anderson, line 180: The opensearch and weviate writers pull specific properties out of data and then pass through all of doc.properties.  I feel like we should do the same thing here (or plan to change the other two writers).

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -151,18 +151,23 @@ def get_guidance_model(self, model) -> Model:
...
             else:
                 cls = AzureOpenAICompletion
 
+            # These random str's are also to appease mypy, since those params might be
+            # None in the OPENAI case, where this code is unreachable but mypy doesn't
+            # know this.
             return cls(
                 model=model.name,
-                azure_endpoint=self.azure_endpoint,
+                azure_endpoint=str(self.azure_endpoint),
```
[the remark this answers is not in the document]
HenryL27: we can't really fix the type sig of AzureOpenAI*; they come from guidance (and were also removed in the next version of guidance btw). 
the vars are checked in __init__, so they are mostly not None if we're in this codepath.
Will work on this... maybe I can trick it with **kwargs shenanigans.
for reference  the AzureOpenAI __init__ sig from guidance is
```python
def __init__(
        self,
        model: str,
        azure_endpoint: str,
        azure_deployment: str = None,
        azure_ad_token_provider=None,
        api_key: str = None,
        tokenizer=None,
        echo=True,
        version: str = None,
        max_streaming_tokens=1000,
        timeout=0.5,
        compute_log_probs=False,
        **kwargs,
    ):
```
which is where all these problems are coming from

lib/sycamore/sycamore/writers/pinecone_writer.py
HenryL27: basically what I'm doing here although I'll clean it up so it's a selection rather than a kitchen sink

lib/sycamore/sycamore/writers/pinecone_writer.py
HenryL27: moving around the logic to better match how the weaviate and opensearch connectors deal with Document -> DB record lets me drop this fn entirely. Also changing it to actively select what Document properties to load rather than just flattening the entire document means I can fully drop the table (as the other writers also do).

Now, we might want to write tables to our other sources as well, but that's a separate issue.

lib/sycamore/sycamore/writers/pinecone_writer.py
HenryL27:
```
ds.with_property("really awkward\nnam\0e", lambda doc: {"another\r\nkey": "dummy"})
```
becomes 
`properties.really awkward name.another key: "dummy"` (in the pinecone query ui)
You could probably commit shenanigans but I'm not too worried

lib/sycamore/sycamore/llms/openai.py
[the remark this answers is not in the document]
HenryL27: fixed with TypedDict variadic shenanigans

eric-anderson (approved):

PR #437, opened 2024-06-01
Demo UI eslint prettier config

sohamkasar19:
- Add eslint prettier config
- Rearrange and restructure existing files
- gpt-4o default model for querying


apps/demo-ui/ui/src/pages/PDFViewer/PdfViewer.tsx
```diff
@@ -0,0 +1,336 @@
+import React, { useRef } from "react";
```
baitsguy, line 1: should this be called index.tsx like the others?

apps/demo-ui/ui/.prettierrc.json
```diff
@@ -0,0 +1,3 @@
+{
```
baitsguy, line 1: how does the styling take effect? do we need to setup the idea to use it or is it all automatic?

apps/demo-ui/ui/src/pages/PDFViewer/PdfViewer.tsx
sohamkasar19: Yes, i missed renaming that file. Will change it

PR #438, opened 2024-06-03
Add utility code for working with tables and showing them as HTML.

bsowell:
The to_html_tables method can be used with a flat_map to extract the tables out of a collection of documents and render them as html. This commit also provides importable implementations of `html_table_filename_fn` and `image_page_filename_fn` to simplify demos and testing.


lib/sycamore/sycamore/utils/html_utils.py
```diff
@@ -0,0 +1,33 @@
...
+        new_docs.append(new_doc)
+
+    return new_docs
+
+
+def html_table_filename_fn(doc: Document) -> str:
+    path = Path(doc.properties["path"])
+    base_name = ".".join(path.name.split(".")[0:-1])
+    if "table_num" in doc.properties:
+        suffix = doc.properties["table_num"]
```
baitsguy, line 30: should we include a page_num here?

baitsguy (approved):

lib/sycamore/sycamore/utils/html_utils.py
bsowell: Sure, I can. Right now the table_num is for the whole document, rather than just the page. I can change to name_page_1_table_2, where the table num resets for each page.

bsowell: Based on conversation we will defer this for now, as the page_number property is potentially a bit fragile.

PR #439, opened 2024-06-03
Fix cross-test error

eric-anderson:
Without the shutdown, running:

poetry run pytest lib/sycamore/sycamore/tests/unit/transforms/test_ske*.py lib/sycamore/sycamore/tests/unit/transforms/test_random_sample.py

yields:

RuntimeError: Maybe you called ray.init twice by accident? This error can be suppressed by passing in 'ignore_reinit_error=True' or by calling 'ray.shutdown()' prior to 'ray.init()'

Unclear why it's only needed here.


bsowell (approved):
Have no explanation for why it's only needed here, but doesn't seem likely to break anything, so LGTM.

Issue #364, opened 2024-04-30
Cannot run unit tests according to CONTRIBUTING.md


hsm207:
**Describe the bug**
Running `poetry run pytest sycamore/tests/unit/` from the project root folder throws:

> ERROR: file or directory not found: sycamore/tests/unit/

**To Reproduce**
1. Go to the project's root directory
2. Run `poetry run pytest sycamore/tests/unit/`
3. See error

**Expected behavior**
Unit tests are run

**Screenshots**
NA

**Desktop (please complete the following information):**
 - OS: Debian GNU/Linux 12 (bookworm)
 - Browser [e.g. chrome, safari]: NA
 - Version [e.g. 22]: NA

**Smartphone (please complete the following information):**
 - Device: NA
 - OS: NA
 - Browser: NA
 - Version: NA

**Additional context**
I am following the instructions in [Developer Setup](https://github.com/aryn-ai/sycamore/blob/8a239f54358be580e2b3be259872ad0769c6e4b3/CONTRIBUTING.md#L27) in CONTRIBUTING.md guide

bsowell:
Ahh, thanks for pointing this out. We did a refactor recently and didn't update the CONTRIBUTING instructions accordingly. The library portion of sycamore has moved to lib/sycamore, so you should be able to run the tests from inside that directory. 

Let me do a rev on the instructions and get a PR out today.

bsowell:
@hsm207 Thanks so much for looking at sycamore! Please feel free to reach out if there is anything else we can do to help!

HenryL27:
Updated the instructions to `poetry run lib/sycamore/sycamore/tests/unit/`, which should be correct

PR #441, opened 2024-06-04
fix pytest commands in contributing guide

HenryL27:
for #364


CONTRIBUTING.md
```diff
@@ -75,7 +75,9 @@ Our tests are implemented using PyTest. You can run unit and integration tests a
 poetry run pytest lib/sycamore/sycamore/tests/unit/
 
 # Integration Tests
-poetry run pytest lib/sycamore/sycamore/tests/integration/
+
+# Warning: as of 2024-05-17 the integration tests are currently broken. We are working on fixing them. 
+# poetry run pytest lib/sycamore/sycamore/tests/integration/
```
bsowell, line 80: What is broken about them at the moment that you had in mind? I know there are some (many) non-deterministic issues that cause failures in github. Is there something deterministic that always fails locally as well?

HenryL27: Not sure what the problems were here - shukri was having issues. Something to do with env vars maybe? We decided to mark them as unlikely to pass. This is not new in this PR tho, all I'm doing here is turning `poetry run pytest sycamore/blah` into `poetry run lib/sycamore/sycamore/blah`

bsowell (approved):

PR #442, opened 2024-06-04
Enable metadata by default.

eric-anderson:
* Leaving flag in so it's easier to disable if something didn't have a suitable test.
* Fix all of the tests that use lower-level APIs, so bypass auto-hiding of metadata in the higher level APIs.
* Add a handful more of the ray.shutdowns() -- not sure why I'm seeing this more.


lib/sycamore/sycamore/writers/file_writer.py
```diff
@@ -217,21 +220,48 @@ def get_filename_for_block(self, block: Block, task_index: int, block_index: int
...
+#         filename_fn: Callable[[Document], str] = default_filename,
+#         doc_to_bytes_fn: Callable[[Document], bytes] = default_doc_to_bytes,
+#     ):
+#         super().__init__(path, filesystem=filesystem, filename_provider=DocToRowFilenameProvider(filename_fn))
+#
+#         self._doc_to_bytes_fn = doc_to_bytes_fn
+#
+#     def write_row_to_file(self, row: dict[str, Any], file: NativeFile):
+#         binary = self._doc_to_bytes_fn(Document.from_row(row))
+#         file.write(binary)
```
HenryL27, line 264: here's the stuff that made me think this isn't compllete

eric-anderson: Removed.

HenryL27 (approved):

PR #447, opened 2024-06-05
Run poetry lock in openai-proxy with poetry-lock-all.sh

eric-anderson:
We haven't fixed the proxy to be consistent, so fix the script to special case that directory. Fixes issue with the poetry.lock file not being updated when it needs to be.


HenryL27 (approved):
lgtm

PR #448, opened 2024-06-05
Build Jupyter image with CUDA stuff included.

alexaryn:
Beware that this increases the size of the image `arynai/sycamore-jupyter` from 7.06GB to 12.7GB.  We should probably look for a more efficient way to do this.  But, in the meantime, this PR is available if people need it.


alexaryn:
Abandoning this approach in favor of #449.

PR #449, opened 2024-06-05
Another approach to CUDA support in Docker, with less bloat.

alexaryn:
The has the Janky `dpkg-divert` stuff to work around weird installation errors.  But, it saves 5.6GB in image size over the method in PR #448.

### DIFF
--- a/apps/jupyter/fixuser.py
+++ b/apps/jupyter/fixuser.py
@@ -93,7 +93,7 @@ def install_cuda():
         aa = a.stdout.split()[0]
         bb = b.stdout.split()[0]
         if aa != bb:
-            print(f"WARNING: diverted version does not match installed version: {divert}")
+            print(f"WARNING: diverted version does not match installed version: {divert}", flush=True)
 
 
 def exec_run_jupyter(uid, gid):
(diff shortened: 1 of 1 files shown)


apps/jupyter/fixuser.py
```diff
@@ -58,6 +60,42 @@ def fix_ids(uid, gid):
...
+    cmds.append(["apt", "-y", "install", "nvidia-cuda-toolkit-gcc", "libnvidia-tesla-cuda1"])
+    for cmd in cmds:
+        subprocess.run(cmd, check=True)
+    for divert in diverts:
+        a = subprocess.run(["sha256sum", divert], capture_output=True, check=True)
+        b = subprocess.run(["sha256sum", f"{divert}.distrib"], capture_output=True, check=True)
+        aa = a.stdout.split()[0]
+        bb = b.stdout.split()[0]
+        if aa != bb:
+            print(f"WARNING: diverted version does not match installed version: {divert}")
```
eric-anderson, line 96: , flush=True

Otherwise we get weird interactions with docker.

eric-anderson (approved):
Lots of discussion on the best approach. Conclusion is:
1. Not intending to have this work outside of a debian cloud instance. Near as I can tell Nvidia driver version compatibility is complicated and makes stuff not work easily on laptops.
2. For the cloud service, restarts are rare, so we don't care that much about how long the container takes to come up; also the download should be fast because it should be a local mirror.
3. The checksum will warn us about potential incompatibility.

PR #450, opened 2024-06-06
Provide better error messages on *Map mis-use

eric-anderson:
If you pass in a not-callable, the error message can be obscure (__name__ not found) Instead check for the property that we want that the callable is properly named.

Discovered by trying to use map + GreedyTextElementMerger.

Also verify that the names we calculate are correct. They were not before.

### DIFF
--- a/lib/sycamore/sycamore/transforms/base.py
+++ b/lib/sycamore/sycamore/transforms/base.py
@@ -24,6 +24,28 @@ def rename(new_function_name: str):
     return decorator
 
 
+def get_name_from_callable(f):
+    # check this condition first. A type will have a __name__ but might not have __call__
+    if isinstance(f, type):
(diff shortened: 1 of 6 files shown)


lib/sycamore/sycamore/transforms/base.py
```diff
@@ -24,6 +24,28 @@ def decorator(f):
...
+        return f.__name__
+    except AttributeError:
+        pass
+
+    if "__call__" in dir(f):
+        return f.__class__.__name__
+    else:
+        raise ValueError("f argument is an object without an __call__ method")
+
+    raise ValueError(f"Unable to extract name from {f}, dir(f): {dir(f)}")
```
HenryL27, line 46: I don't think this line is touchable?

HenryL27 (approved):

PR #451, opened 2024-06-06
Writer abstraction

HenryL27:
Introduce a new BaseDBWriter class. This class contains four inner classes that you have to implement:
- Client: for talking to the database
- Record: an intermediate type to transform Documents into (so the DB can understand them)
- ClientParams: object to store data required to create a client (clients often come with network connections which are less than serializable)
- IndexParams: object to store data required to create an index for the first time

It's a MapBatches

see writers/test_base.py Fake* for an example implementation


lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,160 @@
...
+from ray.data._internal.execution.interfaces import TaskContext
+from ray.data.block import Block, BlockAccessor
+from sycamore.data.document import Document, MetadataDocument
+from sycamore.plan_nodes import Node, Write
+
+
+class BaseDBWriter(Write):
+
+    # Type param for the client
+    class client_t(ABC):
```
eric-anderson, line 18: Based on my understanding of python naming, I'd think this should be ClientT, but I haven't seen people using the T convention in python, so I'd go with either just Client, or AbstractClient.

(same below)

lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,160 @@
...
+        def serialize(self) -> bytes:
+            pass
+
+        @classmethod
+        @abstractmethod
+        def deserialize(cls, byteses: bytes) -> "BaseDBWriter.record_t":
+            pass
+
+    # Type param for the object used to configure a new index if necessary
+    @dataclass
```
eric-anderson, line 50: This works for OpenSearch and Pinecone which both refer to indexes. Weaviate seems to use collection. SQL uses table, S3 uses bucket, filesystems would use directory. Maybe DestinationParams since users will be specifiying a destination for the writes?

lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,160 @@
...
+        self._index_params = index_params
+
+    def get_client(self, client_params: client_params_t) -> client_t:
+        return self.client_t.from_client_params(client_params)
+
+    @classmethod
+    def doc_to_record(cls, doc: Document) -> record_t:
+        return cls.record_t.from_doc(doc)
+
+    def execute(self) -> Dataset:
```
eric-anderson, line 75: Per discussion, this should vanish and be collapsed into a MapBatch.

lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,160 @@
...
+        input_dataset = self.child().execute()
+        record_dataset = self._ray_map_docs_to_records(input_dataset)
+        self.get_client(self._client_params)
+        datasink = BaseDBWriter.InnerDatasink(self._client_params, self._index_params, self.__class__)
+        _check_serializable(datasink)
+        record_dataset.write_datasink(datasink, ray_remote_args=self.resource_args)
+        return input_dataset
+
+    def _ray_map_docs_to_records(self, dataset: Dataset) -> Dataset:
```
eric-anderson, line 84: I think this function vanishes if you build on MapBatch.

lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,160 @@
...
+            index_params: Optional["BaseDBWriter.index_params_t"],
+            owner_cls: type["BaseDBWriter"],
+        ):
+            _check_serializable(client_params, index_params, owner_cls)
+
+            self._client_params = client_params
+            self._index_params = index_params
+            self._owner = owner_cls
+
+        def on_write_start(self) -> None:
```
eric-anderson, line 116: This is one piece that I'm not sure how best to handle. Maybe if we define it as create_index_idempotent it will work, and the logic becomes create if missing -> check if existing index matches params.

That would mean that two scripts running at the same time could safely use the same index (they'd race on creation) and it would mean that if someone had a mismatch between index params and an existing index that would be detected and would error out.

Other option would be to run it at start before the ray part of the exercise.  To handle multiple scripts would still need the alternative API.

lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,160 @@
...
+            if self._index_params:
+                client = self._owner.client_t.from_client_params(self._client_params)
+                client.create_index_if_missing(self._index_params)
+
+        def write(self, blocks: Iterable[Block], ctx: TaskContext) -> Any:
+            builder = DelegatingBlockBuilder()
+            for block in blocks:
+                builder.add_block(block)
+            master_block = builder.build()
```
eric-anderson, line 125: all_blocks?

lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,160 @@
...
+
+            return ray_callable
+
+        ray_callable = build_ray_callable()
+        _check_serializable(ray_callable)
+
+        return dataset.map_batches(ray_callable, **self.resource_args)
+
+
+def _check_serializable(*objects):
```
eric-anderson, line 153: can we move this into some shared class and use it in basemap also (maybe the map rewrite will mean you don't need this anymore, but I can't tell for sure)

lib/sycamore/sycamore/writers/base.py
HenryL27: Going with TargetParams. it's shorter

lib/sycamore/sycamore/writers/base.py
HenryL27: disappears

lib/sycamore/sycamore/tests/unit/writers/test_base.py
```diff
@@ -0,0 +1,101 @@
...
+class FakeClient(BaseDBWriter.Client):
+    def __init__(self, client_params: "FakeClientParams"):
+        self.fspath = client_params.fspath
+
+    @classmethod
+    def from_client_params(cls, params: "BaseDBWriter.ClientParams") -> "FakeClient":
+        assert isinstance(params, FakeClientParams)
+        return FakeClient(params)
+
+    def write_many_records(self, records: list["BaseDBWriter.Record"], target_params: "BaseDBWriter.TargetParams"):
```
eric-anderson, line 17: Do you need the "" around the types? I thought that was only needed for forward declarations.

lib/sycamore/sycamore/tests/unit/writers/test_base.py
```diff
@@ -0,0 +1,101 @@
...
+        for r in records:
+            assert isinstance(r, FakeRecord) and isinstance(target_params, FakeTargetParams)
+            file = self.fspath / target_params.dirname / r.doc_id
+            file.write_text(r.text)
+
+    def create_target_idempotent(self, target_params: "BaseDBWriter.TargetParams"):
```
eric-anderson, line 23: Suggest adding test for existing not matching params by adding mode to the params.

lib/sycamore/sycamore/tests/unit/writers/test_base.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def test_fake_writer_e2e_happy(self, mocker, tmp_path):
+        input_node = mocker.Mock(spec=Node)
+        client_params = FakeClientParams(fspath=tmp_path)
+        target_params = FakeTargetParams(dirname="target")
+        writer = FakeWriter(input_node, client_params, target_params)
+        writer.write_docs(Common.docs)
+        target_path: Path = tmp_path / target_params.dirname
+        files = list(target_path.iterdir())
+        assert len(files) == len([d for d in Common.docs if not isinstance(d, MetadataDocument)])
```
eric-anderson, line 79: why do you have the MetadataDocument check?  docs is all pure docs.

lib/sycamore/sycamore/tests/unit/writers/test_base.py
```diff
@@ -0,0 +1,101 @@
...
+class TestBaseDBWriter(Common):
+
+    def test_fake_writer_e2e_happy(self, mocker, tmp_path):
+        input_node = mocker.Mock(spec=Node)
+        client_params = FakeClientParams(fspath=tmp_path)
+        target_params = FakeTargetParams(dirname="target")
+        writer = FakeWriter(input_node, client_params, target_params)
+        writer.write_docs(Common.docs)
+        target_path: Path = tmp_path / target_params.dirname
+        files = list(target_path.iterdir())
```
eric-anderson, line 78: suggest files.sort() here; otherwise I think this will be flaky on filesystems with hashed directories.

lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,76 @@
...
+            pass
+
+    # Type param for the objects to write to the db
+    class Record(ABC):
+        @classmethod
+        @abstractmethod
+        def from_doc(cls, document: Document, target_params: "BaseDBWriter.TargetParams") -> "BaseDBWriter.Record":
+            pass
+
+    # Type param for the object used to configure a new index if necessary
```
eric-anderson, line 40: a new target

if you think it's needed: , e.g. opensearch or pinecone index, weviate collection, filesystem directory, or s3 bucket.

lib/sycamore/sycamore/writers/base.py
```diff
@@ -0,0 +1,76 @@
...
+    class ClientParams(ABC):
+        pass
+
+    def __init__(
+        self,
+        plan: Node,
+        client_params: ClientParams,
+        target_params: TargetParams,
+        filter: Callable[[Document], bool] = lambda d: True,
+        **ray_remote_args,
```
eric-anderson, line 56: should we just call these kwargs?  You could pass in ones that go to MapBatch or BaseMap also.

eric-anderson (approved):

PR #452, opened 2024-06-07
Lazily clean up temp files from DETR partitioner.

alexaryn:
Tested on my laptop both cases (older and younger than cutoff).

### DIFF
--- a/lib/sycamore/sycamore/transforms/detr_partitioner.py
+++ b/lib/sycamore/sycamore/transforms/detr_partitioner.py
@@ -172,7 +172,7 @@ class SycamorePDFPartitioner:
 
             return deformable_layout
 
-    def _cleanup_tmp(self):
+    def _cleanup_tmp(self) -> None:
         now = time.time()
         dir = tempfile.gettempdir()
         for ent in os.scandir(dir):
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -161,6 +172,23 @@ def partition_pdf(
 
             return deformable_layout
 
+    def _cleanup_tmp(self):
+        now = time.time()
+        dir = tempfile.gettempdir()
+        for ent in os.scandir(dir):
+            if ent.name.startswith(self.tmp_prefix):
+                try:
+                    st = ent.stat()
```
HenryL27, line 181: what's ent?

HenryL27 (approved):
mypy be failing?

lib/sycamore/sycamore/transforms/detr_partitioner.py
alexaryn: Short for dirent (struct dirent in C), which is short for directory entry.  The Python type is os.DirEntry.

HenryL27: can I make you name the variable `entry` or something that in my head doesn't refer to a talking tree person?

alexaryn:
It looks like Python 3.9 TemporaryDirectory() lacks the ignore_cleanup_errors kwarg.  Should be harmless to include it.

PR #454, opened 2024-06-07
Allow disabling CUDA via env var

eric-anderson:
Some of our tests are unable to run on laptops because they try to use the gpu, but specify multiple stages that both need the GPU so hang indefinitely.


HenryL27 (approved):

alexaryn:
We should probably be thinking in terms of GPU, not specifically CUDA.  The Macs have MPS and we should be supporting that.  Rather than "should I use CUDA?", the question should be "what device should I use?"

PR #458, opened 2024-06-08
Choose MPS or CUDA automatically.

alexaryn:


HenryL27:
I think we don't want to do this until we upgrade torch to 2.3

HenryL27 (approved):

PR #461, opened 2024-06-11
convert opensearch writer to use base writer 1/3

HenryL27:
also fixed the query tests to actually run their ingest pipeline so they don't depend on going after other opensearch ITs (which now clean up after themselves)

### DIFF
--- a/lib/sycamore/sycamore/tests/integration/query/test_query_opensearch.py
+++ b/lib/sycamore/sycamore/tests/integration/query/test_query_opensearch.py
@@ -1,9 +1,55 @@
+import pytest
 import sycamore
 from sycamore.data.document import OpenSearchQuery
 from sycamore.tests.config import TEST_DIR
 from sycamore.transforms.embed import SentenceTransformerEmbedder
 from sycamore.transforms.partition import UnstructuredPdfPartitioner
 from sycamore.transforms.query import OpenSearchQueryExecutor
+from opensearchpy import OpenSearch
+
+
+@pytest.fixture(scope="class")
+def setup_index():
+    index_settings = {
+        "body": {
+            "settings": {
+                "index.knn": True,
+                "number_of_shards": 5,
+                "number_of_replicas": 1,
+            },
+            "mappings": {
+                "properties": {
+                    "embedding": {
+                        "type": "knn_vector",
+                        "dimension": 384,
+                        "method": {"name": "hnsw", "engine": "faiss"},
+                    },
+                    "text": {"type": "text"},
+                }
+            },
+        }
+    }
+    paths = str(TEST_DIR / "resources/data/pdfs/")
+
+    context = sycamore.init()
+    ds = (
+        context.read.binary(paths, binary_format="pdf")
+        .limit(1)
+        .partition(partitioner=UnstructuredPdfPartitioner())
+        .explode()
+        .embed(
+            embedder=SentenceTransformerEmbedder(batch_size=100, model_name="sentence-transformers/all-MiniLM-L6-v2")
+        )
+    )
+    ds.write.opensearch(
+        os_client_args=TestQueryOpenSearch.OS_CLIENT_ARGS,
+        index_name=TestQueryOpenSearch.INDEX,
+        index_settings=index_settings,
+    )
+    osc = OpenSearch(**TestQueryOpenSearch.OS_CLIENT_ARGS)
+    osc.indices.refresh(TestQueryOpenSearch.INDEX)
+    yield TestQueryOpenSearch.INDEX
+    osc.indices.delete(TestQueryOpenSearch.INDEX)
 
 
 class TestQueryOpenSearch:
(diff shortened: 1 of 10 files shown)


lib/sycamore/sycamore/writers/opensearch.py
```diff
@@ -82,65 +189,21 @@ def extract_os_document(data):
...
         }
+        data = document.data
         for k, v in default.items():
             if k in data:
                 result[k] = data[k]
             else:
                 result[k] = v
-        return result
-
-    @timetrace("OsrchWrite")
```
eric-anderson, line 92: recover the timetrace; Might need to use the non-decorator use since in the writer base class, you want to timetrace with `self.__class__.__name__`

lib/sycamore/sycamore/tests/integration/test_html_to_opensearch.py
```diff
@@ -25,7 +26,7 @@ def test_html_to_opensearch():
             "settings": {"index.knn": True, "number_of_shards": 5, "number_of_replicas": 1},
             "mappings": {
                 "properties": {
-                    "embeddings": {
+                    "embedding": {
```
eric-anderson, line 29: I'm dumb. Why did this change? If it was wrong before how was it working?

HenryL27: Before it was working by auto-creating a field called "embedding" that contains a list of floats. There are no vector queries in these tests so nothing broke. Now that we validate index configuration (targetparams compat check) this caused an error - when the ITs were not cleaning up after themselves and one of them had it configured correctly

lib/sycamore/sycamore/tests/integration/query/test_query_opensearch.py
```diff
@@ -1,9 +1,55 @@
...
+from opensearchpy import OpenSearch
+
+
+@pytest.fixture(scope="class")
+def setup_index():
+    index_settings = {
+        "body": {
+            "settings": {
+                "index.knn": True,
+                "number_of_shards": 5,
```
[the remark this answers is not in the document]
HenryL27: Not sure whether this helps; probly doesn't. It caught an error where opensearch hands the field back as "index.number_of_shards" - hence the awkward thing I did in the compat check

lib/sycamore/sycamore/writers/opensearch.py
```diff
@@ -1,77 +1,184 @@
...
+                    obj[i] = _string_values_to_python_types(obj[i])
+                return obj
+            if isinstance(obj, str):
+                if obj == "true":
+                    return True
+                elif obj == "false":
+                    return False
+                elif obj.isnumeric():
+                    return int(obj)
+                elif re.match("[0-9]+\.[0-9]+", obj):
```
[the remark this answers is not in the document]
HenryL27: apparently json doesn't allow things like .3 and since opensearch uses a standard json lib for all that I'm gonna say I get the 0. That said I dropped all the negative numbers

lib/sycamore/sycamore/writers/opensearch.py
```diff
@@ -1,77 +1,184 @@
...
+    hosts: list[HostAndPort] = field(default_factory=lambda: [HostAndPort(host="localhost", port=9200)])
+    http_compress: bool = True
+    http_auth: tuple[str, str] = ("admin", "admin")
+    use_ssl: bool = True
+    verify_certs: bool = True
+    ssl_assert_hostname: bool = True
+    ssl_show_warn: bool = True
+    timeout: Optional[int] = None
+    number_of_allowed_failures_per_block: int = 100
+    collect_failures_filepath: str = "failures.txt"
```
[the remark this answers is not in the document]
HenryL27: nice

lib/sycamore/sycamore/writer.py
```diff
@@ -21,8 +26,14 @@ def __init__(self, context: Context, plan: Node):
         self.plan = plan
 
     def opensearch(
```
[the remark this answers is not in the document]
HenryL27: I don't want to break stuff just yet. I feel like having mapwrite.opensearch() vs write.opensearch() is kind of equivalent to having an execute flag but then I have to make sure they're doing the same thing otherwise... 

I feel like if it turns out that the most common use-case is for the write to be nonterminal, then we can just flip the flag default, right? Alternatively maybe some data in the sycamore context object could be used? Will the lineage stuff be explicit in all pipelines?

lib/sycamore/sycamore/writers/opensearch.py
```diff
@@ -147,7 +131,7 @@ def _string_values_to_python_types(obj: Any):
                     return False
                 elif obj.isnumeric():
                     return int(obj)
-                elif re.match("[0-9]+\.[0-9]+", obj):
+                elif re.fullmatch("-?[0-9]+\.[0-9]+", obj):
```
eric-anderson, line 134: Good catch on -. We both forgot about scientific notation.  The internet suggests 
try: 
  return float(obj)
except ValueError:
  return obj

lib/sycamore/sycamore/writers/base.py
```diff
@@ -77,3 +77,10 @@ def write_docs(self, docs: list[Document]) -> list[Document]:
         records = [self.Record.from_doc(d, created_target_params) for d in docs if self._filter(d)]
         client.write_many_records(records, self._target_params)
         return docs
+
+    def _write_docs_tt(self, docs: list[Document]) -> list[Document]:
+        if self._name:
+            with TimeTrace(self._name):
+                return self.write_docs(docs)
+        else:
```
eric-anderson, line 85: with TimeTrace("UnknownWriter"):

eric-anderson (approved):

PR #463, opened 2024-06-12
Move DETR partitioner to use transformers.pipeline

baitsguy:
Simplifies the inference code slightly and adds batch support.

note: I haven't changed the model server response parsing yet, need to verify the API there.


bsowell (commented):
We might end up changing some of this to better handle large pdfs. I am okay if you want to get this in, just with the understanding that it might change again. 

Can you remove the empty file "gpu_pdf.py"?

bsowell (approved):

baitsguy:
> We might end up changing some of this to better handle large pdfs. I am okay if you want to get this in, just with the understanding that it might change again.
> 
> Can you remove the empty file "gpu_pdf.py"?

@bsowell I didn't to send this out, I'm closing this PR since the pipelines out of the box has quality issues.

PR #465, opened 2024-06-12
convert weaviate writer to base db writer 2/3

HenryL27:
convert the weaviate writer to a basedbwriter map. Rather, two - one for docs and one for references - so I can stage them (references must follow docs).
This doesn't change the .write.weaviate() api (though it adds an `execute` param for continual laziness), but we might want to change the API to be a little more strongly typed (that goes for write.opensearch() too).


lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -1,145 +1,235 @@
...
+# this class. This kinda stuff is all over the wv8 codebase
+CollectionConfigCreate: TypeAlias = _CollectionConfigCreate
+
+
+@dataclass
+class WeaviateTargetParams(BaseDBWriter.TargetParams):
+    name: str
+    collection_config: Union[CollectionConfigCreate, CollectionConfig]
+
+    def compatible_with(self, other: BaseDBWriter.TargetParams) -> bool:
```
bsowell, line 44: This may just be my ignorance on the base model, but when is this called?

lib/sycamore/sycamore/tests/unit/test_writer.py
bsowell, line 1: Do we have any tests of the cross-reference writer? It's not immediately obvious to me when/how to use it.

bsowell (commented):
Couple of questions. Nothing major.

lib/sycamore/sycamore/writers/weaviate_writer.py
HenryL27: [here](https://github.com/aryn-ai/sycamore/blob/87d3123ae23c686512b010f18ccaf96d9c5258cf/lib/sycamore/sycamore/writers/base.py#L71). The target creation logic is:
1. Try to create the target, but don't fail if it's already there
2. Get metadata about the target and verify that it matches what you expect
3. Write documents

This lets us scale writes horizontally without worrying about a singular 'start_write' step, and also lets us catch trying to write to a target that exists already and is incompatible (e.g. diff vector dimensions or something)

lib/sycamore/sycamore/tests/unit/test_writer.py
HenryL27: Sorta? The original wv8 writer worked in two steps: first, write the documents, then write the parent-child cross-refs. What I've done here is separate those phases as independent "Writer" objects. The `docset.write.weaviate()` call chains them though - so you get both by default.

lib/sycamore/sycamore/writers/weaviate_writer.py
```diff
@@ -1,145 +1,235 @@
...
-            if embedding is not None:
-                object["vector"] = {"embedding": embedding}
-            if parent:
-                object["references"] = {"parent": parent}
-            return object
-
-        records = BlockAccessor.for_block(block).to_arrow().to_pylist()
-        all = [record_to_object(r) for r in records]
-        return [a for a in all if a is not None]
+                    cfg_crt = target_params.collection_config
```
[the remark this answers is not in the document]
HenryL27: one of _CollectionConfigCreate or CollectionConfig, which are both objects from Weaviate. _CollectionConfigCreate is constructable by a user, but `client.collections.get("MyCollection").config.get()` gives back a CollectionConfig object obtained by parsing a fully fleshed out json schema

bsowell (approved):

PR #468, opened 2024-06-13
convert pinecone writer to base writer 3/3

HenryL27:
This one is largely mechanical, to the degree that one of these can be mechanical.

n.b. I added a `to_dict()` method to bbox that returns the coords named. useful for flattening etc


lib/sycamore/sycamore/writer.py
```diff
@@ -242,8 +242,9 @@ def pinecone(
         dimensions=None,
         distance_metric="cosine",
         api_key=None,
-        **resource_args,
-    ):
+        execute=True,
```
karansampath, line 245: nit: type annotations

karansampath (approved):
Looks good to me, made a small comment

PR #469, opened 2024-06-13
Add LogTime that logs time trace info via logging

eric-anderson:
Instrument detr_partitioner with the logging to help debug some performance issues.

### DIFF
--- a/lib/sycamore/sycamore/utils/time_trace.py
+++ b/lib/sycamore/sycamore/utils/time_trace.py
@@ -4,7 +4,58 @@ import struct
 import resource
 import threading
 import functools
+import logging
 from sys import platform
+
+zero_time = time.time_ns()
+# Mac reports ru_maxrss in bytes. Linux reports it in kB.
+if platform == "darwin":
+    RSS_MULTIPLIER = 1
+else:
+    RSS_MULTIPLIER = 1024
+
+
+class TimeTraceData:
+    def __init__(self, t0, t1, user, sys, rss):
+        self.t0 = t0  # ns
+        self.t1 = t1  # ns
+        self.user = user  # ns
+        self.sys = sys  # ns
+        self.rss = rss  # bytes
+
+    def wall_s(self):
+        return (self.t1 - self.t0) / 1.0e9
+
+    def user_s(self):
+        return self.user / 1.0e9
+
+    def sys_s(self):
+        return self.sys / 1.0e9
+
+    def rss_mib(self):
+        return self.rss / (1024 * 1024)
+
+
+class InMemoryTimeTrace:
+    try:
+        resource_type = resource.RUSAGE_THREAD  # type: ignore
+    except AttributeError:
+        resource_type = resource.RUSAGE_SELF
+
+    def __init__(self):
+        self.t0 = time.time_ns()
+        self.r0 = resource.getrusage(self.resource_type)
+
+    def end(self):
+        t1 = time.time_ns()
+        r1 = resource.getrusage(self.resource_type)
+        r0 = self.r0
+        user = int((r1.ru_utime - r0.ru_utime) * 1.0e9)
+        sys = int((r1.ru_stime - r0.ru_stime) * 1.0e9)
+
+        rss = r1.ru_maxrss * RSS_MULTIPLIER  # could do max(r0,r1), but this is more specific
+
+        return TimeTraceData(self.t0, t1, user, sys, rss)
 
 
 class TimeTrace:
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/utils/time_trace.py
```diff
@@ -28,34 +79,23 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        self.t0 = time.time_ns()
-        self.r0 = resource.getrusage(self.resource_type)
+        self.imtt = InMemoryTimeTrace()
 
     def end(self):
         if TimeTrace.fd < 0:
             return
-        t1 = time.time_ns()
-        r1 = resource.getrusage(self.resource_type)
+        self.data = self.imtt.end()
```
alexaryn, line 87: Do we need to store data in the object, or can it just be a local variable?

lib/sycamore/sycamore/utils/time_trace.py
```diff
@@ -95,3 +135,45 @@ def wrapper(*args, **kwargs):
...
+
+
+class _ZeroRU:
+    def __init__(self):
+        self.ru_utime = 0
+        self.ru_stime = 0
+
+
+class LogTime:
+    def __init__(self, name, point: bool = False, log_start: bool = False):
```
alexaryn, line 147: Might want to add `, /, *,` after `name` to force keyword usage.

lib/sycamore/sycamore/utils/time_trace.py
```diff
@@ -95,3 +135,45 @@ def wrapper(*args, **kwargs):
...
+            self._logpoint(name)
+            return
+
+        self.name = name
+        self.log_start = log_start
+
+    def __enter__(self):
+        self.start()
+        if self.log_start:
+            self._logpoint(self.name + "_start")
```
alexaryn, line 158: Any reason not to have this code in start()?

lib/sycamore/sycamore/utils/time_trace.py
```diff
@@ -4,8 +4,59 @@
...
+    def end(self):
+        t1 = time.time_ns()
+        r1 = resource.getrusage(self.resource_type)
+        r0 = self.r0
+        user = int((r1.ru_utime - r0.ru_utime) * 1.0e9)
+        sys = int((r1.ru_stime - r0.ru_stime) * 1.0e9)
+
+        rss = r1.ru_maxrss * RSS_MULTIPLIER  # could do max(r0,r1), but this is more specific
+
+        return TimeTraceData(self.t0, t1, user, sys, rss)
```
alexaryn, line 58: It's a bit surprising to have end() return computed data.  Maybe a name more like "reap" would be clearer.  There's no start() that needs and end() counterpoint.

lib/sycamore/sycamore/utils/time_trace.py
```diff
@@ -4,8 +4,59 @@
...
+class TimeTraceData:
+    def __init__(self, t0, t1, user, sys, rss):
+        self.t0 = t0  # ns
+        self.t1 = t1  # ns
+        self.user = user  # ns
+        self.sys = sys  # ns
+        self.rss = rss  # bytes
+
+    def wall_s(self):
+        return (self.t1 - self.t0) / 1.0e9
```
alexaryn, line 27: It seems you're not a fan of `1000000000.0`.  Not everybody likes scientific notation, either.  Perhaps it's time for some constant like `BILLION` or `NANOS_IN_ONE`.

lib/sycamore/sycamore/utils/time_trace.py
```diff
@@ -4,8 +4,59 @@
...
+
+    def user_s(self):
+        return self.user / 1.0e9
+
+    def sys_s(self):
+        return self.sys / 1.0e9
+
+    def rss_mib(self):
```
alexaryn, line 35: I know why people use MiB instead of MB, but still whenever I see mib I think of SNMP.

alexaryn (approved):
Looks useful.

One thing to consider is that when used as a context manager, there are additional arguments that __exit__ can take that indicate if there was an exception.  Timings from a block of code that errored-out may not be valid.  With all the OOMing lately, that became a bit of an issue with logging.

Do you want to have more type hints in this code?

lib/sycamore/sycamore/utils/time_trace.py
eric-anderson: My thinking was that the explicit end function seemed weird without a way to get at the result.  I'm fine with changing if if you don't think that's useful, but in that case, I'd collapse everything into __enter__ and __exit__.

lib/sycamore/sycamore/utils/time_trace.py
eric-anderson: done, although the internet tells me it's just `, *,`

lib/sycamore/sycamore/utils/time_trace.py
eric-anderson: I had thought that log_start outside of context didn't make sense, but changed.

lib/sycamore/sycamore/utils/time_trace.py
eric-anderson: changed to .measure()

lib/sycamore/sycamore/utils/time_trace.py
eric-anderson: Yes, because it's hard to tell if its 100000000.0. Do you object to scientific?  It's compact and explicit. I can change it to a constant although I don't understand why that's better.

lib/sycamore/sycamore/utils/time_trace.py
eric-anderson: Changed.

PR #470, opened 2024-06-13
Add ArynPartitioner

MarkLindblad:
A `Partitioner` that acts as an inline client to the Aryn Partitioning Service.


lib/sycamore/sycamore/transforms/aryn_partitioner.py
```diff
@@ -0,0 +1,50 @@
...
+    ):
+        options = {
+            "threshold": threshold,
+            "use_ocr": use_ocr,
+            "ocr_images": ocr_images,
+            "ocr_tables": ocr_tables,
+            "extract_table_structure": extract_table_structure,
+            "extract_images": extract_images,
+        }
+        files: Mapping = {"pdf": file, "options": json.dumps(options).encode("utf-8")}
```
pparmar30, line 36: why are you sending Mapping as a file? Can we send options in the body instead?

lib/sycamore/sycamore/transforms/aryn_partitioner.py
```diff
@@ -0,0 +1,50 @@
...
+        header = {"Authorization": f"Bearer {token}"}
+        response = requests.post(aryn_partitioner_address, files=files, headers=header)
+
+        if response.status_code != 200:
```
pparmar30, line 40: should we retry depending on the error code? Also, what does a failure mean for sycamore pipeline?

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -488,6 +490,82 @@ def partition(self, document: Document) -> Document:
         return document
 
 
+class ArynPartitioner(Partitioner):
```
pparmar30, line 493: Let's update the unit/integ tests. Also, are we still keeping around SycamorePartitioner?

MarkLindblad: We should definitely keep `SycamorePartitioner` around. The question is whether we keep the `model_server_endpoint` part of the `SycamorePartitioner` around.

lib/sycamore/sycamore/transforms/aryn_partitioner.py
MarkLindblad: mypy expects the `files` variable to be a `Mapping` not a `dict` when used in `requests.post(...)`.  Let's keep the API as-is: we need to upload both `options` and the pdf itself.

sliedes: Sounds like there's some confusion here. A `dict` is a `Mapping`. If a function takes a `Mapping`, you can pass it anything that has `__getitem__`, `__iter__` and `__len__` (of the proper type, if it specifies type variables, like `Mapping[str, str]`)—including a `dict`.

Think of `Mapping` as an interface of objects that quack like a `dict`. (There's also `MutableMapping` which a function should take if it doesn't absolutely require it to be a `dict`, but it wants to modify it.)

MarkLindblad: I agree a `dict` is a `Mapping`. For some reason mypy does not see the `dict[str, BinaryIO | bytes]` as a `Mapping[str, SupportsRead[bytes]]`, so I annotated it which silences the error.

`lib/sycamore/sycamore/transforms/aryn_partitioner.py:38: error: Argument "files" to "post" has incompatible type "dict[str, Iterable[object]]"; expected "Mapping[str, SupportsRead[str | bytes] | str | bytes | tuple[str | None, SupportsRead[str | bytes] | str | bytes] | tuple[str | None, SupportsRead[str | bytes] | str | bytes, str] | tuple[str | None, SupportsRead[str | bytes] | str | bytes, str, Mapping[str, str]]] | Iterable[tuple[str, SupportsRead[str | bytes] | str | bytes | tuple[str | None, SupportsRead[str | bytes] | str | bytes] | tuple[str | None, SupportsRead[str | bytes] | str | bytes, str] | tuple[str | None, SupportsRead[str | bytes] | str | bytes, str, Mapping[str, str]]]] | None"  [arg-type]`

sliedes: I think the issue is that `bytes` is not a `SupportsRead[bytes]`—I don't think `bytes` has a `read` method. So in this case I think mypy is correct. It passes as a `Mapping` only because that is (an unsafe, --strict would complain) alias for `Mapping[Any,Any]`. Does the code actually work if passed bytes? And the annotation of the called function is not unnecessarily strict?

lib/sycamore/sycamore/transforms/aryn_partitioner.py
MarkLindblad: Yes. We should retry when we get an error saying try again later.  I'll add this.

lib/sycamore/sycamore/tests/unit/transforms/test_aryn_partitioner.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def text(self) -> str:
+        return ""
+
+
+class TestArynPDFPartitioner:
+    def test_partition(self, mocker) -> None:
+        mocker.patch("requests.post", return_value=MockResponse())
+        with open(TEST_DIR / "resources/data/pdfs/Ray.pdf", "rb") as pdf:
+            ArynPDFPartitioner.partition_pdf(pdf, aryn_token="")
```
pparmar30, line 22: Let's assert on the response from  ArynPDFPartitioner.partition_pdf call! Same below

lib/sycamore/sycamore/transforms/aryn_partitioner.py
```diff
@@ -0,0 +1,68 @@
...
+        while (
+            last_status_code == 500 and last_message == _ARYN_PARTITIONING_SERVICE_WAIT_MESSAGE and tries < max_retries
+        ):
+            response = requests.post(aryn_partitioner_address, files=files, headers=header)
+            last_status_code = response.status_code
+            last_message = response.text
+            if tries > 0:
+                time.sleep(10 * (2**tries))  # Modified binary exponential backoff
+            tries += 1
+
```
pparmar30, line 57: You can simplify the retry logic using tenacity : https://github.com/aryn-ai/managed-service/blob/main/aryn/lib/utils.py#L133

lib/sycamore/sycamore/transforms/aryn_partitioner.py
```diff
@@ -0,0 +1,73 @@
...
+
+class ArynPDFPartitionerException(Exception):
+    def __init__(self, message):
+        super().__init__(message)
+
+
+class ArynContinuePDFPartitionerException(Exception):
+    def __init__(self, message):
+        super().__init__(message)
```
pparmar30, line 22: Instead of having two classes, you can just have a flag in  ArynPDFPartitionerException : can_retry: bool

lib/sycamore/sycamore/transforms/partition.py
bsowell: I don't think we ever expect the model_server_endpoint to be used at this point, do we? There's no way for external people to call it right? You don't need to remove it as part of this PR, but I do think we should remove it.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -490,6 +492,83 @@ def partition(self, document: Document) -> Document:
...
+        aryn_token: str,
+        threshold: float = 0.4,
+        use_ocr: bool = False,
+        ocr_images: bool = False,
+        ocr_tables: bool = False,
+        extract_table_structure: bool = False,
+        extract_images: bool = False,
+        aryn_partitioner_address: str = _DEFAULT_ARYN_PARTITIONER_ADDRESS,
+    ):
+        super().__init__(device="cpu", batch_size=1)
```
bsowell, line 537: I know we discussed this before. Can you remind me about the batch size? Did we decide we are always sending one PDF at a time?

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -490,6 +492,83 @@ def partition(self, document: Document) -> Document:
...
+        self._threshold = threshold
+        self._use_ocr = use_ocr
+        self._ocr_images = ocr_images
+        self._ocr_tables = ocr_tables
+        self._extract_table_structure = extract_table_structure
+        self._extract_images = extract_images
+        self._aryn_partitioner_address = aryn_partitioner_address
+
+    def partition(self, document: Document):
+        binary = io.BytesIO(document.data["binary_representation"])
```
bsowell, line 548: Minor: I would prefer `document.binary_representation` rather than accessing the internal dictionary directly. That way any other code we put in the property accessor is invoked.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -490,6 +492,83 @@ def partition(self, document: Document) -> Document:
...
+        from sycamore.transforms.aryn_partitioner import ArynPDFPartitioner
+
+        try:
+            result = ArynPDFPartitioner.partition_pdf(
```
bsowell, line 552: I'm not sure it's worth refactoring again, but what is the rationale for splitting out the ArynPDFPartitioner class, other than that's what the SycamorePartitioner did? In that case there was enough code and enough change of using a different model that maybe it made sense to split them, but I'm not sure that is the case here.

pparmar30 (approved):

PR #474, opened 2024-06-17
Instrumented more code with TimeTrace decorators.

alexaryn:


lib/sycamore/sycamore/transforms/extract_schema.py
```diff
@@ -142,6 +144,7 @@ def __init__(
         self._num_of_elements = num_of_elements
         self._prompt_formatter = prompt_formatter
 
+    @timetrace("ExtrProps")
```
bsowell, line 147: Not a big deal (and I've approved this PR), but do you think it's worth doing something like automatically annotate all transforms -- you could probably just use the name of the class or do some kind of metaclass shenanigans to make it easy. Or do we want to wait until the lineage stuff is in?

bsowell (approved):

lib/sycamore/sycamore/transforms/extract_schema.py
alexaryn: The issue is that long names don't work in visualizations.  We'll definitely encounter this issue with the lineage stuff.  We may need to distinguish names in code from tags.  And, we really don't want timing of certain trivial things.

PR #475, opened 2024-06-18
Fix for jupyter notebook tutorial

RitxmSaha:
Through going through the tutorial, the pdf's didn't download properly cause it accidently tried to pull from a arxiv link with ".pdf" on the end(making it invalid). This PR fixes this.


HenryL27 (commented):
was very confused cuz I thought we had fixed this but then I realized it's a change to the docs haha. 
the notebook this is based on [here](https://github.com/aryn-ai/sycamore/blob/main/notebooks/jupyter_dev_example.ipynb) uses _location still. Unless something's changed in the ui - ask soham - we should prob keep that. (Also properties starting with an underscore are automatically spread by explode, which is a very useful feature)

PR #476, opened 2024-06-18
Add opensearch reader

baitsguy:
We can improve this to do parallel scrolls/loads once performance becomes an issue. For existing datasets, it hasn't taken more than a couple seconds.


lib/sycamore/sycamore/tests/integration/scans/test_db_scan.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # Delete after
+    client.indices.delete(TestOpenSearchScan.INDEX, ignore_unavailable=True)
+
+
+def filter_doc(obj, exclude):
+    return {k: v for k, v in obj.__dict__.items() if k not in exclude}
+
+
+def compare_docs(doc1, doc2):
```
eric-anderson, line 46: I'd suggest using a keep-list rather than an exclude list. I expect you will need to exclude more than just binary_representation. If you make https://github.com/aryn-ai/sycamore/blob/b0e3ecaf21550c3d2258dd41b5b989e23bba03a1/lib/sycamore/sycamore/writers/opensearch.py#L167 a class variable, then you could allow exactly the values the opensearch writer writes.

lib/sycamore/sycamore/tests/integration/scans/test_db_scan.py
```diff
@@ -0,0 +1,93 @@
...
+            context.read.binary(paths, binary_format="pdf")
+            .limit(1)
+            .partition(partitioner=UnstructuredPdfPartitioner())
+            .explode()
+            .write.opensearch(
+                os_client_args=TestOpenSearchScan.OS_CLIENT_ARGS, index_name=TestOpenSearchScan.INDEX, execute=False
+            )
+        )
+
+        retrieved_docs = context.read.opensearch(
```
eric-anderson, line 85: Not sure why but in my testing, this is returning nothing (I couldn't figure out how it was working given stuff like lineage_id which will be in original but not retrieved).
adding `assert len(original_materialized) == len(retrieved_materialized)`
makes the test fail, and is important to check because len(zip([1,2,3],[])) == 0.

lib/sycamore/sycamore/scans/db_scan.py
eric-anderson, line 1: maybe name this opensearch_scan?

Then each implementation gets its own file.

lib/sycamore/sycamore/scans/db_scan.py
```diff
@@ -0,0 +1,40 @@
...
+        self.query = query
+
+    def execute(self) -> Dataset:
+        result = []
+
+        scroll = "1m"
+        response = self.os_client.search(index=self.index_name, scroll=scroll, size=200, body=self.query)
+        scroll_id = response["_scroll_id"]
+
+        while True:
```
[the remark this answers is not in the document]
baitsguy: Yeah makes sense

lib/sycamore/sycamore/tests/integration/scans/test_db_scan.py
```diff
@@ -0,0 +1,93 @@
...
+        "ssl_show_warn": False,
+        "timeout": 120,
+    }
+
+    def test_ingest_and_read(self, setup_index):
+        """
+        Validates data is readable from OpenSearch, and that we can rebuild processed Sycamore documents.
+        """
+
+        paths = str(TEST_DIR / "resources/data/pdfs/")
```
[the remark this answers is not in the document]
baitsguy: Kept it to be consistent with other tests, but since we're just dealing with 1 file here I can just change

lib/sycamore/sycamore/tests/integration/scans/test_db_scan.py
baitsguy: Good idea - we'll eventually want to abstract this across writers too

lib/sycamore/sycamore/scans/db_scan.py
baitsguy: Keeping it like this so we're consistent with file_scan and materialized_scan, but open to refactoring separately

lib/sycamore/sycamore/tests/integration/scans/test_db_scan.py
baitsguy: Discussed offline, looks like the query can sometimes happen before the data has finished indexing. Hacking a sleep currently to avoid tuning cluster settings for a test.

lib/sycamore/sycamore/scans/db_scan.py
```diff
@@ -0,0 +1,40 @@
...
+            hits = response["hits"]["hits"]
+            for hit in hits:
+                result += [Document(hit["_source"])]
+
+            if not hits:
+                break
+            response = self.os_client.scroll(scroll_id=scroll_id, scroll=scroll)
```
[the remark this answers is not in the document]
baitsguy: Good question, I don't know either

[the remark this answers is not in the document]
eric-anderson: Please then do a manual test with the timeout set to 1s so that the total download takes several seconds. If it returns the same thing then it's extending the timeout.  My guess is that it is, but I couldn't find anything definitive.

lib/sycamore/sycamore/scans/db_scan.py
eric-anderson: ok, my view was that we would probably put readers for weviate and pinecone in separate files (like was done for the writers), but maybe they'll be small and we'll put them all together.

eric-anderson (approved):

lib/sycamore/sycamore/scans/db_scan.py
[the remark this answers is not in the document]
baitsguy: Validated, added a 0.1s sleep between each call, reduced the size for each call to 1 record, and used a timeout of 1s. Works fine and took ~1m to fully run.

PR #478, opened 2024-06-18
Batched sycamore pdf partitioner

eric-anderson:
Operates in batches of pages rather than processing all pages through each stage at a time As a result lowers memory usage.

Also pipelines the pdfminer and model work which speeds up runtime.

Lots of instrumentation left in since running over the 2k page test document memory grows without bound.


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -30,6 +40,31 @@ def _batchify(iterable, n=1):
         yield iterable[i : min(i + n, length)]
 
 
+def display_top(snapshot, key_type="lineno", limit=10):
```
baitsguy, line 43: does this belong outside?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+        extract_images=False,
+        batch_size: int = 1,
+    ) -> List[List["Element"]]:
+        if not table_structure_extractor:
+            table_structure_extractor = DEFAULT_TABLE_STRUCTURE_EXTRACTOR(device=self.device)
+
+        pdfminer = None
+        exec = ProcessPoolExecutor(max_workers=1)
+        if not use_ocr:
+            with LogTime("start_pdfminer", log_start=True):
```
bohou-aryn, line 301: Do we have proof of how this additional process for pdfminer help?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+
+        def capture_exception(q, fn, finish_msg):
+            try:
+                fn()
+            except e:
+                q.put(e)
+
+            q.put(finish_msg)
+
+        def read_stdout(fh, q):
```
baitsguy, line 373: same as above, do we need all this in the detr_partitioner file? It's making it hard to read just the partitioning logic

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+        if not use_ocr:
+            with LogTime("start_pdfminer", log_start=True):
+                pdfminer = exec.submit(self._run_pdfminer, filename)
+
+        deformable_layout = []
+        if tracemalloc.is_tracing():
+            before = tracemalloc.take_snapshot()
+        for i in self._batch_pdf_convert(filename, batch_size):
+            parts = self.process_batch(
+                i,
```
baitsguy, line 309: there's a bunch of flags we pass around, named arguments are probably useful so we don't mess up

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+            table_structure_extractor = DEFAULT_TABLE_STRUCTURE_EXTRACTOR(device=self.device)
+
+        pdfminer = None
+        exec = ProcessPoolExecutor(max_workers=1)
+        if not use_ocr:
+            with LogTime("start_pdfminer", log_start=True):
+                pdfminer = exec.submit(self._run_pdfminer, filename)
+
+        deformable_layout = []
+        if tracemalloc.is_tracing():
```
bohou-aryn, line 305: where we start/stop this, or how do we config to enable/dsiable it?

baitsguy (commented):
Can you move some of the general comparison, measurement, etc. outside the partitioner? We seem to have mixed a lot of performance/execution code with the partitioning/parsing logic

Also, can you put the numbers you have from experiments in the PR description for traceability.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+                    self._supplement_text(d, p)
+
+        if tracemalloc.is_tracing():
+            (current, peak) = tracemalloc.get_traced_memory()
+            logging.info(f"Memory Usage current={current} peak={peak}")
+            top = tracemalloc.take_snapshot()
+            display_top(top)
+        return deformable_layout
+
+    def _batch_pdf_convert(self, filename: str, batch_size: int) -> Generator[List[Image.Image], None, None]:
```
bohou-aryn, line 346: what is this used for?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+        ocr_tables=False,
+        extract_table_structure=False,
+        table_structure_extractor=None,
+        extract_images=False,
+        batch_size: int = 1,
+    ) -> List[List["Element"]]:
+        LogTime("partition_start", point=True)
+        with tempfile.NamedTemporaryFile(prefix="detr-pdf-input-") as pdffile:
+            with LogTime("write_pdf"):
+                data = file.read()
```
alexaryn, line 262: Is this ever an actual file, making this code redundant?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+                data_len = len(data)
+                pdffile.write(data)
+                data = b""
```
alexaryn, line 265: I think it's more idiomatic to do: `del data`

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+                pdffile.flush()
+                logging.info(f"Wrote {pdffile.name}")
+            stat = os.stat(pdffile.name)
+            assert stat.st_size == data_len
+            assert os.path.isfile(pdffile.name), f"missing {pdffile.name}"
```
alexaryn, line 270: Why not use the result from stat() instead of causing another one here?  I'd do `stat.S_ISREG(stat.st_mode)`

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +244,276 @@ def partition_pdf(
...
+        use_ocr=False,
+        ocr_images=False,
+        ocr_tables=False,
+        extract_table_structure=False,
+        table_structure_extractor=None,
+        extract_images=False,
+        batch_size: int = 1,
+    ) -> List[List["Element"]]:
+        if not table_structure_extractor:
+            table_structure_extractor = DEFAULT_TABLE_STRUCTURE_EXTRACTOR(device=self.device)
```
alexaryn, line 296: Do we want to construct one of these even if `extract_table_structure` is false?

alexaryn (commented):
Looks good.  More code than I expected.

It would be cool if the memory debugging stuff had a file all it's own.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Yes, on some of our test documents, the dominant time is pdfminer, but the image processing still takes ~1 minute, so it leads to a 60s speedup.  It depends on the documents for some documents pdfminer is fast.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: You call tracemalloc.start() somewhere.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: It returns a list of batches of images since we thought batching was useful.  It's used at line 307.

eric-anderson: Used in function _partition_pdf_batched_named

lib/sycamore/sycamore/tests/unit/transforms/compare_detr_impls.py
```diff
@@ -0,0 +1,69 @@
+from sycamore.data import Element
+
+
+def assert_deep_eq(a, b, path):
```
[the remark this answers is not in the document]
eric-anderson: Done. I'm not sure if it's general enough.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: moved to a separate utils file.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Moved to utils.pdf

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Done. Didn't seem very useful to me, but we'll see.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Done.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: It was an ordering thing, I wrote the isfile check first and later had issues so came back to do the stat. Stat is now irrelevant, all that matters is the thing exists of the right size and the stat checks the exists condition.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: It's what the old code did, but there's no reason for it so fixed.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: it may be a file or may not. Eventually we'll fix the APIs to call the named version if you have a name.

lib/sycamore/sycamore/utils/deep_eq.py
```diff
@@ -0,0 +1,49 @@
...
+        assert b is None
+        return True
+
+    if isinstance(a, int) or isinstance(a, float) or isinstance(a, str):
+        assert a == b, f"values {a} {b} at {path}"
+        return True
+
+    if isinstance(a, list) or isinstance(a, tuple):
+        assert len(a) == len(b), f"length {len(a)} {len(b)} {path}"
+        for i, v in enumerate(a):
```
alexaryn, line 16: You're not using v.  Either change a[i] to v or make the loop over range(len(a)).

lib/sycamore/sycamore/utils/deep_eq.py
```diff
@@ -0,0 +1,49 @@
...
+            assert_deep_eq(a[i], b[i], path + [i])
+        return True
+
+    if isinstance(a, dict):
+        for k, v in a.items():
```
alexaryn, line 21: Can just have: for k in a.keys(): as I don't see v being used.

lib/sycamore/sycamore/utils/deep_eq.py
```diff
@@ -0,0 +1,49 @@
...
+        return True
+    assert False, f"Don't know how to compare {a}/{type(a)} with {b} at {path}"
+
+
+def deep_eq(a, b):
+    try:
+        assert_deep_eq(a, b, [])
+        return True
+    except AssertionError as e:
+        print(f"Fail {e}")
```
alexaryn, line 48: Is this leftover output?

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -0,0 +1,115 @@
...
+                data = data + part
+            else:
+                logging.debug(f"no reading. have {len(data)}/{need_bytes}")
+
+            if len(data) < need_bytes:
+                continue
+
+            code, size, rgb = tuple(data[0:HEADER_BYTES].split(b"\n")[0:3])
+            size_x, size_y = tuple(size.split(b" "))
+            file_size = len(code) + len(size) + len(rgb) + 3 + int(size_x) * int(size_y) * 3
```
alexaryn, line 48: Why not assign to need_bytes here.  Then the logic below matches above: if len(data) < need_bytes: continue.  I don't think file_size is needed.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -0,0 +1,115 @@
...
+            data = data[file_size:]
+            need_bytes = HEADER_BYTES
+
+    def read_stderr(fh, q):
+        while True:
+            line = fh.readline()
+            if line == b"":
+                break
+
+            q.put(str(line, encoding="utf-8").rstrip())
```
alexaryn, line 65: I think it's preferred to have: line.decode().rstrip().  No need for str() as line is already a str.  UTF-8 is the default for decode().

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -0,0 +1,115 @@
...
+
+    with LogTime("convert_to_image"):
+        # If we don't do this, then if the stderr buffer fills up we could get stuck.
+        # Popen.communicate() reads the entire strings.
+        args = ["pdftoppm", "-r", "200", pdf_path]
+        proc = Popen(args, stdout=PIPE, stderr=PIPE)
+        q: Queue = Queue()
+        t_out = Thread(target=capture_exception, args=(q, lambda: read_stdout(proc.stdout, q), StdoutEOF()))
```
alexaryn, line 73: What joins or otherwise cleans up these threads?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +217,171 @@ def partition_pdf(
...
+
+        pdfminer = None
+        exec = ProcessPoolExecutor(max_workers=1)
+        if not use_ocr:
+            with LogTime("start_pdfminer", log_start=True):
+                pdfminer = exec.submit(self._run_pdfminer, filename)
+
+        deformable_layout = []
+        if tracemalloc.is_tracing():
+            before = tracemalloc.take_snapshot()
```
alexaryn, line 278: Would some RAII-style class be useful for this snapshot comparing stuff, kind of like TimeTrace/LogTime?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +217,171 @@ def partition_pdf(
...
+            if tracemalloc.is_tracing():
+                gc.collect()
+                after = tracemalloc.take_snapshot()
+                top_stats = after.compare_to(before, "lineno")
+
+                print("[ Top 10 differences ]")
+                for stat in top_stats[:10]:
+                    print(stat)
+                before = after
+                display_top(after)
```
alexaryn, line 301: I would think some tracemalloc helper class, perhaps the RAII one, could absorb code like this.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +217,171 @@ def partition_pdf(
...
+            assert len(pdfminer_layout) == len(deformable_layout), f"{len(pdfminer_layout)} vs {len(deformable_layout)}"
+            with LogTime("pdfminer_supplement"):
+                for d, p in zip(deformable_layout, pdfminer_layout):
+                    self._supplement_text(d, p)
+
+        if tracemalloc.is_tracing():
+            (current, peak) = tracemalloc.get_traced_memory()
+            logging.info(f"Memory Usage current={current} peak={peak}")
+            top = tracemalloc.take_snapshot()
+            display_top(top)
```
alexaryn, line 315: It would be cool if it's easy to hide this dirty laundry.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +217,171 @@ def partition_pdf(
...
+
+        count = 0
+        for obj in gc.get_objects():
+            try:
+                if torch.is_tensor(obj) or (hasattr(obj, "data") and torch.is_tensor(obj.data)):
+                    # print(type(obj), obj.size())
+                    count = count + 1
+            except Exception as e:
+                print(f"Exception {e}")
+        print(f"Found {count} tensors")
```
alexaryn, line 340: Could this go in memory_debugging.py ?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -252,6 +466,7 @@ def infer(
                 result["scores"] = result["scores"].tolist()
                 result["labels"] = result["labels"].tolist()
                 result["boxes"] = result["boxes"].tolist()
+
```
alexaryn, line 469: Intended?

alexaryn (commented):
I like it better than the last revision.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -168,6 +217,171 @@ def partition_pdf(
...
+        threshold,
+        use_ocr,
+        ocr_images,
+        ocr_tables,
+        extract_table_structure,
+        table_structure_extractor,
+        extract_images,
+    ) -> Any:
+        with LogTime("infer"):
+            deformable_layout = self.model.infer(batch, threshold, None)
```
bohou-aryn, line 354: How does this differ from the sequential processing? that batchify does not work?

lib/sycamore/sycamore/transforms/detr_partitioner.py
bohou-aryn: This happened in the worker process, right? Is there an easy to start it without modify worker level the transformation code?
Besides, this one captures single process or the OS level usage? does it capture underlying c++ lib running in same process?

eric-anderson: tracemalloc is a python thing https://docs.python.org/3/library/tracemalloc.html. It works there, it does not capture what the C++ libraries are doing.  Neither do the equivalent torch things. I don't know where the RSS is coming from. There are ways to run things at startup of a ray worker. see right above LogTime in time_trace.py for the magic.

lib/sycamore/sycamore/utils/deep_eq.py
eric-anderson: Done.

lib/sycamore/sycamore/utils/deep_eq.py
eric-anderson: Done.

lib/sycamore/sycamore/utils/deep_eq.py
eric-anderson: it prints out how the equality failed. Changed the message.

lib/sycamore/sycamore/utils/pdf.py
eric-anderson: Done.

lib/sycamore/sycamore/utils/pdf.py
eric-anderson: Done.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Maybe, but the only comparing bit I did was chained. I wanted: a .. b .. c .. d and printing d-c, c-b, b-a.  RAII would be harder in that case.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Maybe. I don't want to invest in more helper classes when it's not seeming that useful (the RSS isn't coming from python)

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Sure.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Done.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: Yes. Not having a blank line before the outdent is less readable to me.

lib/sycamore/sycamore/utils/pdf.py
eric-anderson: I believe they're auto-cleaned up when the thread exits (which has to have happened for the loop to finish), but I added the explicit joins.

lib/sycamore/sycamore/transforms/detr_partitioner.py
eric-anderson: I don't understand the question. The net effect of sequential and batched processing should be the same, but how they get there is different.

The sequential processing does:
1. pdf->list[images]
2. list[images]->RGB list[images]
3. for all images, infer(image batch)
4. run ocr or pdfminer and join
5. for each page, extract table structure and images

The batched processing does:
1. Start pdfminer if needed
2. Start incremental pdf -> images conversion
3. For each image batch:
4. .... run infer
5. .... extract ocr if requested
6. .... extract table structure if requested
7. .... extract images if requested
8. If using pdfminer, wait for completion and join

_process_batch is that inner loop (steps 4-7)

bohou-aryn: The sequence process also do batch infer. So the batching here refers to, table/image extraction and orc are batched?

alexaryn (commented):
I have nothing more to add.

PR #479, opened 2024-06-19
Disable Ray progress bar by default.

bsowell:
Currently the Ray progress bar doesn't update correctly due to this issue: https://github.com/ray-project/ray/issues/44983. Until that is fixed, we disable the progress bar by default. You can explicitly enable by passing enable_progress_bars=True in the sycamore.init function.

### DIFF
--- a/lib/sycamore/sycamore/context.py
+++ b/lib/sycamore/sycamore/context.py
@@ -3,12 +3,13 @@ import threading
 from typing import Any, Optional
 
 import ray
+import ray.data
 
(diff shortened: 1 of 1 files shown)


alexaryn (approved):
LGTM

scottjlee:
BTW, we just fixed https://github.com/ray-project/ray/issues/44983, which should be included in the next release. Please feel free to followup in the original issue if you continue to see issues. Thanks!

PR #480, opened 2024-06-20
Add DuckDB Writer

karansampath:
Adds DuckDB writer with the ability to store data in memory, leveraging the lightweight aspects of DuckDB. This builds off the Abstract Base Writer Class and uses similar test cases as done for previous writers to OpenSearch, Weviate and Pinecone.

This implementation does not support in memory databases due to the requirement to stay within the Ray Context.


lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,98 @@
...
+        dict_params = asdict(client_params)
+        db_name = dict_params["db_name"] if "db_name" in dict_params else ":memory:"
+        self._client = duckdb.connect(db_name)
+
+    @classmethod
+    def from_client_params(cls, params: BaseDBWriter.ClientParams) -> "DuckDBClient":
+        assert isinstance(params, DuckDBClientParams)
+        return DuckDBClient(params)
+
+    def write_many_records(self, records: list[BaseDBWriter.Record], target_params: BaseDBWriter.TargetParams):
```
bsowell, line 39: How does the process model work here? Sycamore/Ray will spin up multiple processes, each of which can write out records. What will happen in this case -- will there be a separate duckdb instance in each process?

bsowell: I don't know what the right answer is here, but we should figure out what we want -- is it all processes writing to a single duckdb instance in one process?

bsowell (commented):
Code is looking good. Biggest question is around the multi-process stuff.

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,98 @@
...
+            raise ValueError(f"Cannot write documents without a doc_id. Found {document}")
+        embedding = document.embedding
+        if embedding is None:
+            raise ValueError(f"Cannot write documents without a embedding. Found {document}")
+        properties = {
+            "properties": document.properties,
+            "type": document.type,
+            "text_representation": document.text_representation,
+            "bbox": document.bbox.coordinates if document.bbox else None,
+            "shingles": document.shingles,
```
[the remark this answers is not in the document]
karansampath: Makes sense, have made the change

examples/simple_duck.py
```diff
@@ -0,0 +1,34 @@
...
+    .regex_replace(COALESCE_WHITESPACE)
+    .mark_bbox_preset(tokenizer=tokenizer)
+    .merge(merger=MarkedMerger())
+    .spread_properties(["path"])
+    .split_elements(tokenizer=tokenizer, max_tokens=512)
+    .explode()
+    .embed(embedder=SentenceTransformerEmbedder(model_name=model_name, batch_size=100))
+    .sketch(window=17)
+)
+ds.write.duck()
```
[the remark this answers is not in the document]
karansampath: If we don't specify a location the data would go away after that process is killed, but even then we could print simple outputs that read from it before that? Otherwise we'd need to specify a location to persist it, from which we could directly open the database and show customers (don't necessarily need to write another reader for it)

lib/sycamore/sycamore/tests/integration/test_pdf_to_duck.py
```diff
@@ -0,0 +1,33 @@
...
+        .regex_replace(COALESCE_WHITESPACE)
+        .mark_bbox_preset(tokenizer=tokenizer)
+        .merge(merger=MarkedMerger())
+        .spread_properties(["path"])
+        .split_elements(tokenizer=tokenizer, max_tokens=512)
+        .explode()
+        .embed(embedder=SentenceTransformerEmbedder(model_name=model_name, batch_size=100))
+        .sketch(window=17)
+    )
+    ds.write.duck()
```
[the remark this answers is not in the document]
karansampath: Good idea, added a test for this

lib/sycamore/sycamore/writers/duckdb_writer.py
karansampath: All processes now write to independent csv files that we then bulk load into a DuckDB database helping us get a lot of the benefits from multiprocessing as discussed

lib/sycamore/sycamore/writer.py
```diff
@@ -342,12 +344,15 @@ def pinecone(
         )
         pc.execute()
 
-    def duck(self, db_name=None, table_name="aryn_table", **kwargs):
+    def duckdb(self, db_url=None, table_name=None, csv_directory_location=None, **kwargs):
```
HenryL27: nit: add type hints

lib/sycamore/sycamore/writer.py
```diff
@@ -367,17 +372,38 @@ def duck(self, db_name=None, table_name="aryn_table", **kwargs):
...
-
+        csv_location = csv_directory_location if csv_directory_location is not None else "tmp/duckdb"
+        table_name = table_name if not None else "data"
+        client_params = DuckDBClientParams()
+        target_params = DuckDBTargetParams(parquet_location=csv_location)
         ddb = DuckDBDocumentWriter(
             self.plan, client_params=client_params, target_params=target_params, name="duck_write_documents", **kwargs
         )
-        ddb.execute()
+        ddb.execute().materialize()
```
HenryL27, line 391: add a `execute=False` option, as discussed. This may require turning the CSV import thingy into a transform of its own or something

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -73,18 +119,17 @@ def from_doc(cls, document: Document, target_params: BaseDBWriter.TargetParams)
...
-            "type": document.type,
-            "text_representation": document.text_representation,
-            "bbox": document.bbox.coordinates if document.bbox else None,
-            "shingles": document.shingles,
-        }
-        droperties = drop_types(properties, drop_empty_lists=True)
-        assert isinstance(droperties, dict)
-        return DuckDBDocumentRecord(uuid=uuid, properties=droperties, embeddings=embedding)
+        # if embedding is None:
+        #     raise ValueError(f"Cannot write documents without a embedding. Found {document}")
```
HenryL27, line 123: nit: uncomment or delete

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -1,70 +1,116 @@
...
+        headers = ["uuid", "embeddings", "properties", "text_representation", "bbox", "shingles", "type"]
+
+        def write_batch(batch_data: dict):
+            # If the file doesn't exist, write headers first
+            file_exists = os.path.isfile(file_location)
+
+            with open(file_location, "a", newline="") as csvfile:
+                writer = csv.DictWriter(csvfile, fieldnames=headers)
+                if not file_exists:
+                    writer.writeheader()
```
HenryL27, line 71: potential race condition in this logic. maybe drop headers altogether and assume them in the DB table creation phase?

lib/sycamore/sycamore/writer.py
```diff
@@ -384,6 +384,10 @@ def duckdb(self, db_url=None, table_name=None, csv_directory_location=None, **kw
         ddb = DuckDBDocumentWriter(
             self.plan, client_params=client_params, target_params=target_params, name="duck_write_documents", **kwargs
         )
+        if not execute:
+            from sycamore.docset import DocSet
+
+            return DocSet(self.context, ddb)
```
HenryL27, line 390: this would need a phase where it imports into the table right?

karansampath: Yep, included one as discussed

lib/sycamore/sycamore/writer.py
```diff
@@ -342,6 +342,84 @@ def pinecone(
...
+            db_url=db_url,
+            name="duck_write_documents",
+            **kwargs,
+        )
+        if execute:
+            ddb_csv.execute().materialize()
+            ddb_writer.write_docs([])
+            return None
+        else:
+            raise ValueError("Not Implemented")
```
HenryL27, line 421: Use [NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError) - with a message like "Cannot use delayed execution with DuckDB" or something

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,50 @@
...
+
+from sycamore.data.document import Document
+from sycamore.plan_nodes import Node, Write
+from sycamore.transforms.map import MapBatch
+import glob
+import duckdb
+import os
+
+
+class DuckDB_Writer(MapBatch, Write):
```
HenryL27, line 11: nit: DuckDB_Writer -> DuckDBWriter

HenryL27 (approved):
Approving, although I'll request a slight refactor given where we've ended up:
First - merge duckdb_writer.py and duckdb_csv_writer.py into one file
Second (optional) - make DuckDB_Writer.write_docs just a plain function (with meaningful args rather than list[doc] -> list[doc]). Once we figure out the laziness story better we'll wrap it in a transform.

Sorry, I realize I'm basically asking you to undo all the stuff I said to do yesterday :(

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @classmethod
+    def from_client_params(cls, params: BaseDBWriter.ClientParams) -> "DuckDBClient":
+        assert isinstance(params, DuckDBClientParams)
+        return DuckDBClient(params)
+
+    def write_many_records(
+        self, records: list[BaseDBWriter.Record], target_params: BaseDBWriter.TargetParams, batch_size: int = 1000
+    ):
+        N = 1000 ^ 2  # Bit lesser than 1 MB
```
HenryL27, line 44: 1000 ^ 2 = 1002 apparently. You want 1000 ** 2. (Or if you're ok with exactly 1 MB 2 ** 20)... or just write out the number (python also accepts _'s in numbers so you could say 1_000_000)

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,145 @@
...
+        dict_params = (
+            asdict(target_params)
+            if is_dataclass(target_params)
+            else (target_params.__dict__ if hasattr(target_params, "__dict__") else {})
+        )
```
HenryL27, line 49: I assume this is for mypy? You should be able to do something like:
```python
assert isinstance(target_params, DuckDBTargetParams), f"Wrong kind of target parameters found: {target_params}"
dict_params = asdict(target_params)
```
I've ended up having to do this kind of [type narrowing](https://mypy.readthedocs.io/en/latest/type_narrowing.html) all over the place, which is painful but I think the cleanest option.

Oh you have the assert two lines later

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,145 @@
...
+        #     ]
+        # )
+
+        def write_batch(batch_data: dict):
+            pa_table = pa.Table.from_pydict(batch_data)  # noqa
+            client = duckdb.connect(str(dict_params.get("db_url")))
+            nonlocal creation
+            if creation:
+                client.sql(f"CREATE TABLE {dict_params.get('table_name')} AS SELECT * FROM pa_table")
+                creation = False
```
HenryL27, line 73: This logic should go into `create_target_idempotent`

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,145 @@
...
+            batch_data["text_representation"].append(r.text_representation)
+            batch_data["bbox"].append(r.bbox)
+            batch_data["shingles"].append(r.shingles)
+            batch_data["type"].append(r.type)
+
+            # If we've reached the batch size, write to the database
+            if batch_data.__sizeof__() >= N:
+                batch_data["properties"] = (
+                    [convert_to_str_dict(i) for i in batch_data["properties"]] if batch_data["properties"] else []
+                )
```
HenryL27, line 95: Move this to the initial append in case of weird size inconsistencies

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,145 @@
...
+        if len(batch_data["uuid"]) > 0:
+            write_batch(batch_data)
+
+    def create_target_idempotent(self, target_params: BaseDBWriter.TargetParams):
+        assert isinstance(target_params, DuckDBTargetParams)
+
+    def get_existing_target_params(self, target_params: BaseDBWriter.TargetParams) -> "DuckDBTargetParams":
+        assert isinstance(target_params, DuckDBTargetParams)
+        return DuckDBTargetParams(db_url=target_params.db_url, table_name=target_params.table_name)
+
```
HenryL27, line 107: see above: create_target_independent should _try_ to create the table, and if it fails, just don't throw an error.
get_existing_target_params should check to see if the table exists, and maybe also get the schema or something so you can compare them with TargetParams.compatible_with.

e.g. say I manually create a table with the same name as the table I'm ingesting into but different columns. create_target_idempotent should fail silently, and then get_existing_target_params should find the created table and its columns and report that. Then compatible_with should say "nope, not compatible" and baseWriter will raise an exception

lib/sycamore/sycamore/writers/duckdb_writer.py
```diff
@@ -0,0 +1,145 @@
...
+
+@dataclass
+class DuckDBDocumentRecord(BaseDBWriter.Record):
+    uuid: str
+    embeddings: Optional[list[float]] = None
+    properties: Optional[dict[str, Any]] = None
```
HenryL27, line 113: maybe make this the dict[str, str] and do your conversion here

karansampath: Discussed, would be better doing this later

lib/sycamore/sycamore/writers/duckdb_writer.py
karansampath: Makes sense, thanks for this

PR #482, opened 2024-06-20
Fix typo.

alexaryn:


eric-anderson (approved):
Please add a test that uses time_trace so things like this can't happen again.

PR #484, opened 2024-06-21
Get integration tests working again.

eric-anderson:
Important fixes:
  * Add a ludicrous amount of additional logging so it's rare that something isn't logged after a short wait.  Logs go to error because I was having some issues with info logging not showing up.  That may be fixed now, but unclear.
  * rm -rf /app/.scrapy/imported /app/.scrapy/downloads instead of *; * isn't being expanded
  * Fix imported files via find in the filesystem instead of parsing output. The ray output and importer output are sometimes interleaved on the same line
  * Verify that the scraper downloaded something before continuing
  * Tolerate missing indices on deletion
  * Enable optionally waiting before aborting the test so state is preserved for inspection
  * Add more debugging on the opensearch side, although bugs appeared to all be test-side.
  * Update automation to take advantage of debugging

### DIFF
--- a/apps/integration/integration/automation/runtests.sh
+++ b/apps/integration/integration/automation/runtests.sh
@@ -43,10 +43,21 @@ handle_outputs() {
 runtests() {
   docker system prune -f --volumes
   docker compose up reset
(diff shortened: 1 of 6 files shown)


alexaryn (approved):
Nothing terrible jumped out at me.

PR #486, opened 2024-06-24
pinecone demo nb

HenryL27:
demo notebook writing to pinecone and making a couple queries


jonfritz (changes requested):
1/ Let's use example queries that aren't so violent.

2/ Can we add more comments to cells to instruct what we are doing, to guide the customer along?

jonfritz (approved):
LGTM! Thank you!

PR #489, opened 2024-06-25
Add PDFMiner caching

pparmar30:


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -492,33 +508,44 @@ def _convert_bbox_coordinates(
...
-                        text.type = "text"
-                        text.bbox = BoundingBox(x1 / width, y1 / height, x2 / width, y2 / height)
-                        text.text_representation = obj.get_text()
-                        if text.text_representation:
-                            texts.append(text)
-
-                pages.append(texts)
-            return pages
+        cache_dir = cache_location if cache_location else os.path.dirname(__file__)
+        cache_loc = cache_dir + "/PDFMiner_Cache"
```
austintlee, line 515: Shouldn't this be done _once_ when you initialize sycamore?  It seems a bit awkward that you specify the cache location on every invocation of the method (although I assume in most cases, the default will be chosen).  Actually, maybe we shouldn't allow the caller to be able to specify/change the cache location at all.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -492,33 +508,44 @@ def _convert_bbox_coordinates(
...
+                            text.type = "text"
+                            text.bbox = BoundingBox(x1 / width, y1 / height, x2 / width, y2 / height)
+                            text.text_representation = obj.get_text()
+                            if text.text_representation:
+                                texts.append(text)
+
+                    pages.append(texts)
+                if cache_results:
+                    logging.info("Cache Miss for PDFMiner. Storing the result to the cache.")
+                    CacheManager.cache_result(pages, file_loc)
```
austintlee, line 543: I thought the flag we wanted was for controlling whether or not the request should be served from the cache.  In other words, you want the caller to choose whether or not the result should be computed anew.  This is useful for perf testing.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -469,29 +477,39 @@ def _convert_bbox_coordinates(
...
-                        if text.text_representation:
-                            texts.append(text)
-
-                pages.append(texts)
-            return pages
+
+        hash_key = CacheManager.get_hash_key(filename)
+        cached_result = CacheManager.get(hash_key) if use_cache else None
+        if cached_result:
+            logging.info("Cache Hit for PDFMiner. Getting the result from cache.")
```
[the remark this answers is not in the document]
pparmar30: the hash_key wont be meaningful and file_name is coming from a temp file so probably not useful either

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -446,6 +454,7 @@ def __init__(self):
         param = LAParams()
         self.device = PDFPageAggregator(rm, laparams=param)
         self.interpreter = PDFPageInterpreter(rm, self.device)
+        self.cache_manager = CacheManager("/tmp/SycamoreCache/PDFMinerCache")
```
austintlee, line 457: I don't think you want to instantiate an object of CacheManager here.  Instead, you want an instance passed in.  I am looking at how PDFMinerExtractor is used and it looks like it gets instantiated every time `_run_pdfminer` is invoked!

lib/sycamore/sycamore/utils/cache_manager.py
```diff
@@ -0,0 +1,26 @@
...
+
+class CacheManager:
+    def __init__(self, cache_loc: str):
+        self._cache = Cache(directory=cache_loc)
+
+    def get(self, hash_key: str):
+        return self._cache.get(hash_key)
+
+    def set(self, hash_key: str, hash_value):
+        self._cache.set(hash_key, hash_value)
```
austintlee, line 16: Just to be clear, we are not storing hashed values, but rather actual values?

pparmar30: thats right

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -469,29 +480,39 @@ def _convert_bbox_coordinates(
...
-                        text.type = "text"
-                        text.bbox = BoundingBox(x1 / width, y1 / height, x2 / width, y2 / height)
-                        text.text_representation = obj.get_text()
-                        if text.text_representation:
-                            texts.append(text)
-
-                pages.append(texts)
-            return pages
+
+        hash_key = pdf_miner_cm.get_hash_key(filename)
```
austintlee, line 487: Let's add a TODO to see if we can share the file content used here to compute the hash with the PDF miner routine happening below in the event of a cache miss.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -35,6 +36,9 @@ def _batchify(iterable, n=1):
         yield iterable[i : min(i + n, length)]
 
 
+pdf_miner_cm = CacheManager("/tmp/SycamoreCache/PDFMinerCache")
```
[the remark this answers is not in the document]
pparmar30: Updated it to take in system tempdir

austintlee (approved):

PR #490, opened 2024-06-26
Weaviate Scan

HenryL27:
adds a weaviate collection scan. e.g.
```python
context.read.weaviate(wv_client_args, collection_name).do_stuff()
```
Also refactors some existing opensearch and weaviate stuff around to sycamore/connectors/


lib/sycamore/sycamore/connectors/weaviate/weaviate_scan.py
```diff
@@ -0,0 +1,26 @@
...
+        super().__init__(**kwargs)
+        self._collection_name = collection_name
+        self._connection_params = connection_params
+
+    def execute(self) -> Dataset:
+        documents = []
+        with WeaviateClient(**asdict(self._connection_params)) as wcl:
+            collection = wcl.collections.get(self._collection_name)
+            for object in collection.iterator(include_vector=True):
+                doc = Document(object.vector | object.properties | {"doc_id": str(object.uuid)})  # type: ignore
```
karansampath, line 20: Wouldn't you want to leave the dictionary from `object.properties` as is and not merge its parts? To ensure it matches the input version

lib/sycamore/sycamore/tests/integration/scans/test_weaviate_scan.py
```diff
@@ -0,0 +1,98 @@
...
+        .sketch(window=17)
+        .take_all()
+    )
+    ctx.read.document(docs).write.weaviate(
+        wv_client_args=wv_client_args, collection_name=collection, collection_config=collection_config_params
+    )
+
+    out_docs = ctx.read.weaviate(wv_client_args=wv_client_args, collection_name=collection).take_all()
+    assert len(out_docs) == len(docs)
+    assert {d.doc_id for d in docs} == {d.doc_id for d in out_docs}
```
karansampath, line 98: Would suggest testing for full equality here, as done below: https://github.com/aryn-ai/sycamore/blob/d51348f6ebfcdbc487804057e1a572f208694b10/lib/sycamore/sycamore/tests/integration/scans/test_db_scan.py#L96

lib/sycamore/sycamore/connectors/weaviate/weaviate_scan.py
```diff
@@ -0,0 +1,26 @@
...
+
+class WeaviateScan(Scan):
+    def __init__(self, collection_name: str, connection_params: WeaviateClientParams, **kwargs):
+        super().__init__(**kwargs)
+        self._collection_name = collection_name
+        self._connection_params = connection_params
+
+    def execute(self) -> Dataset:
+        documents = []
+        with WeaviateClient(**asdict(self._connection_params)) as wcl:
```
karansampath, line 17: Similar to the test case, would suggest checking whether the client starts in <timeout seconds

karansampath (commented):
Small comments and edits

lib/sycamore/sycamore/connectors/weaviate/weaviate_scan.py
HenryL27: object.properties is {"text_representation": blah, "type": blah, "properties": nested blah, etc}. We need to tack on the embedding vector and doc_id bc those aren't indexed as weaviate 'properties'

lib/sycamore/sycamore/connectors/weaviate/weaviate_scan.py
HenryL27: The test case does that bc it's using embedded weaviate - so it has to start it up and potentially download a binary etc. 
I'm pretty sure I can assume I have a running weaviate instance when I try to read from it. Also you can add a timeout to the connection params, like [here](https://github.com/aryn-ai/sycamore/blob/bd05d08086711ac54306c8777f13e12161c0fca6/examples/simple_weaviate.py#L36)

karansampath: Makes sense, discussed offline as well

lib/sycamore/sycamore/connectors/weaviate/weaviate_scan.py
karansampath: Ah I see, thanks for clarifying

karansampath (approved):

PR #493, opened 2024-06-27
Add filter_elements method on DocSet.

bsowell:
This is just a convenience method that behaves like map_elements.

### DIFF
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -735,7 +735,7 @@ class DocSet:
         """
 
         def process_doc(doc: Document) -> Document:
-            new_elements = filter(f, doc.elements)
+            new_elements = [e for e in doc.elements if f(e)]
             doc.elements = new_elements
             return doc
 
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -726,6 +726,21 @@ class DocSet:
         filtered = Filter(self.plan, f=f, **resource_args)
         return DocSet(self.context, filtered)
 
+    def filter_elements(self, f: Callable[[Element], bool], **resource_args) -> "DocSet":
+        """
+        Applies the given filter function to each element in each Document in this DocsSet.
+
+        Args:
+            f: A Callable that takes an Element and returns True if the element should be retained.
+        """
+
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/docset.py
```diff
@@ -726,6 +726,21 @@ def custom_filter(doc: Document) -> bool:
...
+        """
+        Applies the given filter function to each element in each Document in this DocsSet.
+
+        Args:
+            f: A Callable that takes an Element and returns True if the element should be retained.
+        """
+
+        def process_doc(doc: Document) -> Document:
+            new_elements = filter(f, doc.elements)
+            doc.elements = new_elements
```
HenryL27, line 739: I've had some issues using `filter` with it giving me a 'filter' object which is an iterable but not a list (lazily evaluated)... mostly typing and serializability things. Might be bc I was trying to chain a bunch of them together, idk. Anyway, would suggest wrapping in `list( . )` or just writing out the comprehension

lib/sycamore/sycamore/tests/unit/test_docset.py
```diff
@@ -199,3 +199,22 @@ def test_map_elements(self):
...
+            doc.elements = [
+                Element(text_representation=f"Document {i} Element {j}", properties={"element_val": j})
+                for j in range(10)
+            ]
+
+        context = sycamore.init()
+        docset = context.read.document(docs).filter_elements(lambda e: e.properties["element_val"] % 2 == 0)
+
+        all_docs = docset.take_all()
+        assert len(all_docs) == len(docs) / 2
```
HenryL27, line 216: wait there was no explode. shouldn't these be the same length? (but each doc only has 5 elts?)

HenryL27: oh it's bc you never appended the constructed `doc`s to `docs` so 0 = 0/2

bsowell: Oh oops. This means the test I copied this from was wrong too.

lib/sycamore/sycamore/docset.py
bsowell: It's interesting that mypy didn't catch this, but I'll switch to a comprehension.

HenryL27 (approved):

PR #494, opened 2024-06-27
Support table deserialization from a dictionary.

bsowell:
This is necessary to deserialize a Document containing a table from the ArynPartitioner.


MarkLindblad (approved):
Looks good to me.

PR #495, opened 2024-06-28
Add flatten option to weaviate writer

HenryL27:
Weaviate nested fields don't do filters or aggregations so we add the option to flatten them to top-level 'properties'.
An 'unflatten' option should be added to the scan, prob when the pinecone scan is written.

### DIFF
--- a/lib/sycamore/sycamore/connectors/weaviate/weaviate_writer.py
+++ b/lib/sycamore/sycamore/connectors/weaviate/weaviate_writer.py
@@ -40,6 +40,7 @@ CollectionConfigCreate: TypeAlias = _CollectionConfigCreate
 class WeaviateTargetParams(BaseDBWriter.TargetParams):
     name: str
     collection_config: Union[CollectionConfigCreate, CollectionConfig]
+    flatten_properties: bool = False
 
     def compatible_with(self, other: BaseDBWriter.TargetParams) -> bool:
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/connectors/weaviate/weaviate_writer.py
```diff
@@ -189,6 +192,10 @@ def from_doc(cls, document: Document, target_params: BaseDBWriter.TargetParams)
             "shingles": document.shingles,
         }
         droperties = drop_types(properties, drop_empty_lists=True)
+        if target_params.flatten_properties:
+            # Property names must be [a-zA-Z][_0-9a-zA-Z]{0,230}, so use __ as a separator rather than .
+            droperties = dict(flatten_data(droperties, allowed_list_types=[int, str, bool, float], separator="__"))
```
karansampath, line 197: Any reason we don't allow for `bytes` or `complex` data types here?

karansampath (approved):
LGTM, made a small suggestion

lib/sycamore/sycamore/connectors/weaviate/weaviate_writer.py
HenryL27: not supported by weaviate: https://weaviate.io/developers/weaviate/config-refs/datatypes

PR #496, opened 2024-06-28
Add DuckDB Demo Notebook

karansampath:
Adds a demo notebook for DuckDB and adds a few fixes to allow it to work. Note: since DuckDB's vector similarity search isn't production ready yet for persistent databases (https://duckdb.org/docs/extensions/vss.html), this notebook transfers the data over to an in-memory database. This can be easily changed later if / when DuckDB adds the capability for this.


lib/sycamore/sycamore/connectors/duckdb/duckdb_writer.py
```diff
@@ -23,7 +23,7 @@ class DuckDBTargetParams(BaseDBWriter.TargetParams):
     schema: Optional[Dict[str, str]] = field(
         default_factory=lambda: {
             "doc_id": "VARCHAR",
-            "embeddings": "DOUBLE[]",
+            "embeddings": "FLOAT[384]",
```
HenryL27, line 26: make this dimension number easy to set by the user independently of all else

lib/sycamore/sycamore/transforms/embed.py
```diff
@@ -98,7 +98,8 @@ def generate_embeddings(self, doc_batch: list[Document]) -> list[Document]:
         assert self._transformer is not None
 
         text_batch = [self.pre_process_document(doc) for doc in doc_batch if doc.text_representation is not None]
-        embeddings = self._transformer.encode(text_batch, batch_size=self.model_batch_size, device=self.device)
+        if len(text_batch) > 0:
+            embeddings = self._transformer.encode(text_batch, batch_size=self.model_batch_size, device=self.device)
```
HenryL27, line 102: slightly prefer 'if len(text_batch) == 0 then return early' just so I don't have to think through the case where `embeddings` is unbound

lib/sycamore/sycamore/connectors/duckdb/duckdb_writer.py
```diff
@@ -104,9 +107,10 @@ def create_target_idempotent(self, target_params: BaseDBWriter.TargetParams):
         client = duckdb.connect(str(dict_params.get("db_url")))
         try:
             if schema:
+                embedding_size = schema.get("embeddings").str.split("[")[0] + str(dict_params.get("dimensions")) + "]"
```
HenryL27, line 110: does this capture the [ or are you gonna end up with `FLOAT768]`?

lib/sycamore/sycamore/writer.py
```diff
@@ -379,6 +379,7 @@ def duckdb(
         db_url: Optional[str] = None,
         table_name: Optional[str] = None,
         execute: bool = True,
+        dimensions: int = 384,
```
HenryL27, line 382: Not sure I'd give this a default value but ok

lib/sycamore/sycamore/connectors/duckdb/duckdb_writer.py
karansampath: Changed it

jonfritz (changes requested):
Great stuff! Thx for driving this. 

A few comments:
1/It looks like the output of the notebook is being stored in the Git file. Isn't it more common to clear it before uploading?
2/In cell 9, I would write #Use DuckDB's semantic search for retrieval.
3/What is the purpose of cell 10 and how is it different than cell 9? Make it clear in the info text in cell 10.
4/Customers are going to care more about RAG as a next step after the first NN search in DuckDB - let's not get too weedsy. As discussed, for the example cells, let's just have: direct NN search (one way), RAG, and the other stuff you have as examples can go after RAG. We don't want to lose readers along the way.
5/For the Langchain example, instead of adding a column, should we just copy the table, change the name of the properties column to metadata, and add the source key:value there (or something like that)? Then, we could add to the RAG example (if we wanted to) with filters, etc.

HenryL27 (approved):
lgtm. I think @jonfritz still needs to stamp it since he requested changes

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "from sycamore.transforms.embed import SentenceTransformerEmbedder"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from sycamore.data.document import Document\n",
```
jonfritz, line 28: Add a comment here explaining what this this cell is. It's used for data cleaning and formatting later in the Sycamore script.

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "fsys = pyarrow.fs.S3FileSystem(region=\"us-east-1\", anonymous=True)\n",
+    "\n",
+    "llm = OpenAI(OpenAIModels.GPT_3_5_TURBO.value)\n",
+    "tokenizer = HuggingFaceTokenizer(\"thenlper/gte-small\")\n",
+    "\n",
+    "ctx = sycamore.init()\n",
+    "\n",
+    "ds = (\n",
+    "    ctx.read.binary(paths, binary_format=\"pdf\", filesystem=fsys)\n",
+    "    # Parition with the sycamore partitioner, pulling out tables and images. ActorPoolStrategy(size=3) works best on my particular hardware\n",
```
jonfritz, line 64: I wouldn't use the word "my" here. Also, where is this ActorPoolStrategy field in the example? If it's not there as default, why are we saying it?

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "    # but your mileage may vary depending on your RAM.\n",
+    "    .partition(partitioner=SycamorePartitioner(extract_table_structure=True, extract_images=True, use_cache=False), compute=ActorPoolStrategy(size=3))\n",
+    "    # Get rid of spurious whitespace charaters\n",
+    "    .regex_replace(COALESCE_WHITESPACE)\n",
+    "    # Automatically determine a schema of additional metadata to extract from documents\n",
```
jonfritz, line 69: Upper case D for documents in all of the helper text (not going to leave a comment for each one)

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Write to a persistent duckdb database\n",
```
jonfritz, line 97: DuckDB is capitalized like this. Pls make sure it's that way in all comments.

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "data_conn = duckdb.connect(\"demo.db\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# DuckDB is not production ready yet for efficient Vector Similarity Search on persistent databases (https://duckdb.org/docs/extensions/vss.html), \n",
```
jonfritz, line 126: Prod ready sounds weird, as per our prior convo. Would recommend: "DuckDB's efficient Vector Similarity Search on persistent databases is still an experimental feature (https://duckdb.org/docs/extensions/vss.html), so we load our data into an in-memory database. Once this feature is more stable, we can query the persistent database directly using VSS."

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "df.dropna(subset=[\"embeddings\"],inplace=True)\n",
+    "\n",
+    "\n",
+    "# Now, we load the data into an in-memory database. Notice that we specify beforehand since we need the embedding column (must be specified as a FLOAT[N] where N is specified) \n",
+    "# for Vector Similarity Search using HNSW to work.\n",
+    "in_memory_db = duckdb.connect(\":default:\")\n",
+    "in_memory_db.install_extension(\"vss\")\n",
+    "in_memory_db.load_extension(\"vss\")\n",
+    "\n",
+    "# To allow RAG to work properly on this database using Langchain (shown later), we must rename the properties column to metadata, \n",
```
jonfritz, line 140: Put quotes around the column names or something so people know it

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "# add a source key within it, and make it of type later. Note that the value 'default' can be modified to other options if needed\n",
```
jonfritz, line 141: Same with source, use quotes or some identifier

jonfritz: "make it of type" is confusing. rewrite it

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "in_memory_db.sql(f\"SELECT doc_id, text_representation, metadata FROM in_memory_table ORDER BY array_distance(embeddings, {embedder.embed_query(q)}::FLOAT[384]) LIMIT 10;\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Pretty print tables\n",
```
jonfritz, line 203: Should we combine this cell with the above one, so we have one cell that has a useful ANN?

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "print(tabulate(table_data, headers=headers, tablefmt=\"grid\"))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# View the first two rows of the text representation to visually confirm results \n",
```
jonfritz, line 225: Move to end?

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "    print(\"-\" * 50)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# We can also see the performance of the DuckDB query and check the index is being used by using the in-built visualization below \n",
```
jonfritz, line 242: Move to end?

notebooks/ntsb-duckdb.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "qa.invoke({\"question\": \"How many accidents happened?\"})"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Finally we can build an index with other distance metrics, the default is Euclidean Distance ('l2sq'), with Cosine similarity ('cosine', 'array_cosine_simarlity') \n",
```
jonfritz, line 299: We aren't really explaining why you'd wand to do this. I might phrase it as "To experiment with XXXXX, we can..."

jonfritz (changes requested):
Looking good! Left some feedback. Thx for driving this!

notebooks/ntsb-duckdb.ipynb
karansampath: Specifying it would help allow you to run queries at the maximum parallelism your machine can support. But I agree, this is additional not needed complexity. Have removed it now

notebooks/ntsb-duckdb.ipynb
karansampath: Discussed on chat, left it there since it helps users intuitively make sense of the result

jonfritz (approved):
LGTM! Awesome.

PR #497, opened 2024-06-28
Combine Aryn and Sycamore Partitioners

MarkLindblad:


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -101,34 +123,103 @@ def partition_pdf(
...
-                table_structure_extractor,
-                extract_images,
-                batch_size,
-                use_cache,
-            )
+            if batch_at_a_time:
+                temp = self._partition_pdf_batched(
+                    file,
+                    threshold,
+                    use_ocr,
```
baitsguy, line 148: nit for future: when passing a bunch of params (specially bools) it's safer to send named arguments, i.e. use_ocr=use_ocr instead of just passing use_ocr. Easy to reorder params accidentily and do the wrong thing

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -36,26 +41,43 @@ def _batchify(iterable, n=1):
...
+        return e.can_retry
+    else:
+        return False
+
+
 pdf_miner_cache = DiskCache(os.path.join(tempfile.gettempdir(), "SycamoreCache/PDFMinerCache"))
 
 
-class SycamorePDFPartitioner:
+class ArynPDFPartitioner:
```
baitsguy, line 65: I would think this belongs in the aryn_partitioner file as we'd want to abstract the detr implementation. But I guess its coupled currently so no harm

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -476,101 +488,48 @@ def partition(self, document: Document) -> Document:
...
-
-        .. code-block:: python
-
-            context = sycamore.init()
-            partitioner = ArynPartitioner(extract_table_structure=True, extract_images=True)
-            context.read.binary(paths, binary_format="pdf")\
-                .partition(partitioner=partitioner)
-    """
-
+class SycamorePartitioner(ArynPartitioner):
```
baitsguy, line 505: If this is for compatibility reasons (I think we have notebooks around) it's fine to keep this but let's document that. Delete otherwise

baitsguy (approved):
No blocking comments, some cleanup questions that can be a follow up

lib/sycamore/sycamore/transforms/detr_partitioner.py
HenryL27: aryn_partitioner.py gets deleted with this pr

PR #498, opened 2024-06-28
Write intermediate results in BaseMap

baitsguy:
_Note: I ran `poetry run pre-commit run --all-files` and it seems to have formatted some files which makes this look larger than it is._ 

Please review the following files:

- lib/sycamore/sycamore/tests/unit/transforms/test_base.py
- lib/sycamore/sycamore/transforms/base.py
- lib/sycamore/sycamore/writers/file_writer.py

The other changes are to the node.execute() interface to accept kwargs and auto-formatting


HenryL27:
why are we doing this?

baitsguy:
> why are we doing this?

Want to be able to inspect results of each step in the execution graph (to answer things like what data got filtered out). It's an alternative to writing a .write after each node in your script which can be messy.

HenryL27:
Any chance of using the basewriter interface (inserting maps into the dag)? Trying to abstract ray. Also is having a bunch of datasinks in the ray path going to force a whole bunch of re-execution?

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -109,20 +109,41 @@ def __init__(
...
         else:
-            return input_dataset.map_batches(self._map_function(), **self.resource_args)
+            result = input_dataset.map_batches(self._map_function(), **self.resource_args)
+
+        if write_intermediate_data:
+            assert intermediate_datasink is not None
+            if isinstance(intermediate_datasink, type):
+
+                # ensure each nodes data is written in a separate directory
+                intermediate_datasink_kwargs["path"] = intermediate_datasink_kwargs["path"] + "/" + self._name
```
HenryL27, line 141: this seems dangerous

HenryL27 (commented):
ok. I don't love bringing back ray.data.Datasink since I've mostly been working on eliminating that but I understand the reasoning

lib/sycamore/sycamore/transforms/base.py
baitsguy: Yeah will likely modify this as a follow up. It's a little annoying because the 'name' is only really accessible here and we want it in the path

HenryL27: I'm worried about key errors in the 'path' lookup, not the private name reference

baitsguy: It's a required property in the sinks I see

HenryL27 (approved):
lgtm

PR #500, opened 2024-06-28
Extract Aryn token from config for ArynPartitioner

MarkLindblad:
Check `$ARYN_API_KEY` and read `.aryn/config.yaml` in that order to find the Aryn token (Aryn API Key).

### DIFF
--- a/lib/sycamore/sycamore/utils/aryn_config.py
+++ b/lib/sycamore/sycamore/utils/aryn_config.py
@@ -1,0 +1,50 @@
+import os
+import pathlib
+import yaml
+import threading
+from typing import Optional
+import logging
+
+_DEFAULT_PATH = os.path.join(pathlib.Path.home(), ".aryn", "config.yaml")
+
+class ArynConfig:
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
+
+_global_aryn_config_lock = threading.Lock()
+_global_aryn_config: Optional[ArynConfig] = None
+
+def get_aryn_api_key(requested_yaml_config_path: str = None) -> str:
+    api_key = os.environ.get("ARYN_API_KEY")
+    if api_key:
+        return api_key
+    
+    global_aryn_config = get_aryn_config(requested_yaml_config_path)
+    return global_aryn_config.__dict__.get("aryn_token", "")
+
+
+def get_aryn_config(requested_yaml_config_path: str = None) -> ArynConfig:
+    config_path = requested_yaml_config_path or os.environ.get("ARYN_CONFIG") or _DEFAULT_PATH
+    global _global_aryn_config
+    with _global_aryn_config_lock:
+        if _global_aryn_config is None:
+            config = ArynConfig()
+            try:
+                with open(config_path, "r") as yaml_file:
+                    aryn_env = yaml.safe_load(yaml_file)
+                    if aryn_env is None or not isinstance(aryn_env, dict):
+                        logging.debug("Aryn YAML config appears to be empty.")
+                        return
+                    config.__dict__.update(aryn_env)
+                    logging.debug(f"Aryn configuration: {vars(config)}")
+            except FileNotFoundError as err:
+                if config_path == _DEFAULT_PATH:
+                    logging.debug(f"Aryn config YAML not present: {err}")
+                else:
+                    logging.error(f"Unable to load specified aryn config {config_path}: {err}")
+            except yaml.scanner.ScannerError as err:
+                logging.error(f"Unable to load {config_path}: {err}. Ignoring the config file.")
+            _global_aryn_config = config
+        
+        return _global_aryn_config
+    
--- a/lib/sycamore/sycamore/utils/aryn_config.py
+++ b/lib/sycamore/sycamore/utils/aryn_config.py
@@ -1,8 +1,6 @@
 import os
 import pathlib
 import yaml
-import threading
-from typing import Optional
 import logging
 
 _DEFAULT_PATH = os.path.join(pathlib.Path.home(), ".aryn", "config.yaml")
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -0,0 +1,41 @@
...
+
+_DEFAULT_PATH = os.path.join(pathlib.Path.home(), ".aryn", "config.yaml")
+
+
+class ArynConfig:
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
+
+
+def get_aryn_api_key(requested_yaml_config_path: str = "") -> str:
```
alexaryn, line 14: `requested_yaml_config_path` is a very long variable name; how about `config_path`?

lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -0,0 +1,41 @@
...
+    api_key = os.environ.get("ARYN_API_KEY")
+    if api_key:
+        return api_key
+
+    global_aryn_config = get_aryn_config(requested_yaml_config_path)
```
alexaryn, line 19: It looks like we parse the config file every time and throw the result away right after.

lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_aryn_config(requested_yaml_config_path: str = "") -> ArynConfig:
+    config_path = requested_yaml_config_path or os.environ.get("ARYN_CONFIG") or _DEFAULT_PATH
+    config = ArynConfig()
+    try:
+        with open(config_path, "r") as yaml_file:
+            aryn_env = yaml.safe_load(yaml_file)
+            if aryn_env is None or not isinstance(aryn_env, dict):
+                logging.debug("Aryn YAML config appears to be empty.")
```
alexaryn, line 30: This is probably not "debug" level, but "warning" level.

lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -0,0 +1,41 @@
+import os
+import pathlib
+import yaml
+import logging
+
+_DEFAULT_PATH = os.path.join(pathlib.Path.home(), ".aryn", "config.yaml")
+
+
+class ArynConfig:
```
alexaryn, line 9: I think what we really want here is a a singleton class.  It should probably read the config file once and keep the result.  I'd look into using class-variables for this.

lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
```
alexaryn, line 11: Where do we use this cleverness?

I'd worry about the namespace overlap.  What if someone passes a kwarg called 'get_aryn_api_key'?

alexaryn (commented):
Looks like a good idea, but I think it could use some refinement.

lib/sycamore/sycamore/utils/aryn_config.py
MarkLindblad: Removed.

lib/sycamore/sycamore/utils/aryn_config.py
MarkLindblad: Changed.

lib/sycamore/sycamore/utils/aryn_config.py
MarkLindblad: Changed.

lib/sycamore/sycamore/utils/aryn_config.py
MarkLindblad: Fixed.

lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -2,40 +2,52 @@
...
+        with cls._global_aryn_config_lock:
+            if cls._global_aryn_config:
+                return
+            config_path = config_path or os.environ.get("ARYN_CONFIG") or _DEFAULT_PATH
+
+            try:
+                with open(config_path, "r") as yaml_file:
+                    aryn_config = yaml.safe_load(yaml_file)
+                    cls._global_aryn_config = aryn_config
+                    if aryn_config is None or not isinstance(aryn_config, dict):
```
alexaryn, line 43: None is NoneType so you can just use the isinstance().

lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -2,40 +2,52 @@
...
+                        logging.warning("Aryn YAML config appears to be empty.")
+                        return
+                    logging.debug(f"Aryn configuration: {aryn_config}")
+            except FileNotFoundError as err:
+                if config_path == _DEFAULT_PATH:
+                    logging.debug(f"Aryn config YAML not present: {err}")
+                else:
+                    logging.error(f"Unable to load specified aryn config {config_path}: {err}")
```
alexaryn, line 51: I would get rid of the enclosing if-statement and just have this message, but drop the word "specified".

lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -2,40 +2,52 @@
...
+            except yaml.scanner.ScannerError as err:
+                logging.error(f"Unable to load {config_path}: {err}. Ignoring the config file.")
```
alexaryn, line 53: "Unable to parse" would be more accurate/useful.  It's debatable that it's valid to continue in this case.

alexaryn (approved):
Much better.  Just some nits I trust you to address.

lib/sycamore/sycamore/utils/aryn_config.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/aryn_config.py
MarkLindblad: Changed message.

lib/sycamore/sycamore/utils/aryn_config.py
MarkLindblad: Done

PR #501, opened 2024-06-28
Add PDF to Image cache

pparmar30:


lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -108,14 +113,45 @@ def read_stderr(fh, q):
         t_err.join()
 
 
-def convert_from_path_streamed_batched(filename: str, batch_size: int) -> Generator[List[Image.Image], None, None]:
+def convert_from_path_streamed_batched(
+    filename: str, batch_size: int, hash_key: str, use_cache: bool = True
```
austintlee, line 117: I think as a function argument, a better name would be 'file_checksum' instead of 'hash_key'.  It just leaks internals.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -108,14 +113,45 @@ def read_stderr(fh, q):
...
+) -> Generator[List[Image.Image], None, None]:
     """Note: model service will call this to get batches of images for processing"""
+    images = pdf_to_ppm_cache.get(hash_key) if use_cache else None
+    if images:
+        print(f"PDFToPPM Cache Hit. Getting the results from cache.")
```
austintlee, line 122: logging

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -108,14 +113,45 @@ def read_stderr(fh, q):
...
+        yield from yield_batches(images, batch_size)
+    else:
+        print(f"PDFToPPM Cache Miss. Getting the results.")
```
austintlee, line 125: logging

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -108,14 +113,45 @@ def read_stderr(fh, q):
...
     for i in convert_from_path_streamed(filename):
         batch.append(i)
         if len(batch) == batch_size:
             yield batch
+            all_images.extend(batch)
             batch = []
 
     if len(batch) > 0:
+        all_images.extend(batch)
         yield batch
```
austintlee, line 154: Is there a difference between calling all_images.extend() before or after yield.  You have them in different orders.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -108,14 +113,45 @@ def read_stderr(fh, q):
...
+
+    if use_cache:
+        pdf_to_ppm_cache.set(hash_key, all_images)
```
austintlee, line 157: One thing to check is whether we need this flag to cache.  This flag is mainly for skipping cache.  I think we decided we will always cache the results once computed.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -108,14 +113,45 @@ def read_stderr(fh, q):
...
+    for image in images:
+        batch.append(image)
+        if len(batch) == batch_size:
+            yield batch
+            batch = []
+    if len(batch) > 0:
+        yield batch
+
+
+def load_images(
```
austintlee, line 140: How about 'yield_cached_batches' and 'yield_batches'.  I think 'load' is a bit confusing?

lib/sycamore/sycamore/utils/pdf.py
pparmar30: it's the same

lib/sycamore/sycamore/utils/pdf.py
pparmar30: updated

lib/sycamore/sycamore/utils/pdf.py
pparmar30: updated

lib/sycamore/sycamore/utils/pdf.py
pparmar30: updated

lib/sycamore/sycamore/utils/pdf.py
pparmar30: updated

lib/sycamore/sycamore/utils/pdf.py
pparmar30: updated

austintlee (approved):

PR #503, opened 2024-07-02
Add S3 cache implementation, add cache option to Llm

baitsguy:
1. Add a basic S3 cache implementation
2. Allow caching for exact LLM calls

### DIFF
--- a/lib/sycamore/sycamore/llms/openai.py
+++ b/lib/sycamore/sycamore/llms/openai.py
@@ -1,3 +1,4 @@
+import json
 from dataclasses import dataclass
 from enum import Enum
 import logging
@@ -14,6 +15,8 @@ from openai.lib.azure import AzureADTokenProvider
 from openai.types.chat import ChatCompletionMessageParam
 from sycamore.llms.llms import LLM
 from sycamore.llms.prompts import GuidancePrompt
+from sycamore.utils.cache_manager import CacheManager
+import hashlib
 
 logger = logging.getLogger(__name__)
 
@@ -210,6 +213,7 @@ class OpenAI(LLM):
         api_key=None,
         client_wrapper: Optional[OpenAIClientWrapper] = None,
         params: Optional[OpenAIClientParameters] = None,
+        cache: Optional[CacheManager] = None,
         **kwargs,
     ):
         if isinstance(model_name, OpenAIModels):
@@ -225,8 +229,7 @@ class OpenAI(LLM):
             logger.warn("text-davinci-003 is deprecated. Falling back to gpt-3.5-turbo-instruct")
             self.model = OpenAIModels.GPT_3_5_TURBO_INSTRUCT.value
 
-        self.model_name = self.model.name
-        super().__init__(self.model_name)
+        super().__init__(self.model.name, cache)
 
         # This is somewhat complex to provide a degree of backward compatibility.
         if client_wrapper is None:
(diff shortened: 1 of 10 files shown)


lib/sycamore/sycamore/utils/cache_manager.py
```diff
@@ -1,11 +1,73 @@
...
+import time
+from typing import Any
 
-from diskcache import Cache
+import diskcache
+from botocore.exceptions import ClientError
 
 
-class CacheManager:
+class Cache:
```
pparmar30, line 10: Should we instead to CacheManager --> S3CacheManager(CacheManager) and DiskCacheManager(CacheManager) ?

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -251,18 +254,45 @@ def __reduce__(self):
...
+        return deserializer, (kwargs,)
 
     def is_chat_mode(self):
         return self.model.is_chat
 
+    @staticmethod
+    def _get_cache_key(prompt_kwargs: dict, llm_kwargs: Optional[dict] = None):
+        combined = {"prompt_kwargs": prompt_kwargs, "llm_kwargs": llm_kwargs}
+        json_str = json.dumps(combined, sort_keys=True)
+        hash_obj = hashlib.sha256(json_str.encode("utf-8"))
```
pparmar30, line 268: This is most likely going to be a common operation! We can use the method from CacheManager.

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -251,18 +254,45 @@ def __reduce__(self):
...
+
         if llm_kwargs is not None:
-            return self._generate_using_openai(prompt_kwargs, llm_kwargs)
+            result = self._generate_using_openai(prompt_kwargs, llm_kwargs)
         else:
-            return self._generate_using_guidance(prompt_kwargs)
+            result = self._generate_using_guidance(prompt_kwargs)
+
+        if self._cache:
+            key = self._get_cache_key(prompt_kwargs, llm_kwargs)
```
pparmar30, line 292: we can use the key generated above!

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -251,18 +254,45 @@ def __reduce__(self):
...
     def generate(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> Any:
+        if self._cache:
+            key = self._get_cache_key(prompt_kwargs, llm_kwargs)
+            hit = self._cache.get(key)
+            if hit:
+                if hit.get("prompt_kwargs") == prompt_kwargs and hit.get("llm_kwargs") == llm_kwargs:
+                    return hit.get("result")
+                else:
+                    logger.warning(
+                        "Found cache content mismatch, key=%s prompt_kwargs=%s llm_kwargs=%s",
```
pparmar30, line 280: In case of a mismatch, do we want to remove the older cache?

lib/sycamore/sycamore/utils/cache_manager.py
baitsguy: I wanted to rename CacheManager to Cache actually since there really isn't any management going on - it's just a cache interface. Thoughts on that?

lib/sycamore/sycamore/llms/openai.py
baitsguy: The hashing bit yeah, we can add a common function that accepts a json input to hash. The prompt and llm args are unique to the llm cache

lib/sycamore/sycamore/llms/openai.py
baitsguy: They're both in conditional ops so the IDE was complaining :B Let me cleanup though

lib/sycamore/sycamore/llms/openai.py
baitsguy: I haven't added any evict functionality explicitly. The code under this will overwrite results anyway

baitsguy: The warning here is exposing a hash collision more than anything

lib/sycamore/sycamore/llms/openai.py
pparmar30: true but the cachemanager functions accepts bytes! so the content should not matter.

pparmar30:
lgtm

pparmar30 (approved):

PR #504, opened 2024-07-02
Add some writer unit tests

HenryL27:
adds a whole bunch of unit tests for the opensearch writer, weaviate writer, and some common functions
also fixed bugs that those tests exposed

### DIFF
--- a/lib/sycamore/sycamore/connectors/writers/common.py
+++ b/lib/sycamore/sycamore/connectors/writers/common.py
@@ -94,30 +94,36 @@ def drop_types(
     drop_additional_types: list[type] = [],
 ) -> Union[dict, list, tuple]:
     if isinstance(data, dict):
-        droppedd = {
-            k: drop_types(v, drop_nones, drop_empty_lists, drop_empty_dicts, drop_additional_types)
+        dropped_dict = {
+            k: (
+                drop_types(v, drop_nones, drop_empty_lists, drop_empty_dicts, drop_additional_types)
+                if isinstance(v, (list, tuple, dict))
+                else v
+            )
             for k, v in data.items()
         }
         if drop_nones:
-            droppedd = _filter_dict(_none_filter, droppedd)
+            dropped_dict = _filter_dict(_none_filter, dropped_dict)
         if drop_empty_lists:
-            droppedd = _filter_dict(_empty_list_filter, droppedd)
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/connectors/writers/common.py
```diff
@@ -140,6 +143,6 @@ def _empty_dict_filter(x: Any) -> bool:
 
 def _make_type_filter(types: list[type]) -> Callable[[Any], bool]:
     def _type_filter(x):
-        return isinstance(x, tuple(types))
+        return not isinstance(x, tuple(types))
```
karansampath, line 146: Having an exclusion check is potentially more inefficient since it guarantees having to check all values, could be worth having both an inclusion and exclusion check if possible

lib/sycamore/sycamore/tests/unit/connectors/writers/test_common.py
```diff
@@ -0,0 +1,96 @@
+from sycamore.connectors.writers.common import convert_to_str_dict, drop_types, flatten_data
+
+
+def test_flatten_data_happy():
```
karansampath, line 4: Would suggest adding / modifying this test to have some strings with a separator in them eg: `"a.."`

lib/sycamore/sycamore/tests/unit/connectors/writers/test_common.py
```diff
@@ -0,0 +1,96 @@
...
+    dropped = drop_types(data)
+    assert isinstance(dropped, list)
+    assert dropped == ["Hello"]
+
+
+def test_drop_types_tuple_to_tuple():
+    data = (None, "Hello")
+    dropped = drop_types(data)
+    assert isinstance(dropped, tuple)
+    assert dropped == ("Hello",)
```
karansampath, line 96: Shouldn't this be `(, "Hello")`?

karansampath (commented):
Mostly looks good. Have a couple of comments

lib/sycamore/sycamore/tests/unit/connectors/writers/test_common.py
HenryL27: My LSP complains about that syntax. fwiw I've only ever seen the comma after the singletuple elt

karansampath (approved):
LGTM

PR #511, opened 2024-07-04
OpenAI messages interface fix + OpenAI LLM cache fix

baitsguy:
1. Get `OpenAI` class to return a string response when using openai directly or guidance
2. Lazy initialize S3 client in S3Cache to handle serialization issues
3. Fixed serialization override in OpenAI and S3Cache classes

Added unit test for guidace/openai cases, extended pdf->opensearch test to cache LLM responses
Validated serialization with the following (fails prior to fix)
```
cache = S3Cache("s3://bucket/key")
llm = OpenAI("gpt-3.5-turbo", cache=cache)
pickled_obj = pickle.dumps(llm)
unpickled_obj = pickle.loads(pickled_obj)
```


lib/sycamore/sycamore/utils/cache.py
```diff
@@ -33,20 +34,23 @@ def set(self, hash_key: str, hash_value):
...
         self._s3_path = s3_path
         self._freshness_in_seconds = freshness_in_seconds
+        self._s3_client = None
 
     def _get_s3_bucket_and_key(self, key):
         parts = self._s3_path.replace("s3://", "").strip("/").split("/", 1)
         return parts[0], "/".join([parts[1], key]) if len(parts) == 2 else key
 
     def get(self, key: str):
+        if not self._s3_client:
```
baitsguy, line 47: @bsowell let me know your thoughts on this. I chose not to use the clientwrapper interface you've used in `OpenAI` because it seemed counter-intuitive to have to pass a CacheProvider instead of a cache. Happy to revisit

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -250,21 +254,19 @@ def __init__(
...
     # recreate the client on the other end.
     def __reduce__(self):
-        def deserializer(kwargs):
-            return OpenAI(**kwargs)
 
-        kwargs = {"client_wrapper": self.client_wrapper, "model_name": self._model_name}
+        kwargs = {"client_wrapper": self.client_wrapper, "model_name": self._model_name, "cache": self._cache}
 
-        return deserializer, (kwargs,)
+        return openai_deserializer, (kwargs,)
```
bsowell, line 260: Just out of curiosity, what changed with the deserializer?

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
```diff
@@ -92,33 +97,67 @@ def test_pdf_to_opensearch():
 
         """
 
-    paths = str(TEST_DIR / "resources/data/pdfs/")
+    s3_cache_base_path = os.environ["SYCAMORE_S3_TEMP_PATH"]
```
bsowell, line 100: Is this a net new variable we need for the integ tests? Is the plan to add it as a variable/secret in github.

lib/sycamore/sycamore/tests/unit/test_delete_me.py
```diff
@@ -0,0 +1,11 @@
+import pickle
```
bsowell, line 1: I presume you will delete this :).

lib/sycamore/sycamore/utils/cache.py
bsowell: I guess the big question is whether you ever need more parameters to pass to the client. In the OpenAI case, there are a bunch of different options that we need to expose, so we would need to pickle those, regardless. Here, I guess, the question is whether we would want to customize how we get credentials. Probably not a big deal since you can always use env vars if you need to.

bsowell (approved):
A few minor comments, but nothing major.

lib/sycamore/sycamore/llms/openai.py
baitsguy: No logical change, pickle just seems to complain if the function is part of the class you're serializing (based on the test in the description)

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
baitsguy: I stole it from `test_file_writer_to_s3.py` https://github.com/aryn-ai/sycamore/blob/472d22e116f1ef2e1250ae7221a19fb2dba27ab7/lib/sycamore/sycamore/tests/integration/connectors/file/test_file_writer_to_s3.py#L67

lib/sycamore/sycamore/tests/unit/test_delete_me.py
baitsguy: oops

lib/sycamore/sycamore/utils/cache.py
baitsguy: Yeah that's fair, I think we will need to at some point if we're using fancier client configs. Fine to defer until then

PR #515, opened 2024-07-07
process_batch() defaults table_structure_extractor.

alexaryn:
This will allow a remote call without serializing the whole TableTransformerStructureExtractor.

Also added a missing timetrace decorator.


austintlee (approved):
Thanks for making this change!

PR #516, opened 2024-07-07
Don't start PyTorch thread just doing import.

alexaryn:
The following modules:
- `easyocr`
- `sentence_transformers`
- `transformers`
- `torchvision`

all pull some shennanigans with `__init__.py` whereby the mere act of importing them starts a PyTorch thread.  Aside from being rude, this has the side effect of attaching to the GPU and holding a reference count on the GPU devices in `/dev/`.  In a Ray context, we don't want the driver to use the GPU, just the workers.  Also, just generally, importing something should not be so heavy-handed.


bsowell (approved):
LGTM

PR #517, opened 2024-07-08
Add Elasticsearch Writer

karansampath:
Adds a writer to [Elasticsearch](https://www.elastic.co/elasticsearch) with the necessary configuration changes needed


lib/sycamore/pyproject.toml
```diff
@@ -59,6 +59,7 @@ weaviate-client = "^4.6.1"
 pinecone-client = { version = "^4.1.0", extras = ["grpc"] }
 pinecone-text = "^0.9.0"
 duckdb = "1.0.0"
+elasticsearch = {version = "^8.14.0", optional = false}
```
HenryL27, line 62: nit: maybe leave a comment explaining why this was necessary

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    flatten_properties: bool = False
+
+    def compatible_with(self, other: BaseDBWriter.TargetParams) -> bool:
+        if not isinstance(other, ElasticTargetParams):
+            return False
+        if self.index_name != other.index_name:
+            return False
+        if self.mappings != other.mappings:
+            return False
```
HenryL27, line 43: I feel like this is gonna bite you when you try to write to an existing index and there are already mappings for already-existing documents.

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+class ElasticClientParams(BaseDBWriter.ClientParams):
+    url: str
+    es_client_args: dict = field(default_factory=lambda: {})
+
+
+@dataclass
+class ElasticTargetParams(BaseDBWriter.TargetParams):
+    index_name: str
+    mappings: dict[str, Any] = field(
+        default_factory=lambda: {
```
HenryL27, line 23: does elastic have index settings? would we want to expose them?

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+        assert isinstance(params, ElasticClientParams)
+        client = Elasticsearch(params.url, **params.es_client_args)
+        return ElasticClient(client)
+
+    def write_many_records(self, records: list[BaseDBWriter.Record], target_params: BaseDBWriter.TargetParams):
+        assert isinstance(target_params, ElasticTargetParams)
+        assert _narrow_list_of_doc_records(records), f"Found a bad record in {records}"
+        with self._client:
+
+            def insert_into_index():
```
HenryL27, line 62: nit: name this something like `bulk_action_generator()`

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+                for r in records:
+                    yield {
+                        "_index": target_params.index_name,
+                        "_id": r.doc_id,
+                        "properties": r.properties,
+                        "embeddings": r.embeddings,
+                    }
+
+            for success, info in parallel_bulk(self._client, insert_into_index()):  # generator must be consumed
```
HenryL27, line 71: what happens if generator is not consumed?

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+                if not success:
+                    print(f"Insert Operation Unsuccessful: {info}")
+
+    def create_target_idempotent(self, target_params: BaseDBWriter.TargetParams):
+        assert isinstance(target_params, ElasticTargetParams)
+        try:
+            self._client.indices.create(index=target_params.index_name, mappings=target_params.mappings, timeout="30s")
+        except ApiError as e:
+            if e.status_code == 400:
+                return
```
HenryL27, line 81: might wanna do a validation that the cause of the 400 was that the index exists already?

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+    def get_existing_target_params(self, target_params: BaseDBWriter.TargetParams) -> "ElasticTargetParams":
+        assert isinstance(target_params, ElasticTargetParams)
+        try:
+            data = self._client.indices.get_mapping(index=target_params.index_name)
+            mappings = data[target_params.index_name]["mappings"]
+            if "properties" in mappings and "embeddings" in mappings["properties"]:
+                try:
+                    mappings["properties"]["embeddings"].pop("index_options")
+                except Exception as e:
+                    print(f"Failed to pop index_options from embeddings: {e}")
```
HenryL27, line 93: generally prefer keeping lots of information in the retrieved target params and handling the inconsistencies from this in `compatible_with`. That's why it's `compatible_with` and not `__eq__`

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+        except ApiError as e:
+            raise e
```
HenryL27, line 95: I think 
```python
try:
    # do stuff
except e:
    raise e
```
is equivalent to 
```python
# do stuff
```
?

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+        properties = {
+            "properties": document.properties,
+            "type": document.type,
+            "text_representation": document.text_representation,
+            "bbox": document.bbox.coordinates if document.bbox else None,
+            "shingles": document.shingles,
+        }
+        droperties = drop_types(properties, drop_empty_lists=True, drop_empty_dicts=True)
+        if target_params.flatten_properties:
+            # Property names must be [a-zA-Z][_0-9a-zA-Z]{0,230}, so use __ as a separator rather than .
```
HenryL27, line 125: Is this statement true in elastic also?

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -0,0 +1,140 @@
...
+        if doc_id is None:
+            raise ValueError(f"Cannot write documents without a doc_id. Found {document}")
+        properties = {
+            "properties": document.properties,
+            "type": document.type,
+            "text_representation": document.text_representation,
+            "bbox": document.bbox.coordinates if document.bbox else None,
+            "shingles": document.shingles,
+        }
+        droperties = drop_types(properties, drop_empty_lists=True, drop_empty_dicts=True)
```
HenryL27, line 123: do empty lists and objects also cause problems in elastic?

lib/sycamore/sycamore/tests/integration/connectors/elasticsearch/test_pdf_to_elasticsearch.py
```diff
@@ -0,0 +1,34 @@
...
+        .regex_replace(COALESCE_WHITESPACE)
+        .mark_bbox_preset(tokenizer=tokenizer)
+        .merge(merger=MarkedMerger())
+        .spread_properties(["path"])
+        .split_elements(tokenizer=tokenizer, max_tokens=512)
+        .explode()
+        .embed(embedder=SentenceTransformerEmbedder(model_name=model_name, batch_size=100))
+        .sketch(window=17)
+    )
+    ds.write.elasticsearch(url=url, index_name=index_name)
```
HenryL27, line 34: prob worth doing a cleanup

HenryL27 (commented):
Main thing I care about is the `compatible_with` comments

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
karansampath: Yeah my bad, forgot to remove this had another print statement in there

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
karansampath: Yep, have included a way to expose them

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
karansampath: Fair, have added a better way of checking this

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
karansampath: The operation doesn't run. You could use a deque with `max_len = 0` for this but printing out unsuccessful results seems better to me

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -32,16 +32,18 @@ class ElasticTargetParams(BaseDBWriter.TargetParams):
             }
         }
     )
-    flatten_properties: bool = False
+    wait_for_completion: str = "false"
```
HenryL27, line 35: this one prob doesn't need to be a target param. I'm pretty sure the "wait for all actions before refreshing" option is always correct

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -120,19 +124,29 @@ def from_doc(cls, document: Document, target_params: BaseDBWriter.TargetParams)
...
+        if key not in dict2:
+            return False
+        if isinstance(value, dict):
+            if not isinstance(dict2[key], dict):
+                return False
+            if not compare_nested_dicts(value, dict2[key]):
+                return False
+        elif dict1[key] != dict2[key]:
+            return False
+    return True
```
HenryL27, line 147: prob move to common.
also in opensearch I ran into an issue where I could put in a key like so
```json
{
  "index.knn": "true"
}
```
and it would come out as 
```json
{
  "index": {
    "knn": "true"
  }
}
```
You can check out the opensearch compat function for the shenanigans I pulled to make it work, but the point is, I'm worried about elastic having similar issues

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
HenryL27: Oh I see you use it in places in the test. ok, nvm

lib/sycamore/sycamore/tests/integration/connectors/elasticsearch/test_pdf_to_elasticsearch.py
```diff
@@ -31,4 +33,10 @@ def test_to_elasticsearch():
         .embed(embedder=SentenceTransformerEmbedder(model_name=model_name, batch_size=100))
         .sketch(window=17)
     )
-    ds.write.elasticsearch(url=url, index_name=index_name)
+    count = ds.count()
+    ds.write.elasticsearch(url=url, index_name=index_name, wait_for_completion=wait_for_completion)
+    es_client = Elasticsearch(url)
+    es_count = int(es_client.cat.count(index=index_name, format="json")[0]["count"])
+    es_client.indices.delete(index=index_name)
+    es_client.close()
```
HenryL27, line 41: generally good practice to use 
```python
with Elasticsearch(url) as es_client:
    # blah
```
just in case of errors

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
karansampath: No I don't think so, I added it as consistency so that we could use the same unflatten function. Have removed it now

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
karansampath: No, just felt cleaner to do this

lib/sycamore/sycamore/writer.py
```diff
@@ -497,11 +499,19 @@ def elasticsearch(
         )
 
         client_params = ElasticClientParams(url=url, es_client_args=es_client_args)
-        target_params = ElasticTargetParams(index_name=index_name, flatten_properties=flatten_properties)
-        if mappings:
-            target_params = ElasticTargetParams(
-                index_name=index_name, mappings=mappings, flatten_properties=flatten_properties
-            )
+        target_params = ElasticTargetParams(index_name=index_name)
```
HenryL27, line 502: delete?

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
karansampath: You're right good catch, I hadn't tested it on this. Have updated it to use the opensearch code

HenryL27 (approved):
lgtm

PR #519, opened 2024-07-08
Move initialization of easyocr Reader to Actor level.

bsowell:
This improves performance by a signficant amount (order 2x).


alexaryn (approved):
Looks pretty straightforward.

PR #521, opened 2024-07-09
Bugfix intermediate data

baitsguy:
Was modifying 'intermediate_datasink_kwargs' which was resulting in nested directories for each node.

### DIFF
--- a/lib/sycamore/sycamore/tests/unit/transforms/test_base.py
+++ b/lib/sycamore/sycamore/tests/unit/transforms/test_base.py
@@ -325,7 +325,7 @@ class TestBaseMapTransform(Common):
         for d in Common.dicts:
             truth_id_to_content[d["doc_id"]] = d
 
-        # assert
+        # assert there is a directory for each node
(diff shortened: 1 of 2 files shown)


HenryL27 (approved):
splat splat

not sure how this was a bug but if it fixes then ok.
wait does it fix?

baitsguy:
> splat splat
> 
> not sure how this was a bug but if it fixes then ok. wait does it fix?

`intermediate_datasink_kwargs["path"] = intermediate_datasink_kwargs["path"] + "/" + self._name` was modifying `intermediate_datasink_kwargs`. We send `intermediate_datasink_kwargs` to the parent execution node

```
input_dataset = self.child().execute(
            write_intermediate_data=write_intermediate_data,
            intermediate_datasink=intermediate_datasink,
            intermediate_datasink_kwargs=intermediate_datasink_kwargs,
        )
```
So each child execution was appending it's own name to the path property of the shared `intermediate_datasink_kwargs` object we were sending accross

PR #523, opened 2024-07-10
Allow custom names for Map/FlatMap nodes

baitsguy:
Based on the ordering of params, we're unable to inject a custom name into a Map node. This fixes the order


HenryL27 (approved):
lgtm! any chance of switching `resource_args` to `kwargs` in this spot? My rule has been, if I touch it I change it over

HenryL27 (approved):

PR #524, opened 2024-07-11
Fix Pinecone Test

karansampath:
Fixes the `test_pinecone_scan` integration test that was not working due to eventual consistency issues

### DIFF
--- a/lib/sycamore/sycamore/tests/integration/connectors/pinecone/test_pinecone_scan.py
+++ b/lib/sycamore/sycamore/tests/integration/connectors/pinecone/test_pinecone_scan.py
@@ -11,6 +11,8 @@ from sycamore.transforms.partition import UnstructuredPdfPartitioner
 from sycamore.transforms.embed import SentenceTransformerEmbedder
 from sycamore.tests.config import TEST_DIR
 from pinecone.grpc import PineconeGRPC
+from pinecone import PineconeException
+import time
 
 
 def test_pinecone_scan():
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/tests/integration/connectors/pinecone/test_pinecone_scan.py
```diff
@@ -54,3 +60,24 @@ def test_pinecone_scan():
...
+    """
+    ready = False
+    timeout = 30
+    deadline = time.time() + timeout
+    index = client.Index(index_name)
+    while not ready:
+        try:
+            desc = dict(index.fetch(ids=[doc_id], namespace=namespace)["vectors"]).items()
+            if len(desc) > 0:
+                ready = True
```
HenryL27, line 77: suggest:
`while time.time() < deadline:`
...
`if len(desc) > 0: return`

HenryL27 (approved):
lgtm. pay no mind to the suggestion unless you want to.
I guess let's see if it gets the test to pass

lib/sycamore/sycamore/tests/integration/connectors/pinecone/test_pinecone_scan.py
```diff
@@ -66,18 +66,15 @@ def wait_for_write_completion(client: PineconeGRPC, index_name: str, namespace:
...
             desc = dict(index.fetch(ids=[doc_id], namespace=namespace)["vectors"]).items()
             if len(desc) > 0:
-                ready = True
+                return
         except PineconeException:
             # NotFoundException means the last document has not been entered yet.
             pass
         time.sleep(1)
-        if time.time() > deadline:
-            raise RuntimeError(f"Pinecone failed to write results in {timeout} seconds. Doc_id: {doc_id}")
```
HenryL27: still want to raise this after the loop I think? I guess you can just yolo the read, eh

karansampath: Yeah I think that'll be better, there might be multiple reasons for it failing

PR #528, opened 2024-07-12
Use printer-style page selection option to batch pages in ArynPartitioner

MarkLindblad:


lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -413,6 +413,7 @@ def __init__(
         aryn_api_key: str = "",
         aryn_partitioner_address: str = DEFAULT_ARYN_PARTITIONER_ADDRESS,
         use_cache=False,
+        selection_size: int = 25,
```
bsowell, line 416: Why do you like `selection_size` better than `num_pages_per_call`? The latter sounds a bit clearer to me, though I can be convinced otherwise.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -238,6 +244,47 @@ def _partition_remote(
...
+        threshold: float = 0.4,
+        use_ocr: bool = False,
+        ocr_images: bool = False,
+        ocr_tables: bool = False,
+        extract_table_structure: bool = False,
+        extract_images: bool = False,
+        selection_size: int = 25,
+    ) -> List[Element]:
+        file.seek(0)
+        pdf_reader = PyPDF2.PdfReader(file)
```
bsowell, line 261: Do we need to bring in PyPDF2 for this? It looks like you can get teh number of pages with pdfminer as well, which we already import (see https://stackoverflow.com/a/47686921). We keep getting deprecation warnings about PyPDF2, so would be nice to avoid if its easy.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -238,6 +244,47 @@ def _partition_remote(
...
+        page_count = len(pdf_reader.pages)
+        file.seek(0)
+
+        result = []
+        low = 0
+        high = selection_size - 1
+        while low < page_count:
+            result.extend(
```
bsowell, line 269: Did we ever change the result to be a dict, or is it still just a list of elements? Seems here that it's still a list, but I thought we decided otherwise. If we change it, this will probably need to change as well.

lib/sycamore/sycamore/transforms/partition.py
MarkLindblad: Changed to `pages_per_call`.

lib/sycamore/sycamore/transforms/detr_partitioner.py
MarkLindblad: Yes, we have changed the result to be a dict, and this is accounted for: see [`detr_partitioner:237`](https://github.com/aryn-ai/sycamore/blob/7e2373de6ec1f4f07a00e0ea8d59cb2652ab59fe/lib/sycamore/sycamore/transforms/detr_partitioner.py#L237)

lib/sycamore/sycamore/transforms/detr_partitioner.py
MarkLindblad: Thanks for the optimization. Done.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -413,6 +413,7 @@ def __init__(
         aryn_api_key: str = "",
         aryn_partitioner_address: str = DEFAULT_ARYN_PARTITIONER_ADDRESS,
         use_cache=False,
+        pages_per_call: int = 25,
```
bsowell, line 416: How did you pick 25? I'm debating myself whether I think the default should be something like this or just sending the whole doc. 

I do wonder if we can have some value that says send the whole doc, without requiring the caller to know the number of pages in the pdf. Maybe None or -1?

MarkLindblad: Turned off by default.  Added the option `-1` to mean send the whole doc.

bsowell (approved):

PR #529, opened 2024-07-12
add weaviate writer docs. Also add connectors to ToC

HenryL27:
The compose file looks big here, but it's in a collapsible and looks like this by default:
Screenshot 2024-07-12 at 2 11 29 PM

also adds to the ToC like so
```
Data Ingestion and Preparation
  ..
  ..
  Transforms     v
  Connectors     ^
    DuckDB
    Weaviate
```

### DIFF
--- a/docs/source/data_ingestion_and_preparation/connectors/weaviate.md
+++ b/docs/source/data_ingestion_and_preparation/connectors/weaviate.md
@@ -58,7 +58,7 @@ To write a DocSet to a Weaviate Collection from Sycamore, use the `docset.write.
 - `collection_name`: The name of the collection to write to.
 - `collection_config`: (optional) A dictionary of keyword parameters passed to the Weaviate client's [`collection.create(...)`](https://weaviate.io/developers/weaviate/client-libraries/python#instantiate-a-collection) method. A name specified here must match the `collection_name` argument.
 - `flatten_properties`: (optional, default=`False`) Whether to flatten nested property objects during the write. Weaviate can store flattened and nested properties, but will only filter and aggregate top-level properties, meaning they need to be flattened.
-- `execute`: (optional, defaul=`True`) Whether to execute this sycamore pipeline now, or return a docset to add more transforms.
+- `execute`: (optional, default=`True`) Whether to execute this sycamore pipeline now, or return a docset to add more transforms.
 
 To write a docset to the weaviate instance run by the docker compose above, we can write the following:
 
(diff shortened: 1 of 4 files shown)


docs/source/data_ingestion_and_preparation/connectors/weaviate.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Writing to Weaviate
+
+To write a DocSet to a Weaviate Collection from Sycamore, use the `docset.write.weaviate(...)` function. The Weaviate writer takes the following arguments:
+
+- `wv_client_args`: A dictionary of arguments to pass to the Weaviate client constructor, as in an [explicit conection](https://weaviate.io/developers/weaviate/client-libraries/python#python-client-v4-explicit-connection).
+- `collection_name`: The name of the collection to write to.
+- `collection_config`: (optional) A dictionary of keyword parameters passed to the Weaviate client's [`collection.create(...)`](https://weaviate.io/developers/weaviate/client-libraries/python#instantiate-a-collection) method. A name specified here must match the `collection_name` argument.
+- `flatten_properties`: (optional, default=`False`) Whether to flatten nested property objects during the write. Weaviate can store flattened and nested properties, but will only filter and aggregate top-level properties, meaning they need to be flattened.
+- `execute`: (optional, defaul=`True`) Whether to execute this sycamore pipeline now, or return a docset to add more transforms.
```
karansampath, line 61: nit: default

docs/source/data_ingestion_and_preparation/connectors/weaviate.md
```diff
@@ -0,0 +1,94 @@
...
+
+To write a docset to the weaviate instance run by the docker compose above, we can write the following:
```
karansampath, line 63: Won't this be in the API reference already? Best to avoid duplication if possible

HenryL27: I guess the reason to have this here is that it works with the specific configuration set in the docker compose above, so the reader can better see how the things fit together

karansampath: It seems very similar to the API [example](https://github.com/aryn-ai/sycamore/blob/24f7f65401a8950db3ba89a48014f6ae68e48a3d/lib/sycamore/sycamore/writer.py#L175)? The connection parameters and vectorizer_config is the same. Let me know what you think, this is not done in the others so I think it's worth having a standardized approach to this

jonfritz (approved):
LGTM!

PR #531, opened 2024-07-12
Change ArynPartitioner API to use `use_partitioning_service` instead of `local`

MarkLindblad:


lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -409,12 +409,12 @@ def __init__(
...
         batch_size: int = 1,
         batch_at_a_time: bool = False,
-        local: bool = False,
+        use_partitioning_service: bool = True,
         aryn_api_key: str = "",
         aryn_partitioner_address: str = DEFAULT_ARYN_PARTITIONER_ADDRESS,
         use_cache=False,
     ):
-        if local:
+        if not use_partitioning_service:
```
karansampath, line 417: Could you switch around the logic for this? Would like to avoid the use of the negative `if not` if possible

karansampath: Is there a need for the if statement at all? You could set `device = choose_device(device)` regardless since it won't matter if the remote partitioner is being used?

karansampath (approved):
LGTM, small changes

lib/sycamore/sycamore/transforms/partition.py
MarkLindblad: Changed logic around.

MarkLindblad: Since we [haven't yet changed](https://github.com/aryn-ai/sycamore/blob/3d83de4f90dff2e2271b7971bc4f177538acc15b/lib/sycamore/sycamore/transforms/detr_partitioner.py#L81) the `ArynPDFPartitioner` to conditionally load the model, this makes sure we don't take up GPU memory. I'll make another pr to fix this.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -381,7 +381,7 @@ class ArynPartitioner(Partitioner):
         extract_images: If true, crops each region identified as an image and attaches it to the associated
              ImageElement. This can later be fed into the SummarizeImages transform.
         local: If false, runs the partitioner remotely. Defaults to false
```
pparmar30, line 383: probably need to update this comment

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -409,15 +409,15 @@ def __init__(
...
-        local: bool = False,
+        use_partitioning_service: bool = True,
         aryn_api_key: str = "",
         aryn_partitioner_address: str = DEFAULT_ARYN_PARTITIONER_ADDRESS,
         use_cache=False,
     ):
-        if local:
-            device = choose_device(device)
-        else:
+        if use_partitioning_service:
```
pparmar30, line 417: Don't you want `if not use_partitioning_service` ?

MarkLindblad: No

lib/sycamore/sycamore/transforms/partition.py
MarkLindblad: Yes. Good catch - I'll fix this in a new pr.

PR #533, opened 2024-07-15
Add Reader API and move DuckDB

karansampath:
Adds the Reader API as discussed and moves the DuckDB implementation over to the new structure. This adds new functionality and still allows the old tests to function as (i.e. defaults to a scan if arguments not specified)


lib/sycamore/sycamore/connectors/base_reader.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(
+        self,
+        plan: Optional[Node],
+        client_params: ClientParams,
+        query_params: QueryParams,
+        **kwargs,
+    ):
+        super().__init__(plan, **kwargs)
+        check_serializable(client_params, query_params, filter)
```
HenryL27, line 59: filter?

lib/sycamore/sycamore/connectors/duckdb/duckdb_reader.py
```diff
@@ -0,0 +1,76 @@
...
+        con = duckdb.connect(database=query_params.db_url, read_only=True)
+        results = []
+        if query_params.on_input_docs and query_params.query:
+            for doc in input_docs:  # noqa
+                results.append(DuckDBReaderDocumentRecord(output=con.execute(f"{query_params.query}")))
+        else:
+            if query_params.query:
+                results = [DuckDBReaderDocumentRecord(con.execute(query_params.query))]
+            else:
+                results = [DuckDBReaderDocumentRecord(con.execute(f"SELECT * from {query_params.table_name}"))]
```
HenryL27, line 45: I feel like a lot of the control flow logic here is independent of the connector. e.g. for opensearch I want the same set of behaviors. Can we factor it up to the superclass somehow?

lib/sycamore/sycamore/connectors/duckdb/duckdb_reader.py
```diff
@@ -0,0 +1,76 @@
...
+        return results
+
+    def check_target_presence(self, query_params: BaseDBReader.QueryParams):
+        assert isinstance(query_params, DuckDBReaderQueryParams)
+        client = duckdb.connect(query_params.db_url, read_only=True)
+        try:
+            client.sql(f"SELECT * FROM {query_params.table_name}")
+            return True
```
HenryL27, line 53: Not sure if it matters, but if the query specifies particular columns or something I might check that they exist here too

lib/sycamore/sycamore/connectors/duckdb/duckdb_reader.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@dataclass
+class DuckDBReaderClientParams(BaseDBReader.ClientParams):
+    pass
+
+
+@dataclass
+class DuckDBReaderQueryParams(BaseDBReader.QueryParams):
+    db_url: str
```
HenryL27, line 17: slight preference for putting db_url in ClientParams but either way

lib/sycamore/sycamore/connectors/base_reader.py
```diff
@@ -0,0 +1,76 @@
...
+    def read_docs(self, input_docs: list[Document]) -> Dataset:
+        client = self.Client.from_client_params(self._client_params)
+
+        if not client.check_target_presence(self._query_params):
+            raise ValueError("Target is not present\n" f"Parameters: {self._query_params}\n")
+        records = client.read_records(input_docs=input_docs, query_params=self._query_params)
+        docs = [self.Record.to_doc(r, self._query_params) for r in records]
+        flat_docs = [item for sublist in docs for item in sublist]
+        input_docs.extend(flat_docs)
+        return from_items(items=[{"doc": doc.serialize()} for doc in input_docs])
```
HenryL27, line 72: Suggestion: move the from_items call to execute (keep all the dataset/ray stuff in one place)

lib/sycamore/sycamore/connectors/base_reader.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def execute(self, input_docs: list[Document] = [], **kwargs) -> Dataset:
+        with TimeTrace("Reader"):
+            return self.read_docs(input_docs=input_docs)
```
HenryL27, line 76: If this node has a child, the child's data gets dropped here. See [here](https://github.com/aryn-ai/sycamore/blob/c72da4d03ab4cbb32c40a4ce74898ff17d42cdd7/lib/sycamore/sycamore/transforms/base.py#L122) for how to make execute work right.

HenryL27 (commented):
This is sufficiently complicated that I'd like to see some unit tests.

Tbh, I'm not sure I would combine the [nothing-to-docset] and [docset-to-docset] versions into one api. Maybe we have the DocSetReader methods and then a DocSet.query method that accepts a BaseDBReader and runs read_docs as a flat_map?

lib/sycamore/sycamore/connectors/base_reader.py
karansampath: Removed

lib/sycamore/sycamore/connectors/base_reader.py
karansampath: Not changing since this implementation will not have children now

lib/sycamore/sycamore/connectors/duckdb/duckdb_reader.py
karansampath: Don't think we'd want to do this since it would require a lot of parsing work / has potential to break with not much value add

lib/sycamore/sycamore/connectors/duckdb/duckdb_reader.py
karansampath: Don't think this is worth it now that we've simplified this a bunch?

lib/sycamore/sycamore/connectors/duckdb/duckdb_reader.py
karansampath: I agree generally but in DuckDB since it initializes the connection quickly I'd prefer starting the work as late as possible

lib/sycamore/sycamore/connectors/duckdb/duckdb_reader.py
```diff
@@ -0,0 +1,67 @@
...
+class DuckDBReaderDocumentRecord(BaseDBReader.Record):
+    output: duckdb.DuckDBPyConnection
+
+    @classmethod
+    def to_doc(cls, record: "BaseDBReader.Record", query_params: "BaseDBReader.QueryParams") -> list[Document]:
+        assert isinstance(record, DuckDBReaderDocumentRecord)
+        data = record.output.fetchdf().to_dict(orient="records")
+        result = []
+        for object in data:
+            result.append(Document(object))
```
HenryL27, line 59: do you need to unstringify `properties`?

karansampath: Yep, good catch

lib/sycamore/sycamore/connectors/common.py
```diff
@@ -148,6 +148,31 @@ def convert_to_str_dict(data: dict[str, Any]) -> dict[str, str]:
...
+            except ValueError:
+                try:
+                    result[key] = float(value)
+                except ValueError:
+                    try:
+                        # Try to parse as JSON (for lists and dicts)
+                        result[key] = json.loads(value)
+                    except json.JSONDecodeError:
+                        # If all else fails, keep it as a string
+                        result[key] = value
```
HenryL27, line 172: This is awesome! I can't help but wonder if there's a better way but hell yeah

HenryL27 (approved):

PR #537, opened 2024-07-15
Fix partitioner docstring

HenryL27:
I'm like 96% sure that this is incorrect in some way, so please read

### DIFF
--- a/lib/sycamore/sycamore/transforms/partition.py
+++ b/lib/sycamore/sycamore/transforms/partition.py
@@ -383,7 +383,8 @@ class ArynPartitioner(Partitioner):
         device: Device on which to run the partitioning model locally. One of 'cpu', 'cuda', and 'mps'. If
              not set, Sycamore will choose based on what's available. If running remotely, this doesn't
              matter.
-        batch_size: How many pages to partition at once. Default is 1.
+        batch_size: How many pages to partition at once, when running locally. Default is 1. Ignored when
+             running remotely.
         batch_at_a_time: When running remotely, send only one batch at a time. Default is False
         local: If false, runs the partitioner remotely. Defaults to false
         aryn_api_key: The account token used to authenticate with Aryn's servers.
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -380,9 +380,17 @@ class ArynPartitioner(Partitioner):
...
+             matter.
+        batch_size: How many pages to partition at once. Default is 1.
+        batch_at_a_time: When running remotely, send only one batch at a time. Default is False
         local: If false, runs the partitioner remotely. Defaults to false
-        aryn_token: The account token used to authenticate with Aryn's servers.
+        aryn_api_key: The account token used to authenticate with Aryn's servers.
         aryn_partitioner_address: The address of the server to use to partition the document
+        use_cache: Cache results from the partitioner for faster inferences on the same documents in future runs.
+        pages_per_call: Number of pages to send in a single call to the remote service. Default is -1,
+             which means any.
```
MarkLindblad, line 393: ....Default is -1 which means **send all pages in one call.**

HenryL27: But I got the batch stuff right???

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -380,9 +380,17 @@ class ArynPartitioner(Partitioner):
              Ignored when local mode is false.
         extract_images: If true, crops each region identified as an image and attaches it to the associated
              ImageElement. This can later be fed into the SummarizeImages transform.
+        device: Device on which to run the partitioning model locally. One of 'cpu', 'cuda', and 'mps'. If
+             not set, Sycamore will choose based on what's available. If running remotely, this doesn't
+             matter.
+        batch_size: How many pages to partition at once. Default is 1.
```
MarkLindblad, line 386: locally.  Ignored when running remotely.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -380,9 +380,17 @@ class ArynPartitioner(Partitioner):
...
+        batch_at_a_time: When running remotely, send only one batch at a time. Default is False
```
MarkLindblad, line 388: batch_at_a_time: When running locally, True means inference is run on the pages in batches - not all pages are loaded into memory at the same time.

MarkLindblad (approved):

PR #540, opened 2024-07-16
set table text rep to csv rep of table

HenryL27:
instead of whatever text representation it would ordinarily be. Can set it to something else if needed tho with just `table.text_representation = 'blah'`


bsowell (approved):
LGTM. Thanks for doing this.

PR #541, opened 2024-07-16
Seperate element reordering function into two parts for use in Partitioning Service

MarkLindblad:


alexaryn (approved):
Looks fine to me.

PR #542, opened 2024-07-16
Add __str__ methods to Document and Element.

mdwelsh:
This is helpful for debugging as it does not dump the entire contents of, e.g., embedding and shingles.

### DIFF
--- a/lib/sycamore/sycamore/data/document.py
+++ b/lib/sycamore/sycamore/data/document.py
@@ -200,10 +200,12 @@ class Document(UserDict):
             "lineage_id": self.lineage_id,
             "type": self.type,
             "text_representation": self.text_representation[0:40] + "..." if self.text_representation else None,
-            "binary_representation": f"<{len(self.binary_representation)} bytes>" if self.binary_representation else None,
+            "binary_representation": (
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/data/document.py
```diff
@@ -192,6 +193,23 @@ def to_row(self) -> dict[str, bytes]:
...
+    def __str__(self) -> str:
+        """Return a pretty-printed string representing this document."""
+        d = {
+            "doc_id": self.doc_id,
+            "lineage_id": self.lineage_id,
+            "type": self.type,
+            "text_representation": self.text_representation[0:40] + "..." if self.text_representation else None,
+            "binary_representation": f"<{len(self.binary_representation)} bytes>" if self.binary_representation else None,
+            "elements": [str(e) for e in self.elements],
+            "embedding": f"<{len(self.embedding)} floats>" if self.embedding else None,
```
karansampath, line 205: I'd suggest showing some sample values as well, would help with debugging, same with `shingles` below

karansampath (approved):
LGTM, made a small comment

lib/sycamore/sycamore/data/document.py
mdwelsh: Good idea, done!

PR #544, opened 2024-07-16
Move Writer Notebooks to ArynPartitioner

karansampath:
Moves the notebooks to the Aryn Partitioner

### DIFF
--- a/notebooks/duckdb-writer.ipynb
+++ b/notebooks/duckdb-writer.ipynb
@@ -1,13 +1,21 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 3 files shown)


notebooks/duckdb-writer.ipynb
```diff
@@ -47,6 +55,16 @@
     "    return doc"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "aryn_api_key = os.environ.get(\"ARYN_API_KEY\") ## REPLACE WITH YOUR-ARYN-API-KEY\n",
```
Soeb-aryn, line 64: looks good to me, just one thing. Jon wants to pass api key as a string in notebooks, can you check with him also ?

Soeb-aryn (approved):

PR #545, opened 2024-07-17
HTTP client that acts more like curl.

alexaryn:
Example use:
```
c = OneShotKaClient("https://test-api.foo.com/v1/magic")
tok = "ey...AQ"
hdr = {"Authorization": f"Bearer {tok}"}
form = {
    "options": '{"use_foo": true}',
    "pdf": "@files/doc.pdf",
}
buf = c.post(headers=hdr, form=form)
print(buf)
print(c.resp.status, c.resp.getheaders())
```

### DIFF
--- a/lib/sycamore/sycamore/utils/http.py
+++ b/lib/sycamore/sycamore/utils/http.py
@@ -1,0 +1,122 @@
+import ssl
+import random
+import mimetypes
+import http.client
+import urllib.parse
+
+from typing import Optional
+
+
+class OneShotKaClient:
+    """
+    Acts like a keepalive client, but actually closes the connection at
+    the end.  Basically, doesn't send Connection: close header as
+    urllib.request does.  This makes it more like curl.
+    """
+
+    alphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+
+    def __init__(self, url: str) -> None:
+        self.url = url
+        self.up = urllib.parse.urlparse(url)
+        self.agent = "ArynKa/1.0"
+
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/utils/http.py
```diff
@@ -0,0 +1,122 @@
...
+            "Accept": "*/*",
+            "Accept-Encoding": "identity",
+        }
+        hdr.update(headers)
+
+        self._htconn()
+        self.conn.request("GET", self.up.path, headers=hdr)
+        return self._finish()
+
+    def post(self, *, headers: dict[str, str] = {}, form: dict[str, str] = {}) -> bytes:
```
MarkLindblad, line 40: I think you narrowly miss [this issue](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python) by how carefully the code is written. No need for a change.

MarkLindblad: I am not sure how this would affect multithreading.

MarkLindblad (approved):
Beautiful code that looks easy to use.

lib/sycamore/sycamore/utils/http.py
alexaryn: I was debating defaulting to None.  Perhaps it's a better policy for the future.

I don't think threading is an issue for this class, as there's no real sharing.  One instance of OneShotKaClient can be passed from thread to thread as long as only one thread uses it at a time.

MarkLindblad: My concern is that all instances are sharing the same default dictionary and that they need to be careful to take turns changing it.

PR #546, opened 2024-07-17
Graph metadata extractor and Graph Extractor Transform

RitxmSaha:
This PR contains an extractor to take metadata attached from documents and turn them into nodes that can be loaded into neo4j.


lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -0,0 +1,155 @@
...
+    def extract_metadata(self, doc: Document) -> Document:
+        """
+        Extracts metadata from documents and stores them in the 'nodes' key of 'properties in each document
+        """
+        nodes: Dict[str, Dict[str, Any]] = {}
+        for m in self.metadata:
+            if m.nodeKey not in doc["properties"]:
+                continue
+
+            key = str(m.nodeKey)
```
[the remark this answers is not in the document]
RitxmSaha: oops yes. sorry while I was developing I was dealing with a lot of ray serialization stuff in the aggregate phase so I was playing around with the casting.

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -0,0 +1,155 @@
...
+            value = str(doc["properties"][m.nodeKey])
+            node = {"type": "metadata", "properties": {key: value}, "label": str(m.nodeLabel), "relationships": {}}
+            nodes[key + "_" + value] = node
+
+            rel = {
+                "TYPE": str(m.relLabel),
+                "properties": {},
```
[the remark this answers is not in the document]
RitxmSaha: The doc properties are currently stored in the document node(this is a map function I have that I don't have pushed). I added the properties tag here just in case down the line we wanted to add properties here unsupervised.

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -0,0 +1,155 @@
...
+        docs = dataset.take_all(None)
+        docs = [Document.deserialize(d["doc"]) for d in docs]
+
+        for doc in docs:
+            if "properties" in doc:
+                if "nodes" in doc["properties"]:
+                    del doc["properties"]["nodes"]
+
+        for value in result["nodes"].values():
+            docs[0]["elements"].append(value)
```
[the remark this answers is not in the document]
RitxmSaha: Yeah so this is because once we explode after this step, all the elements become top level documents(so in this case each node gets turned into a top level document). Then I feed this into a neo4j loader. I wish there was a better solution cause this seems scrappy. do you have any ideas?

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class GraphMetadata(GraphData):
+    """
+    Object which handles what fields to extract metadata from and what fields to represent them as in neo4j
+
+    Args:
+        nodeKey: Key used to access document metadata in the document['properties] dictionary
+        nodeLabel: The label used in neo4j the node of a piece of metadata
+        relLabel: The label used in neo4j for the relationship between the document and a piece of metadata
```
dtecuci, line 24: Is there a description of the type of entity to be extracted? Or are you relying only on the name (nodeLabel)?

RitxmSaha: This is for the metadata extractor currently so it extracts by the key of the metadata provided in the initial jsonmanifest

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -139,15 +95,58 @@ def finalize(nodes):
...
+
+        return docset
+
+    def extract_metadata(self, doc: Document) -> Document:
+        """
+        Extracts metadata from documents and stores them in the 'nodes' key of 'properties in each document
+        """
+        nodes: Dict[str, Dict[str, Any]] = {}
+        for m in self.metadata:
+            if m.nodeKey not in doc["properties"]:
```
baitsguy, line 131: fun note from @alexaryn : you can replace this with

```
key = ...
value = doc["properties"].get(key)
if value is None:
  continue
# remaining logic

```
This results in 1 map read instead of 2. Not really an issue here but interesting

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -139,15 +95,58 @@ def finalize(nodes):
...
+                continue
+
+            key = m.nodeKey
+            value = str(doc["properties"][m.nodeKey])
+            node = {"type": "metadata", "properties": {key: value}, "label": str(m.nodeLabel), "relationships": {}}
+            nodes[key + "_" + value] = node
```
baitsguy, line 137: why not just do this after creating the relationship below?

baitsguy (approved):
Couple small nits but not blocking. I suspect we'll need to revisit some of this logic as we test, mainly curious about how the new Document that contains all graph nodes will behave.

PR #549, opened 2024-07-17
Add HTML table operations

HenryL27:
Parse html tables into sycamore.data.Table objects.
Use that in HtmlPartitioner
Add TEDs implementation for Table objects.

Next thing is to squish the TEDs into the eval framework and run PubTabNet or something through it, but that's a different PR


lib/sycamore/sycamore/data/table.py
```diff
@@ -143,6 +145,96 @@ def from_dict(cls, dict_obj: dict[str, Any]) -> "Table":
...
+                    )
+                )
+
+                cur_col += colspan
+
+            elif tag.name == "caption":
+                caption = tag.get_text()
+
+            else:
+                continue
```
baitsguy, line 219: remove or maybe log if we shouldn't be here?

lib/sycamore/sycamore/data/table.py
```diff
@@ -143,6 +145,96 @@ def from_dict(cls, dict_obj: dict[str, Any]) -> "Table":
...
+
+        # Traverse the tree of elements in a pre-order traversal.
+        for tag in root.find_all(recursive=True):
+            if tag.name == "tr":
+                cur_row += 1  # TODO: Should this be based on rowspan?
+                cur_col = 0
+
+            elif tag.name in {"td", "th"}:
+                # We allow a missing tr for the first row to handle the case when
+                # they have a thead.
```
baitsguy, line 195: is it possible to have multiple theads? Can a single table contain 2 header rows

lib/sycamore/sycamore/data/table.py
```diff
@@ -143,6 +145,96 @@ def from_dict(cls, dict_obj: dict[str, Any]) -> "Table":
...
+                continue
+
+        # Fix columns where rowspans should be inserted
+        for c in cells:
+            if len(c.rows) == 1:
+                continue
+            rows_to_increment = c.rows[1:]
+            increment_by = len(c.cols)
+            increment_after = c.cols[0]
+            # Yes this is quadratic. I'm assuming tables are smaller than LLMs.
```
baitsguy, line 228: lol

lib/sycamore/sycamore/transforms/explode.py
```diff
@@ -36,6 +37,8 @@ def explode(parent: Document) -> list[Document]:
             cur = Document(element.data)
             cur.doc_id = str(uuid.uuid4())
             cur.parent_id = parent.doc_id
+            if isinstance(element, TableElement):
```
baitsguy, line 40: why do we need special handling here?

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -332,24 +332,9 @@ def partition(self, document: Document) -> Document:
                 if len(table.find_all("table")) > 0:
                     continue
 
-                table_element = TableElement()
```
baitsguy, line 335: nice

baitsguy (commented):
Looks good, just couple small comments

lib/sycamore/sycamore/data/table.py
HenryL27: yep

lib/sycamore/sycamore/transforms/explode.py
HenryL27: In a previous PR I changed the behavior of TableElement.text_representation to 
```
if overridden: overridden_text
else if table: table.to_csv()
else: text_representation
```
(and setting text_representation sets the overridden).

But just importing the underlying data essentially deletes this behavior when the table gets promoted to a document

baitsguy (approved):

PR #552, opened 2024-07-18
Add Sort as a Transform

tranade:
Add sort to sycamore transforms


lib/sycamore/sycamore/transforms/sort.py
```diff
@@ -0,0 +1,56 @@
...
+        # creates dataset
+        ds = self.child().execute(**kwargs)
+
+        # adds a "key" column containing desired field
+        map_fn = self.make_map_fn_sort()
+        ds = ds.map(map_fn)
+
+        # sorts the dataset
+        ds = ds.sort("key", descending=self._descending)
+        return ds
```
baitsguy, line 30: should we remove the key after?

baitsguy (commented):
Nice work! Some comments to clarify

lib/sycamore/sycamore/transforms/sort.py
tranade: I don't actually know if it makes a difference once you convert to docset, but will do just in case

baitsguy (approved):

lib/sycamore/sycamore/transforms/sort.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def make_map_fn_sort(self):
+        def ray_callable(input_dict: dict[str, Any]) -> dict[str, Any]:
+            doc = Document.from_row(input_dict)
+
+            try:
+                fields = self._field.split(".")
+                val = getattr(doc, fields[0])
+                if len(fields) > 1:
+                    for f in fields[1:]:
```
bsowell, line 41: Is the idea that index 0 is "properties" so you skip it? Might be worth asserting that just in case someone passes in something weird.

lib/sycamore/sycamore/transforms/sort.py
```diff
@@ -0,0 +1,57 @@
...
+        self._field = field
+        self._default_val = default_val
+
+    def execute(self, **kwargs) -> "Dataset":
+        # creates dataset
+        ds = self.child().execute(**kwargs)
+
+        # adds a "key" column containing desired field
+        map_fn = self.make_map_fn_sort()
+        ds = ds.map(map_fn)
```
bsowell, line 26: Are metadata documents included here? If so, how do they get sorted?

lib/sycamore/sycamore/transforms/sort.py
```diff
@@ -0,0 +1,57 @@
...
+
+        # sorts the dataset
+        ds = ds.sort("key", descending=self._descending)
```
bsowell, line 29: I think your implementation is fine, but I do have questions about what this guarantees downstream. Are all of our other transformations order preserving (I think probably)? From the docs, it looks like sort will materialize the dataset in Ray's objet store. Not sure what implications that has for subsequent operations and when it gets freed.

bsowell (commented):
Your code looks good, and I appreciate the tests. Have a few questions about how this works, but nothing that should block you.

lib/sycamore/sycamore/transforms/sort.py
tranade: Actually have not tested with metadata documents — how can this be done? I'm thinking that having a default value should deal with them though

lib/sycamore/sycamore/transforms/sort.py
tranade: Hmm that makes sense to consider, we can talk about it!

PR #554, opened 2024-07-18
Temporarily serialize partition calls with `ArynPartitioner` when running remotely

MarkLindblad:


lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -585,7 +585,8 @@ def __init__(
         self, child: Node, partitioner: Partitioner, table_extractor: Optional[TableExtractor] = None, **resource_args
     ):
         ops = []
-
+        if isinstance(partitioner, ArynPartitioner) and partitioner._use_partitioning_service:
+            resource_args["compute"] = ActorPoolStrategy(size=1)
```
MarkLindblad, line 589: Should this be `concurrency` instead?

MarkLindblad: Only `compute` seems to be supported: [`base.py:130`](https://github.com/aryn-ai/sycamore/blob/6aa288635e2fe3b6abbf48a1b9a066ec70abd134/lib/sycamore/sycamore/transforms/base.py#L130)

eric-anderson (approved):

lib/sycamore/sycamore/transforms/partition.py
alexaryn: compute is what we use in our code, but it's officially deprecated.  concurrency is preferred by Ray and our code seems to ignore it, but pass it through.  Frankly, I think we need to sweep through our code and convert to the new way.  For now, maybe check to see which one is specified, if any.  Maybe we just need to assert until we can properly support the recommended way.  Perhaps talk to Eric.

PR #555, opened 2024-07-18
Add Elasticsearch Reader

karansampath:
Adds a reader and testing for Elasticsearch using the Reader API


lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_reader.py
```diff
@@ -0,0 +1,70 @@
...
+class ElasticsearchReaderClientParams(BaseDBReader.ClientParams):
+    url: str
+    es_client_args: dict = field(default_factory=lambda: {})
+
+
+@dataclass
+class ElasticsearchReaderQueryParams(BaseDBReader.QueryParams):
+    index_name: str
+    query: Dict = field(default_factory=lambda: {"match_all": {}})
+    kwargs: Dict = field(default_factory=lambda: {})
```
HenryL27, line 19: what goes in here?

HenryL27 (approved):
Do we need to do anything special to support `scroll` queries?

karansampath:
> Do we need to do anything special to support `scroll` queries?

Good idea, I'll add a flag for this

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_reader.py
karansampath: https://elasticsearch-py.readthedocs.io/en/v8.14.0/api/elasticsearch.html#elasticsearch.Elasticsearch.search

Several additional parameters that users might want to specify but are not required. I could remove `query` as well (and include as part of `kwargs`) but since it is required felt better to separate it out

PR #557, opened 2024-07-19
Move Weaviate Reader to the Reader API

karansampath:
Moves the Weaviate reader to the Reader API and adds testing


lib/sycamore/sycamore/connectors/weaviate/weaviate_reader.py
```diff
@@ -0,0 +1,93 @@
...
+                        if hasattr(query, method_name):
+                            query = getattr(query, method_name)(**value)
+                        else:
+                            raise ValueError(f"Error: Method '{method_name}' not found in query object.")
+                query = query.objects  # type: ignore
+            else:
+                collection = list(
+                    self._client.collections.get(query_params.collection_name).iterator(include_vector=True)
+                )
+            results = WeaviateReaderQueryResponse(collection=collection, query_output=query)
```
HenryL27, line 63: I think this is genius, which makes me want to see a bunch of unit tests showing various cases

karansampath: Just added

HenryL27 (approved):
Thanks for doing this. I know it's a pain

PR #560, opened 2024-07-19
Creating the Aryn SDK 

mashah:
Created a small python function to partition PDFs using the Aryn Partitioning Service and return it into a JSON structure.


apps/arynsdk/python/client.py
```diff
@@ -0,0 +1,116 @@
...
+
+    return resp.json()
+
+
+def add_bbox_to_pdf():
+
+    return "Unimplemented"
+
+
+def test_partition_file():
```
HenryL27, line 74: Move tests to a separate file `test_client.py`. pytest will discover tests where 'test' is in the name of the file and 'test' is also in the name of the test function.

apps/arynsdk/python/client.py
```diff
@@ -0,0 +1,116 @@
...
+        raise requests.exceptions.HTTPError(
+            f"Error: status_code: {resp.status_code}, reason: {resp.text}", response=resp
+        )
+
+    return resp.json()
+
+
+def add_bbox_to_pdf():
+
+    return "Unimplemented"
```
HenryL27, line 71: nit: `raise NotImplementedError()`

apps/arynsdk/python/client.py
```diff
@@ -0,0 +1,116 @@
...
+
+# Replace with your token
+aryn_token = "YOUR_TOKEN"
+
+# Replace with test files
+simple_test_file = "PATH_TO_SIMPLE_FILE"
+large_test_file = "PATH_TO_LARGE_FILE"
+
+# URL for Aryn Partitioning Service (APS)
+aps_url = "https://api.aryn.cloud/v1/document/partition"
```
HenryL27, line 15: nit: `aps_url` -> `APS_URL`

apps/arynsdk/python/client.py
```diff
@@ -0,0 +1,116 @@
...
+
+    options_str = json.dumps(kwargs)
+
+    print(f"{options_str}")
+
+    files: Mapping = {"options": options_str.encode("utf-8"), "pdf": file}
+
+    http_header = {"Authorization": "Bearer {}".format(token)}
+
+    resp = requests.post(aps_url, files=files, headers=http_header)
```
HenryL27, line 59: maybe make aps_url overridable, so we could point it at our test env or something?

apps/arynsdk/python/client.py
```diff
@@ -0,0 +1,116 @@
...
+#        The defaults are what the Service will use, if not passed into the function
+
+
+def partition_file(file: BinaryIO, token: str, **kwargs) -> dict:
+
+    check_options(kwargs)
+
+    options_str = json.dumps(kwargs)
+
+    print(f"{options_str}")
```
HenryL27, line 53: drop or convert to `logging`

apps/arynsdk/python/client.py
```diff
@@ -0,0 +1,116 @@
...
+#        threshold:  value in [0.0 .. 1.0] to specify the cutoff for detecting bounding boxes.
+#                    default: 0.4
+#        use_ocr:    boolean to specify extracting text using an OCR model instead of extracting embedded text in PDF.
+#                    default: False
+#        extract_table_structure: boolean to specfy extracting tables and their structural content.
+#                    default: False
+#        extract_images: boolean that Mark doesn't know what it does.
+#                    default: False
+#
+#        The defaults are what the Service will use, if not passed into the function
```
HenryL27, line 44: stick this in a `'''blah'''` style docstring under the function signature

apps/arynsdk/python/client.py
```diff
@@ -0,0 +1,116 @@
...
+def test_large_file():
+
+    f = open(large_test_file, "rb")
+    my_resp = partition_file(f, aryn_token, threshold=0.35)
+    print("----------")
+    pprint(my_resp)
+
+
+# test_partition_file()
+# test_large_file()
```
HenryL27, line 116: drop

HenryL27 (commented):
hbd

mashah (commented):
I wouldn't put table_to_pandas as an option in the partition function. The partition function should return a Document structure. We should iterate through the Document structure to find the tables, and then call the to_pandas function on those tables.

Eventually, Sycamore should also separate out those structures into a "core" that both the arynsdk and Sycamore full depend on.

.github/workflows/draft_aryn-sdk_release.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install
+        run: poetry install
+        working-directory: lib/aryn-sdk
+      - name: build
+        run: poetry build
+        working-directory: lib/aryn-sdk
+      - name: Generate Changelog
+        run: |
+          git log --format="* %s by %aE" $(git describe --tags --abbrev=0 --match 'sdk-v*' \
+          --match 'v0.1.16' @^)..@ -- lib/aryn-sdk > ${{ github.workspace }}-CHANGELOG.txt
```
bsowell, line 32: Is this filtering the commits to only show those related to the SDK? That's kind of cool. I wonder if we should try to do that for the library release notes as well. I suppose this is also going to screw up the releases list on the right in Github, right? The version numbers will go back and forth between sdk releases and library releases?

lib/aryn-sdk/aryn_sdk/config.py
```diff
@@ -0,0 +1,31 @@
...
+        if self._aryn_api_key is not None:
+            return self._aryn_api_key
+        if _ARYN_API_KEY_ENV_VAR in os.environ:
+            return os.environ[_ARYN_API_KEY_ENV_VAR]
+        if Path(self._aryn_config_path).exists():
+            with open(self._aryn_config_path, "r") as f:
+                data = yaml.safe_load(f)
+                if "aryn_token" in data:
+                    return data["aryn_token"]
+        logging.warn(
```
bsowell, line 27: Should we throw an exception here? Is there ever a case where we want to move forward without an api key?

lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -0,0 +1,189 @@
...
+
+def partition_file(
+    file: BinaryIO,
+    aryn_api_key: Optional[str] = None,
+    aryn_config: ArynConfig = ArynConfig(),
+    threshold: Optional[float] = None,
+    use_ocr: bool = False,
+    extract_table_structure: bool = False,
+    extract_images: bool = False,
+    selected_pages: Optional[list[int]] = None,
```
bsowell, line 26: I think this needs to be a more complicated type. Something like: `Optional[list[Union[int, list[int]]]]`. Per Mark, you can pass in page selections like [1, [3,5], 17], which would process pages 1, 3-5 inclusive, and 17.

lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -0,0 +1,189 @@
...
+                columns=flattened_header if max_header_prefix_row >= 0 else None,
+            )
+            results.append((e, df))
+        else:
+            results.append((e, None))
+
+    return results
+
+
+def add_bbox_to_pdf():
```
bsowell, line 203: Is this a placeholder? Do you want to keep it the pypi release before it is implemented?

.github/workflows/draft_aryn-sdk_release.yml
HenryL27: yep on both counts I think. I couldn't quite find a way around the release list on gh... would be interested in what our monorepo gurus think although my prediction is they'll say "oh we had internal tooling for that"...
For now I'm just assuming that a/ the vast majority of commits will be relevant to lib/sycamore so if a few sdk ones sneak into the notes, that's ok, and b/ we'll just do sycamore-ai releases after aryn-sdk ones so it'll always show up as the most recent

lib/aryn-sdk/aryn_sdk/config.py
HenryL27: Maybe in some internal testing/local development of the model server it would be nice to run in less protected modes and forego this? Eh, can explode here, ok

lib/aryn-sdk/aryn_sdk/partition/partition.py
HenryL27: oh, neat

lib/aryn-sdk/aryn_sdk/partition/partition.py
HenryL27: dropped

.github/workflows/draft_aryn-sdk_release.yml
bsowell: Do we need to do a github release at all? I don't think the pypi release really depends on it, or at most it uses it as a way to get a specific commit hash, which we can do in a bunch of ways. 

I'm not /that/ worried about the gh release list, but if it's easy to avoid it might be worth it.

lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -0,0 +1,223 @@
...
+
+def _json_options(
+    threshold: Optional[float] = None,
+    use_ocr: bool = False,
+    ocr_images: bool = False,
+    extract_table_structure: bool = False,
+    extract_images: bool = False,
+    selected_pages: Optional[list[Union[list[int], int]]] = None,
+) -> str:
+    # isn't type-checking fun
```
bsowell, line 136: :) It sure is!

bsowell (approved):
Looks good. Thanks for all the changes. As noted, I think we could potentially skip the gh release process, but if that's too much work, I don't think it's a huge deal either way.

.github/workflows/draft_aryn-sdk_release.yml
eric-anderson: So it looks to me that github really wants single release per repo [1]. Every repo I can find has a single "latest" release. It appears that the standard pattern is multiple pypi packages from a single repository [2]. I also found a github action from google to do releases [3] along with a ton of unrelated pages I looked at.

So given that, I would not do a github release for the library and only do a pypi release.

[1] https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository
[2] https://stackoverflow.com/questions/54265944/python-multiple-packages-in-one-repository-or-one-package-per-repository
[3] https://github.com/googleapis/release-please-action

lib/aryn-sdk/aryn_sdk/partition/art.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    assert (
+        "elements" in partitioning_data
+    ), 'There are no "elements" in the provided partitioning data, so drawing them is impossible'
+    if isinstance(pdf_file, str):
+        pdf_file = Path(pdf_file)
+    if isinstance(pdf_file, PathLike):
+        with open(pdf_file, "rb") as f:
+            pdf_file = io.BytesIO(f.read())
+    images = pdf2image.convert_from_bytes(pdf_file.read())
```
bsowell, line 76: Was pdf2image already in the pyproject.toml? You should confirm that you don't have it from the dev dependency on sycamore, because that presumably won't come with pypi. 

This also requires poppler to be installed, which we should add to the docs.

bsowell (approved):

PR #561, opened 2024-07-21
Migrate from PyPDf2 to pypdf.

bsowell:
PyPDF2 has been deprecated and renamed to pypdf. We have also found that there are documents that pypdf can process that PyPDF2 cannot.


HenryL27 (approved):
lgtm

HenryL27:
IT failed to import pypdf

bsowell:
> IT failed to import pypdf

Oops. I updated all the poetry.lock files, but I missed the one at the root, which is of course the one that matters for the integ tests. Fixed now.

PR #564, opened 2024-07-22
Get logging working in ray

eric-anderson:
* Log the incremental output from the server
* Handle errors that happen partway through processing
* Move ray logging setup into default sycamore init
* for testing, add a sycamore.shutdown() to clean up sycamore.init()
* Add support for not loading the model when we're only going to call the server
* Add assertions to code to check for user error
* Switch partitioner to using logger. instead of logging. It's better style, it turned out to
  not be necessary.
* Remove caching from aryn_config. It caused problems when called with different parameters
  If we're going to re-introduce this we can use @cache which will respect parameters.
* Make SycamorePartitionerExample notebook runnable. The example failed because of missing
  files/directories.

### DIFF
--- a/examples/simple_duckdb.py
+++ b/examples/simple_duckdb.py
@@ -8,7 +8,6 @@ from sycamore.transforms.merge_elements import MarkedMerger
 from sycamore.transforms.partition import SycamorePartitioner
 from sycamore.transforms.extract_entity import OpenAIEntityExtractor
 from sycamore.transforms.embed import SentenceTransformerEmbedder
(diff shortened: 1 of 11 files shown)


MarkLindblad:
There is a leftover import of the old `ray_logging_setup` at [`test_partition.py:70`](https://github.com/aryn-ai/sycamore/blob/31672bfbb7bf26014eeaf265dcfec5c4ced8e9ce/lib/sycamore/sycamore/tests/integration/transforms/test_partition.py#L70)

MarkLindblad (approved):

PR #566, opened 2024-07-23
Migrate LUnA code to sycamore.query.

mdwelsh:
This is an initial migration of the `query-engine` code to `sycamore.query`.

There are a few cleanup things I need to follow up on, such as moving over some demo Jupyter notebooks and adding this to the Sycamore documentation, but this will unblock the team allowing us to do development work in the Sycamore tree directly.


lib/sycamore/sycamore/query/client.py
```diff
@@ -0,0 +1,255 @@
...
+        file_handler.setLevel(log_level)
+        structlog.configure_once(
+            processors=processors,
+            logger_factory=CustomLoggerFactory(file_handler),
+            wrapper_class=structlog.make_filtering_bound_logger(log_level),
+            cache_logger_on_first_use=True,
+        )
+
+
+class LunaClient:
```
bsowell, line 95: I know we want to get this in quickly, so I'm okay if this is in a separate PR, but presumably we should rename "Luna" to "SycamoreQuery" or something similar?

lib/sycamore/sycamore/query/execution/aggregation.py
```diff
@@ -0,0 +1,90 @@
...
+        new_doc = doc.to_row()
+        new_doc[field] = agg_val
+
+        return new_doc
+
+    return ray_callable
+
+
+def accumulate_row(doc, doc_bytes, fn):
+    # print(f"In accumulate_row: {doc} {doc_bytes}")
```
bsowell, line 31: There are a bunch of commented out prints in this file that I would probably remove.

lib/sycamore/sycamore/query/execution/aggregation.py
```diff
@@ -0,0 +1,90 @@
...
+    return fn(Document.deserialize(doc), Document.deserialize(doc_bytes["doc"])).serialize()
+
+
+def merge(x, y, fn):
+    # print(f"In merge: {Document.deserialize(x)} {Document.deserialize(y)}")
+    return fn(Document.deserialize(x), Document.deserialize(y)).serialize()
+
+
+# Most general version. Keeps everything as Documents. Lots of unnecessary serialization/deserializaiton.
+class DocSetAggregate(AggregateFn):
```
bsowell, line 41: Hey, I recognize this code :). I think this should eventually live in the core library, but I gather that's coming in a separate PR.

bsowell (commented):
I didn't go through all the code, but seems fairly safe to add to sycamore.

lib/sycamore/pyproject.toml
```diff
@@ -13,7 +13,7 @@ packages = [{ include = "sycamore" }]
 "Documentation" = "https://sycamore.readthedocs.io"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.12"
+python = ">=3.9,<3.9.7 || >3.9.7,<3.12"
```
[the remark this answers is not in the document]
mdwelsh: Apparently streamlit doesn't work with it. I have no idea why. Since I'm moving the streamlit app to apps/ this will go away and only apply to the streamlit app!

lib/sycamore/pyproject.toml
```diff
@@ -79,6 +81,11 @@ nbmake = "^1.4.5"
 jupyterlab = "^4.0.11"
 jupyter-lsp = "^2.2.2"
 ipywidgets = "^8.1.0"
+matplotlib = "^3.9.1"
+rich = "^13.7.1"
+streamlit = "^1.36.0"
+streamlit-ace = "^0.1.1"
+streamlit-agraph = "^0.0.45"
```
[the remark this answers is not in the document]
mdwelsh: Yes! Good idea. Will do.

lib/sycamore/sycamore/query/client.py
mdwelsh: Yes, that's the plan. I'll probably do a quick follow up PR.

lib/sycamore/sycamore/query/execution/aggregation.py
mdwelsh: Yep.

mdwelsh:
OK, this passes tests and mypy (on my local machine, will look at CI). Ready for another look.

baitsguy (approved):

PR #568, opened 2024-07-23
Add Image to JSON cache

pparmar30:


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -74,7 +74,7 @@ class ArynPDFPartitioner:
     ArynPartitioner class.
     """
 
-    def __init__(self, model_name_or_path=ARYN_DETR_MODEL, device=None):
+    def __init__(self, model_name_or_path, device=None, cache: Optional[Cache] = None):
```
alexaryn, line 77: Why drop the default parameter value?  Doesn't seem to have much to do with caching.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -640,19 +643,14 @@ def __init__(self, model_name_or_path, device=None):
...
-        results = []
-        inputs = self.processor(images=images, return_tensors="pt").to(self._get_device())
-        outputs = self.model(**inputs)
-        target_sizes = torch.tensor([image.size[::-1] for image in images])
-        results.extend(
-            self.processor.post_process_object_detection(outputs, target_sizes=target_sizes, threshold=threshold)
-        )
+    def infer(self, images: List[Image.Image], threshold: float, use_cache: bool = False) -> List[List[Element]]:
+        if use_cache and self.cache:
+            print("Checking Cache for Image to JSON")
```
alexaryn, line 648: May not want to keep this print and the one after.  There are already messages for both the hit and miss cases.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -666,8 +664,52 @@ def infer(self, images: List[Image.Image], threshold: float) -> List[List[Elemen
                 )
                 elements.append(element)
             batched_results.append(elements)
+            if self.cache:
+                key = Cache.get_hash_key(image.tobytes())
```
alexaryn, line 668: Should we hash-in some representation of the software version or processing options?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -666,8 +664,52 @@ def infer(self, images: List[Image.Image], threshold: float) -> List[List[Elemen
...
 
+    def _get_cached_inference(self, images: List[Image.Image], threshold: float) -> list:
+        results = []
+        uncached_images = []
+        uncached_indices = []
+
+        # First, check the cache for each image
+        for index, image in enumerate(images):
+            key = Cache.get_hash_key(image.tobytes())
+            cached_image = self.cache.get(key)
```
alexaryn, line 681: Is this really an image?  I thought it would be a layout.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -666,8 +664,52 @@ def infer(self, images: List[Image.Image], threshold: float) -> List[List[Elemen
...
+            if cached_image:
+                print("Cache hit. Returning the result from cache")
```
alexaryn, line 683: What would be nicer here would be a cache hit-rate from self.cache.  We know what a cache hit is.  We don't need explanation that we're using the result from the cache.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -666,8 +664,52 @@ def infer(self, images: List[Image.Image], threshold: float) -> List[List[Elemen
...
+                results.append(cached_image)
+            else:
+                uncached_images.append(image)
+                uncached_indices.append(index)
```
alexaryn, line 687: Rather than these parallel data structures, I'd use a list of tuples: (index, image).  That said, I don't think you even need that.  Just run through the results array and anywhere you have a None, you need to do the actual work.  You already have the images array as a parameter, and the indices will be the same in the two arrays, right?

alexaryn (commented):
I think we can make this a little less complicated.

lib/sycamore/sycamore/transforms/detr_partitioner.py
pparmar30: it was a mistake

lib/sycamore/sycamore/transforms/detr_partitioner.py
pparmar30: updated

lib/sycamore/sycamore/transforms/detr_partitioner.py
pparmar30: the model just uses threshold! So hashing-in threshold for now.

lib/sycamore/sycamore/transforms/detr_partitioner.py
pparmar30: updated

lib/sycamore/sycamore/transforms/detr_partitioner.py
pparmar30: I want to re-use the logic in _get_uncached_inference! Also, if the batch size changes to be always 1 (As per Eric's recommendation) this logic will get simplified.

lib/sycamore/sycamore/transforms/detr_partitioner.py
pparmar30: updated

lib/sycamore/sycamore/transforms/detr_partitioner.py
alexaryn: I see it.

lib/sycamore/sycamore/utils/cache.py
```diff
@@ -11,23 +11,44 @@
 BLOCK_SIZE = 1048576  # 1 MiB
 
 
+class HashWrapper:
```
alexaryn, line 14: I'm not sure what value is added by this wrapper.  Maybe we could get by with a string constant set to "sha256" and callers can just call hashlib.new(STRING_CONSTANT) themselves.

lib/sycamore/sycamore/utils/cache.py
```diff
@@ -39,10 +60,15 @@ def get_hash_context_file(file_path: str, hash_context: Optional[hashlib._Hash]
 
 class DiskCache(Cache):
     def __init__(self, cache_loc: str):
+        super().__init__()
         self._cache = diskcache.Cache(directory=cache_loc)
 
     def get(self, hash_key: str):
-        return self._cache.get(hash_key)
+        v = self._cache.get(hash_key)
+        if v:
```
alexaryn, line 68: I'd use `is not None` here, just in case the hashed content is a zero-byte string.

alexaryn (commented):
It keeps getting better.

lib/sycamore/sycamore/utils/cache.py
pparmar30: It helps passing around the hash ctx as Python/mypy/IDE does not like me accessing _Hash from hashlib

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -368,7 +367,8 @@ def _partition_pdf_sequenced(
         with LogTime("all_batches"):
             for i, batch in enumerate(batches):
                 with LogTime(f"infer_one_batch {i}/{len(images) / batch_size}"):
-                    deformable_layout += self.model.infer(batch, threshold)
+                    assert self.model is not None
```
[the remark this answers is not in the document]
pparmar30: Yes, it is.

lib/sycamore/sycamore/utils/cache.py
pparmar30: updated

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -666,8 +662,53 @@ def infer(self, images: List[Image.Image], threshold: float) -> List[List[Elemen
...
         return batched_results
 
+    def _get_cached_inference(self, images: List[Image.Image], threshold: float) -> list:
+        results = []
+        uncached_images = []
+        uncached_indices = []
+
+        # First, check the cache for each image
+        for index, image in enumerate(images):
+            key = Cache.get_hash_context(image.tobytes()).update(threshold).hexdigest()
```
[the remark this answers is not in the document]
pparmar30: updated

lib/sycamore/sycamore/utils/cache.py
alexaryn: Got it.  Maybe a comment on the class saying that would help.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -666,8 +662,55 @@ def infer(self, images: List[Image.Image], threshold: float) -> List[List[Elemen
...
+            self.processor.post_process_object_detection(outputs, target_sizes=target_sizes, threshold=threshold)
+        )
+        for result in results:
+            result["scores"] = result["scores"].tolist()
+            result["labels"] = result["labels"].tolist()
+            result["boxes"] = result["boxes"].tolist()
+        return results
+
+    def _get_hash_key(self, image: Image.Image, threshold: float) -> str:
+        return Cache.get_hash_context([image.tobytes(), str(threshold).encode()]).hexdigest()
```
alexaryn, line 712: For threshold, I'd want `f"{threshold:.6f}".encode()` just to be sure.  Also, we should put something like `_VERSION = "0.2024.07.24"` at the top of the file and hash it in here also.

alexaryn (approved):
Trusting you to make final round of changes.

PR #571, opened 2024-07-24
Chunker

dtecuci:


lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -135,7 +135,151 @@ def merge(self, elt1: Element, elt2: Element) -> Element:
...
+        # 'image' + 'text' = 'image+text'
+        # 'image+text' + 'text' = 'image+text'
+        # 'Section-header' + 'table' = 'Section-header+table'
+
+        if (elt1.type == "Image" or elt1.type == "Image+Text") and elt2.type == "Text":
+            new_elt.type = "Image+Text"
+        elif elt1.type == "Text" and elt2.type == "Text":
+            new_elt.type = "Text"
+        elif elt1.type == "Section-header" and elt2.type == "table":
+            new_elt.type = "Section-header+table"
```
bsowell, line 214: I'm wondering a little bit about changing the element types here and using a generic Element. It means we lose all of the utility methods in TableElement/Table an ImageElement. Another option might be to add a "title" field to a TableElement (it already has a caption) and add some description field to an ImageElement. 

I won't be a stickler here -- this will obviously work for a lot of use cases, just trying to think through how element types should work going forward.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -135,7 +135,151 @@ def merge(self, elt1: Element, elt2: Element) -> Element:
...
+        else:
+            new_elt.type = "????"
+         
+        # Merge binary representations by concatenation
```
bsowell, line 218: Is this ever useful? I confess I don't have a better way to merge binary in general, but I can't imagine this resulting in something useable.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -135,7 +135,151 @@ def merge(self, elt1: Element, elt2: Element) -> Element:
...
+
+    def preprocess_element(self, element: Element) -> Element:
+        if element.type == 'Image' and "summary" in element.properties:
+            element.data["token_count"] = len(self.tokenizer.tokenize(element.properties['summary'] or ""))
+        else:
+            element.data["token_count"] = len(self.tokenizer.tokenize(element.text_representation or ""))
+        return element
+
+    def postprocess_element(self, element: Element) -> Element:
+        del element.data["token_count"]
```
baitsguy, line 152: I know we haven't previously but I wonder if keeping token_count around in elements is actually useful, it'll help us to cost estimation for running LLMs across a set of elements. Thoughts? @bsowell too

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -135,7 +135,151 @@ def merge(self, elt1: Element, elt2: Element) -> Element:
...
+        return element
+
+    def should_merge(self, element1: Element, element2: Element) -> bool:
+        # deal with empty elements
+        if "page_number" not in element1.properties or "page_number" not in element2.properties or element1.type == None or element2.type == None:
```
baitsguy, line 157: what are cases where an elements type is none? Is that an invalid state/should we fail?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -135,7 +135,151 @@ def merge(self, elt1: Element, elt2: Element) -> Element:
...
+
+        # DO NOT MERGE across pages
+        if not self.merge_across_pages and element1.properties["page_number"] != element2.properties["page_number"]:
+            return False
+
+        if element1.data["token_count"] + 1 + element2.data["token_count"] > self.max_tokens:
+            return False
+        
+        # MERGE adjacent 'text' elements (but not across pages - see above)
+        if ((element1 != None) and (element1.type == 'Text') and
```
baitsguy, line 168: we should just fail/return early if the element is None right?

lib/sycamore/sycamore/transforms/merge_elements.py
baitsguy: We should at least keep the table data accessible when merging elements. I'd also vote for Section-header+Table just being an augmented Table object. We can update the text_representation and store the title as a field

dtecuci: Agreed that more though is needed in how the new element types are named and built. 
What I wanted to show with these names is that the merged elements are no longer just a Table, or an Image.

lib/sycamore/sycamore/transforms/merge_elements.py
dtecuci: Yes, that makes sense. I guess the only case where this might be marginally useful is where the binary_representation is actual text (when reading from a file), but even then the text_representation should be populated. I will remove the concatenation. Will replace with keeping the binary of the Image element.

lib/sycamore/sycamore/transforms/merge_elements.py
dtecuci: Some of the previous unit tests had empty elements. I agree that they should not be there, in principle.

lib/sycamore/sycamore/transforms/merge_elements.py
dtecuci: Yes, this is made superfluous by the test in line 157. Will remove

baitsguy (approved):
Since this is a new merger I think we can merge it and evolve if we find issues

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -135,7 +135,151 @@ def merge(self, elt1: Element, elt2: Element) -> Element:
 
         return new_elt
 
+class GreedySectionMerger(ElementMerger):
```
baitsguy, line 138: Should probably a small description about the general logic

PR #572, opened 2024-07-24
Add LLM Query as a Transform

karansampath:
Adds querying an OpenAI LLM using a user defined prompt as a transform

### DIFF
--- a/lib/sycamore/sycamore/transforms/llm_query.py
+++ b/lib/sycamore/sycamore/transforms/llm_query.py
@@ -4,7 +4,6 @@ from sycamore.data import Element, Document
 from sycamore.plan_nodes import NonCPUUser, NonGPUUser, Node
 from sycamore.llms import LLM
 from sycamore.transforms.map import Map
-from sycamore.transforms.augment_text import JinjaTextAugmentor
 from sycamore.utils.time_trace import timetrace
 from jinja2.sandbox import SandboxedEnvironment
 
@@ -17,8 +16,9 @@ class LLMTextQueryAgent:
         prompt: A prompt to be passed into the underlying LLM execution engine
         llm: (LLM) An instance of the LLM class to be pass into the user
         output_property: (Optional) The output property to add results in. Defaults to 'llm_response'
-        format_kwargs: (Optional) Formatting arguments passed in to define the prompt
-        number_of_elements: (Optional) if not executed per element
+        format_kwargs: (Optional) Formatting arguments passed in to define the prompt, uses a Jinja Sandbox
+        number_of_elements: (Optional) Parameter to either limit the number of elements or to add an llm response to the
+        entire document using a prefix of elements
         llm_kwargs: (Optional) LLM keyword argument for the underlying execution engine
         per_element: (Optional) Whether to execute the call per each element or on the Document itself. Defaults to
         True.
@@ -55,9 +55,9 @@ class LLMTextQueryAgent:
 
     def execute_query(self, document: Document) -> Document:
         if self._per_element:
-            elements = []
+            elements = document.elements
             for idx, element in enumerate(document.elements):
-                elements.append(self._query_text_object(element))
+                elements[idx] = self._query_text_object(element)
                 if self._number_of_elements and idx >= self._number_of_elements:
                     break
             document.elements = elements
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -0,0 +1,94 @@
...
+        self._prompt = prompt
+        self._output_property = output_property
+        self._llm_kwargs = llm_kwargs
+        self._per_element = per_element
+
+    def execute_query(self, document: Document) -> Document:
+        if self._per_element:
+            elements = []
+            for element in document.elements:
+                elements.append(self._summarize_text_element(element))
```
dtecuci, line 67: Does this mean that when per_element=True we can only execute the llm_query on the summaries of elements?

dtecuci: it doesn't look like summarize_text_element does summarization, though.

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -0,0 +1,94 @@
...
+                .partition(partitioner=UnstructuredPdfPartitioner())
+                .llm_query(query_agent=llm_query_agent)
+    """
+
+    model = OpenAIModels.GPT_4O
+
+    def __init__(
+        self,
+        prompt: str,
+        openai_model: Optional[OpenAI] = None,
```
bsowell, line 48: I think you should just take an LLM here rather than an OpenAI model/client wrapper.

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -0,0 +1,94 @@
...
+            pdf_docset = context.read.binary(paths, binary_format="pdf")
+                .partition(partitioner=UnstructuredPdfPartitioner())
+                .llm_query(query_agent=llm_query_agent)
+    """
+
+    model = OpenAIModels.GPT_4O
+
+    def __init__(
+        self,
+        prompt: str,
```
bsowell, line 47: One common pattern is to have a prompt that is parameterized based on the element text or properties. Consider a prompt like [this](https://github.com/aryn-ai/pimco-scripts/blob/b8e206f9560d1538c29af5e5b9b7b7c8a5901193/pimco_scripts/questions.py#L349), which is parameterized not only by the text, but also the `company`, `quarter`, and `year`. We should think about how to support that here. Maybe pass in the properties dict + text representation to str.format? An alternative would be to use something like Jinja (which @HenryL27 used [here](https://github.com/aryn-ai/sycamore/blob/de50a997047aeccf4acfd88aa5cefd27192ea243/lib/sycamore/sycamore/transforms/augment_text.py#L54)). I think a good goal would be to support all of the llm calls in that first repo without custom llm code.

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -0,0 +1,94 @@
...
+        self._per_element = per_element
+
+    def execute_query(self, document: Document) -> Document:
+        if self._per_element:
+            elements = []
+            for element in document.elements:
+                elements.append(self._summarize_text_element(element))
+            document.elements = elements
+        else:
+            if document.text_representation:
```
bsowell, line 70: I wonder if there are cases where you need to support passing a prefix of the elements (like we do in our LLM transforms today). When we process PDFs, the document won't have a useful text_representation, but we might still want to attach the llm response to the document rather than each element.

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -0,0 +1,94 @@
...
+
+from sycamore.data import Element, Document
+from sycamore.llms.openai import OpenAIModels
+from sycamore.plan_nodes import NonCPUUser, NonGPUUser, Node
+from sycamore.llms import OpenAI, OpenAIClientWrapper
+from sycamore.transforms.map import Map
+from sycamore.utils.time_trace import timetrace
+
+
+class LLMQueryAgent(ABC):
```
bsowell, line 12: What's the rationale for this interface? Do we expect multiple implementations?

lib/sycamore/sycamore/transforms/llm_query.py
karansampath: No, it was executing a query per element, renamed it to a more appropriate version

lib/sycamore/sycamore/transforms/llm_query.py
karansampath: Discussed offline, can add it in later if we add more implementations. Have removed it for now

lib/sycamore/sycamore/transforms/llm_query.py
karansampath: Good point, added support for this

dtecuci (approved):

PR #573, opened 2024-07-24
Update cache manager

pparmar30:


lib/sycamore/sycamore/utils/cache.py
```diff
@@ -16,10 +18,24 @@ def set(self, hash_key: str, hash_value):
         pass
 
     @staticmethod
-    def get_hash_key(data: bytes) -> str:
+    def get_hash_bytes(data: bytes) -> hashlib.sha256:
```
alexaryn, line 21: Typically, this is a called a "hash context" and not "bytes".  The reason is that it's much bigger than the final hash.  It also has the ability to be updated incrementally, and can be copied for cases where common prefixes are involved.

Bytes from digest() might be  0x30, 0x31, 0x32, 0x33, 0x34
String from hexdigest() might be "30313234" (using 2 characters per byte)
The SHA-256 context is around 104 bytes

lib/sycamore/sycamore/utils/cache.py
```diff
@@ -16,10 +18,24 @@ def set(self, hash_key: str, hash_value):
...
         hash_sha256.update(data)
-        return hash_sha256.hexdigest()
+        return hash_sha256
+
+    @staticmethod
+    def get_hash_file(file_path: str) -> hashlib.sha256:
+        sha = hashlib.sha256()
+        with open(file_path, "rb") as file:
+            file_buffer = file.read(BLOCK_SIZE)
+            while len(file_buffer) > 0:
```
alexaryn, line 31: Can just say: while file_buffer:

I don't like to repeat code, so I usually do:
```
while True:
  file_buffer = file.read(BLOCK_SIZE)
  if not file_buffer:
    break
  sha.update(file_buffer)
```

lib/sycamore/sycamore/utils/cache.py
```diff
@@ -16,10 +18,24 @@ def set(self, hash_key: str, hash_value):
...
     @staticmethod
-    def get_hash_key(data: bytes) -> str:
+    def get_hash_bytes(data: bytes) -> hashlib.sha256:
         hash_sha256 = hashlib.sha256()
         hash_sha256.update(data)
-        return hash_sha256.hexdigest()
+        return hash_sha256
+
+    @staticmethod
+    def get_hash_file(file_path: str) -> hashlib.sha256:
```
alexaryn, line 27: This function may be even more useful if you can pass in the hash context to be used.

lib/sycamore/sycamore/utils/cache.py
```diff
@@ -16,10 +18,24 @@ def set(self, hash_key: str, hash_value):
...
+        sha = hashlib.sha256()
+        with open(file_path, "rb") as file:
+            file_buffer = file.read(BLOCK_SIZE)
+            while len(file_buffer) > 0:
+                sha.update(file_buffer)
+                file_buffer = file.read(BLOCK_SIZE)
+        return sha
+
+    @staticmethod
+    def update_hash(existing_hash: hashlib.sha256, new_content: Any) -> hashlib.sha256:
```
alexaryn, line 37: It's unclear to me what the value of this function is?  Given it's signature, it needs to be passed a SHA-256 context.  There's no abstraction.  Why shouldn't a caller just call hash_ctx.update(buf) themselves?

lib/sycamore/sycamore/utils/cache.py
```diff
@@ -16,10 +18,24 @@ def set(self, hash_key: str, hash_value):
...
+        return existing_hash.update(new_content)
```
alexaryn, line 38: I think this returns None.

alexaryn (commented):
I like the direction.

lib/sycamore/sycamore/utils/cache.py
```diff
@@ -1,12 +1,15 @@
...
 import json
 import time
-from typing import Any
+from typing import Any, Optional
 
 import boto3
 import diskcache
 from botocore.exceptions import ClientError
 
+BLOCK_SIZE = 1000000  # 1 MB
```
alexaryn, line 11: I should have said 1MiB.  That would be 1048576.  We want a clean power of two because disk sectors are 512 bytes, memory pages are 4096 bytes, and the larger structures are built upon them.

lib/sycamore/sycamore/utils/cache.py
```diff
@@ -16,10 +19,22 @@ def set(self, hash_key: str, hash_value):
...
-    def get_hash_key(data: bytes) -> str:
+    def get_hash_context(data: bytes) -> hashlib._Hash:
         hash_sha256 = hashlib.sha256()
         hash_sha256.update(data)
-        return hash_sha256.hexdigest()
+        return hash_sha256
+
+    @staticmethod
+    def get_hash_context_file(file_path: str, hash_context: Optional[hashlib._Hash] = None) -> hashlib._Hash:
+        if not hash_context:
```
alexaryn, line 29: Nice.

alexaryn (approved):
Cool!  Just tweak the block size.

PR #574, opened 2024-07-24
Run Sycamore integ tests only on Python 3.9.

bsowell:
This should provide roughly the same coverage, but be significantly cheaper and faster to execute. We run with Python 3.9, the minimum python version we support, to catch cases where we have inadvertently added dependencies on features added in later versions.


HenryL27 (approved):
lgtm

PR #577, opened 2024-07-24
Add code to convert pdf to images via files.

eric-anderson:
Will remove the old code once everything is switched over.

### DIFF
--- a/lib/sycamore/sycamore/utils/pdf.py
+++ b/lib/sycamore/sycamore/utils/pdf.py
@@ -2,6 +2,7 @@ import logging
 
 from io import BytesIO
 from PIL import Image
+from pathlib import Path
 from queue import Queue
 from subprocess import PIPE, Popen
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -119,3 +123,112 @@ def convert_from_path_streamed_batched(filename: str, batch_size: int) -> Genera
 
     if len(batch) > 0:
         yield batch
+
+
+def pdf_to_image_files(pdf_path: str, file_dir: Path) -> Generator[Path, None, None]:
+    """Writes the files (streamed) into file_dir.  Caller is responsible for calling
```
MarkLindblad, line 129: Should we have a similar note warning that the Aryn Partitioning Service will call this function?

MarkLindblad (approved):

PR #578, opened 2024-07-24
Add shared lock over pytorch file baton lock file

MarkLindblad:


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -627,12 +629,14 @@ def __init__(self, model_name_or_path, device=None):
         self.device = device
         self._model_name_or_path = model_name_or_path
 
-        from transformers import AutoImageProcessor, DeformableDetrForObjectDetection
+        lock = fasteners.InterProcessLock(_DETR_LOCK_FILE)
+        with lock:
```
alexaryn, line 633: It's idiomatic to have:
```
with fasteners.InterProcessLock(_DETR_LOCK_FILE):
```
because you don't actually reference `lock`

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -627,12 +629,14 @@ def __init__(self, model_name_or_path, device=None):
...
+            from transformers import AutoImageProcessor, DeformableDetrForObjectDetection
 
-        LogTime("loading_model", point=True)
-        with LogTime("load_model", log_start=True):
-            self.processor = AutoImageProcessor.from_pretrained(model_name_or_path)
-            self.model = DeformableDetrForObjectDetection.from_pretrained(model_name_or_path).to(self._get_device())
+            LogTime("loading_model", point=True)
+            with LogTime("load_model", log_start=True):
+                self.processor = AutoImageProcessor.from_pretrained(model_name_or_path)
```
alexaryn, line 638: I'm not sure where, but you should have code that does something like os.walk(f"{os.environ['HOME']}/.cache/torch_extensions") and deletes files called 'lock'

MarkLindblad: Done.

lib/sycamore/sycamore/transforms/detr_partitioner.py
MarkLindblad: Done

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -34,9 +36,10 @@
 from sycamore.utils.memory_debugging import display_top, gc_tensor_dump
 from sycamore.utils.pdf import convert_from_path_streamed_batched
 from sycamore.utils.time_trace import LogTime, timetrace
-
+from sycamore.utils.pytorch_dir import get_pytorch_build_directory
 
 logger = logging.getLogger(__name__)
+_DETR_LOCK_FILE = "~/.cache/Aryn-Detr.lock"
```
alexaryn, line 42: Beware that not everything interprets tilde (~) as $HOME.  It's safer to use os.environ["HOME"] or to pick a fixed path like in /tmp.  If you really want to do this right, use pwd.getpwuid(os.getuid()).pw_dir.  I'd convert this string constant to a suffix and put the home-dir logic below.

MarkLindblad: Done.

alexaryn (approved):
lgtm

PR #583, opened 2024-07-25
Add caching and fix query logic.

mdwelsh:
For some reason I'm not getting cache hits, which is odd, but I don't think that has to do with the streamlit app.

### DIFF
--- a/apps/query-ui/queryui/queryui.py
+++ b/apps/query-ui/queryui/queryui.py
@@ -113,7 +113,7 @@ def run_query(query: str, index: str, plan_only: bool, do_trace: bool, use_cache
     if not plan_only:
         with st.spinner("Running query..."):
             _, result = client.run_plan(plan)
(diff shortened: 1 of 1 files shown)


tranade (approved):

apps/query-ui/queryui/queryui.py
```diff
@@ -90,13 +93,17 @@ def show_dag(plan: LogicalPlan):
     agraph(nodes=nodes, edges=edges, config=config)
 
 
-def run_query(query: str, index: str, plan_only: bool, do_trace: bool):
+def run_query(query: str, index: str, plan_only: bool, do_trace: bool, use_cache: bool):
```
tranade, line 96: UI change suggestion in this block: change `st.text_area("Result", result, height=400)` to
```
st.subheader('Result', divider='rainbow')
st.success(result)
```

PR #584, opened 2024-07-25
Add Helicone support.

mdwelsh:
This PR adds support for an environment variable, `SYCAMORE_HELICONE_API_KEY`, which, when set, will cause the OpenAI client to use Helicone (https://helicone.ai) to issue LLM calls. This allows for greater visibility and debugging, and things like cost tracking etc. It is completely optional.

To use this, one must set up a Helicone account, and get a Helicone API key, and set the env var to this value.

### DIFF
--- a/lib/sycamore/sycamore/llms/openai.py
+++ b/lib/sycamore/sycamore/llms/openai.py
@@ -65,6 +65,7 @@ class OpenAIClientWrapper:
         api_version: Optional[str] = None,
         azure_ad_token: Optional[str] = None,
         azure_ad_token_provider: Optional[AzureADTokenProvider] = None,
+        disable_helicone: Optional[bool] = None,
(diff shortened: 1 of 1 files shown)


HenryL27 (commented):
I think there should be a way to turn this off programmatically (other than dropping the env var). Like if someone has this in dotfiles that they export without thinking and sticks it in a box in a restricted env then there could be issues

mdwelsh:
> I think there should be a way to turn this off programmatically (other than dropping the env var). Like if someone has this in dotfiles that they export without thinking and sticks it in a box in a restricted env then there could be issues

Fair enough, though I think there are a lot of configuration params already here. I added another param `disable_helicone` to disable this.

tranade (approved):

bsowell (approved):

HenryL27 (approved):

PR #587, opened 2024-07-25
Add support for subtasks

aanya-p:
Changed the EvaluationPipeline to be able to answer a question using subtasks loaded in from an external file


lib/sycamore/sycamore/evaluation/pipeline.py
```diff
@@ -36,46 +41,60 @@ def __init__(
...
-        query_body["query"]["hybrid"]["queries"][0] = hybrid_query_match
-        hybrid_query_neural = query_body["query"]["hybrid"]["queries"][1]
-        hybrid_query_neural["neural"]["embedding"]["filter"] = {
-            "bool": {"must": [{"match_phrase": {k: filters[k]}} for k in filters]}
-        }
-        query_body["query"]["hybrid"]["queries"][1] = hybrid_query_neural
+        hybrid_query_match = query_body["query"]["knn"]["embedding"]["filter"]["bool"]["must"]
+        for key, val in filters.items():
+            hybrid_query_match.append({"match_phrase": {key: val}})
+        query_body["query"]["knn"]["embedding"]["filter"]["bool"]["must"] = hybrid_query_match
```
HenryL27, line 51: wait what happens if it's a hybrid query?

lib/sycamore/sycamore/evaluation/pipeline.py
```diff
@@ -36,46 +41,60 @@ def __init__(
...
         return query_body
 
     def _build_opensearch_query(self, doc: Document) -> Document:
         assert doc.type == "EvaluationDataPoint"
         query = OpenSearchQuery(doc)
         query["index"] = self._index
+
+        if self._subtask_docs and doc.properties["subtasks_reqd"]:
+            filtered_docs = [
+                st_doc for st_doc in self._subtask_docs if st_doc.parent_id == doc["raw"]["financebench_id"]
```
HenryL27, line 61: maybe make "financebench_id" parametrizable or more general?

HenryL27: hmm - `self._subtask_docs` is a DocSet, right? I don't think you can just iterate over a DocSet like this. Perhaps you want `filtered_docs = self._subtask_docs.filter(lambda st_doc: st_doc.parent_id == doc['raw']['financebench_id']`?

lib/sycamore/sycamore/evaluation/subtasks.py
```diff
@@ -0,0 +1,139 @@
...
+
+openai_llm = OpenAI(OpenAIModels.GPT_3_5_TURBO.value)
+prompt = TaskIdentifierZeroShotGuidancePrompt()
+
+
+class SubtaskExecutor:
+    def __init__(
+        self,
+        filepath: Optional[str],
+        subtask_data: dict[str, Any],
```
HenryL27, line 20: make Optional[dict[str, Any]] I think

lib/sycamore/sycamore/evaluation/subtasks.py
```diff
@@ -0,0 +1,139 @@
...
+from sycamore.data.element import Element
+from sycamore.docset import DocSet
+from sycamore.evaluation.data import EvaluationDataPoint
+from sycamore.llms.openai import OpenAI, OpenAIModels
+from sycamore.llms.prompts.default_prompts import TaskIdentifierZeroShotGuidancePrompt
+from sycamore.transforms.embed import Embedder, SentenceTransformerEmbedder
+from sycamore.transforms.query import QueryExecutor
+
+openai_llm = OpenAI(OpenAIModels.GPT_3_5_TURBO.value)
+prompt = TaskIdentifierZeroShotGuidancePrompt()
```
HenryL27, line 13: These should be overrideable - suggested change is put these in the `__init__` args as default values and save them as instance variables.

lib/sycamore/sycamore/evaluation/subtasks.py
```diff
@@ -0,0 +1,139 @@
...
+    def _get_results(self, element: Element) -> Element:
+        return self._query_executor.query(element)
+
+    def execute(self, ds: DocSet) -> DocSet:
+        # Convert docset of all questions to a docset of formulas
+        formula_ds = ds.flat_map(self._get_formulas)
+        subtasks_ds = formula_ds.map(self._get_subtasks)
+        results_ds = subtasks_ds.map_elements(self._get_results)
+
+        return results_ds.take_all()
```
HenryL27, line 139: this execute function says it returns a docset - results_ds is a docset... results_ds.take_all() is a list of documents. What should we be doing here?

lib/sycamore/sycamore/evaluation/subtasks.py
```diff
@@ -0,0 +1,139 @@
...
+            term for term in self._subtask_data["subtask_instructions"].keys() if term in document.text_representation
+        ]
+
+        for term in terms:
+            subtask = self._subtask_data["subtask_instructions"][term].format(**document.properties["subtask_filters"])
+            sub_doc = OpenSearchQuery()
+            sub_doc.index = self._index
+
+            subtask += "Return only the code " + term + " alongside the amount found and no additional information."
+            qn_embedding = self._embedder.generate_text_embedding(subtask)
```
HenryL27, line 91: what if self._embedder is None?

HenryL27 (commented):
Hey where does the EvaluationPipeline diver to the SubtaskExecutor again? Trying to re-figure out how these things play with each other

lib/sycamore/sycamore/evaluation/pipeline.py
aanya-p: fixed the financebench_id

`self._subtask_docs` is a list of Documents (the type returned by the SubtaskExecutor `execute` needs to be corrected). filtering a DocSet would require a `take_all` for every question, so it seemed cleaner to have just one `take_all` and execute all of the subtasks before they are passed into the EvalPipeline.

lib/sycamore/sycamore/evaluation/pipeline.py
aanya-p: replaced knn query with hybrid

this is the workflow:

DocSet of questions --> DocSet of all formulas to be solved --> DocSet of formulas: elements of each document are subtasks --> DocSet of formulas: elements of each document are subtask answers

results of subtask execution passed to the EvaluationPipeline
in EvaluationPipeline's `_build_opensearch_query`, relevant results of subtask execution are added to original question

HenryL27 (approved):
ship it!

PR #589, opened 2024-07-26
Fix long line issue in partitioner client

eric-anderson:
The default requests code can be quadratic in the data size if it finds a single long line without a newline because it is doing the standard append to a string trick.  Switch over to chunking based on raw data and explicitly handle splitting by lines for the early status output.

### DIFF
--- a/lib/sycamore/sycamore/transforms/detr_partitioner.py
+++ b/lib/sycamore/sycamore/transforms/detr_partitioner.py
@@ -221,23 +221,46 @@ class ArynPDFPartitioner:
 
         logger.debug(f"ArynPartitioner POSTing to {aryn_partitioner_address} with files={files}")
         response = requests.post(aryn_partitioner_address, files=files, headers=header, stream=True)
-        lines = []
+        content = []
         in_status = False
-        for line in response.iter_lines():
-            if line:
-                lines.append(line)
+        in_bulk = False
+        partial_line = b""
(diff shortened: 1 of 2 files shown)


MarkLindblad:
Update the MockResponses in `test_aryn_partitioner.py` to include an `iter_content`

```
FAILED sycamore/tests/unit/transforms/test_aryn_partitioner.py::TestArynPDFPartitioner::test_partition - AttributeError: 'MockResponseNoTables' object has no attribute 'iter_content'
FAILED sycamore/tests/unit/transforms/test_aryn_partitioner.py::TestArynPDFPartitioner::test_partition_extract_table_structure - AttributeError: 'MockResponseTables' object has no attribute 'iter_content'
```

MarkLindblad (approved):
Looks good.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -221,23 +221,46 @@ def _call_remote_partitioner(
...
+            if not part:
+                continue
+
+            content.append(part)
+            if in_bulk:
+                continue
+            partial_line = partial_line + part
+            if b"\n" not in part:
+                # Make sure we don't go O(n^2) from constantly appending to our partial_line.
+                if len(partial_line) > 100000:
```
MarkLindblad, line 238: Maybe make the 10,000 a global variable?

PR #592, opened 2024-07-26
Use Helicone for CI runs.

mdwelsh:


HenryL27 (approved):
lg. I assume this won't divert the azure oai tests (or will do so correctly)?

mdwelsh:
> lg. I assume this won't divert the azure oai tests (or will do so correctly)?

It should not impact Azure at all.

PR #595, opened 2024-07-27
Add codegen versions of basic query integration tests

baitsguy:
Same caveats of existing tests, but this verifies the basic cases can generate runnable code.


tranade (commented):
seems to be an import error in `test_planner.py` integ test (from https://github.com/aryn-ai/sycamore/pull/594)?

tranade (approved):

PR #599, opened 2024-07-29
Support user tracking for Helicone

tranade:
Send `user` parameter in OpenAI call based on environment variable `SYCAMORE_HELICONE_USER`


mdwelsh:
Thanks! I didn't notice this was unrelated to Helicone - you can call it SYCAMORE_OPENAI_USER instead. Thanks!

mdwelsh (approved):
Maybe name it SYCAMORE_OPENAI_USER instead

PR #601, opened 2024-07-29
Added Entity Extractor + HierarchicalDocument

RitxmSaha:
This PR adds the EntityExtractor to the graph extractor class and also adds HierarchicalDocument as a new experimental data model.

GraphExtractor Class:
- Implemented EntityExtractor
- Standardized resolve to work with all types of GraphExtractor

HierarchicalDocument:
- Implemented as inherited class of document, no longer uses elements property, uses children which is list[HierarchicalDocument]
- Implemented recursive version of explode if isinstance(doc,HierarchicalDocument)


lib/sycamore/sycamore/transforms/explode.py
```diff
@@ -28,17 +29,21 @@ def __init__(self, child: Node, **resource_args):
...
+        documents: list[Document] = [parent]
         import uuid
 
         for i, element in enumerate(parent.elements):
             cur = Document(element.data)
-            if "doc_id" in cur.data:
-                cur.doc_id = cur.data["doc_id"]
-            else:
-                cur.doc_id = str(uuid.uuid4())
+            cur.doc_id = str(uuid.uuid4())
```
baitsguy, line 46: why did we change this?

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -30,6 +34,20 @@ def __init__(self, nodeKey: str, nodeLabel: str, relLabel: str):
         self.relLabel = relLabel
 
 
+class GraphEntity(GraphData):
```
baitsguy, line 37: Is GraphNode a more generally used convention or entity?

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -46,6 +68,16 @@ def resolve(self, docset: "DocSet") -> "DocSet":
         from sycamore.reader import DocSetReader
         from ray.data.aggregate import AggregateFn
 
+        reader = DocSetReader(docset.context)
+
+        # Get list[Document] representation of docset, trigger execute with take_all()
+        execution = Execution(docset.context, docset.plan)
+        dataset = execution.execute(docset.plan)
+        docs = dataset.take_all(None)
```
baitsguy, line 76: you can remove the None, a little confusing with it there

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -154,10 +185,121 @@ def extract_metadata(self, doc: Document) -> Document:
...
+    def extract(self, docset: "DocSet") -> "DocSet":
+        """
+        Extracts entities from documents then creates a document in the docset where they are stored as nodes
+        """
+        docset.plan = ExtractFeatures(docset.plan, self)
+        docset = self.resolve(docset)
+        return docset
+
+    def _extract(self, doc: HierarchicalDocument) -> HierarchicalDocument:
+        from multiprocessing import Pool
```
baitsguy, line 210: I'm a little worried about doing this in here. It's unclear that an EntityExtractor can internal spawn threads with its own configurations. Maybe calling it MultithreadedEntityExtractor or something will help?

@alexaryn @eric-anderson I'm curious about your thoughts here since you've played with this stuff in other parts

baitsguy (commented):
Looks good overall, a few clarifications and wanted some feedback on the threading

lib/sycamore/sycamore/transforms/explode.py
RitxmSaha: This was the previous implementation that I made. It was to make sure each entity in the entities property did not have their doc_id rewritten(since I had already built relationships between those id's and parent documents). We don't need this anymore since we switched over to using hierarchicaldocument class where the id's are assigned to children when they are converted to hierarchicaldocuments.

lib/sycamore/sycamore/data/document.py
```diff
@@ -301,6 +303,66 @@ def split_data_metadata(all: list[Document]) -> tuple[list[Document], list[Metad
...
+
+    @property
+    def elements(self) -> list[Element]:
+        raise ValueError("MetadataDocument does not have elements")
+
+    @elements.setter
+    def elements(self, elements: list[Element]):
+        raise ValueError("MetadataDocument does not have elements")
+
+    def __str__(self) -> str:
```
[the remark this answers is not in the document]
RitxmSaha: I tried putting it in the base class, but the base class does not have the children property that I use in HierarchicalDocument. My hope was that as we transition over to a hierarchical data model, this would not be a seperate class anymore so the bloat wouldn't be as big.

lib/sycamore/sycamore/tests/unit/transforms/test_graph_extractor.py
```diff
@@ -1,48 +1,112 @@
...
-from sycamore.data import Document
+from sycamore.transforms.extract_graph import GraphMetadata, MetadataExtractor, GraphEntity, EntityExtractor
+from sycamore.data import HierarchicalDocument
 from collections import defaultdict
 
 
 class TestGraphExtractor:
-    docs = [
-        Document(
+    metadata_docs = [
```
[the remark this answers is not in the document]
RitxmSaha: Not exactly sure what this means. I put metadata_docs for the documents used for test_metadata_extractor and entity_docs for test_entity_extractor.

lib/sycamore/sycamore/transforms/extract_graph.py
RitxmSaha: Not sure if that would be better since it is technically used to extract a specific type of node, not a singular node.

EntityExtractor uses the label and description of GraphEntity s to extract all nodes that fit that specific label + description.

lib/sycamore/sycamore/transforms/extract_graph.py
RitxmSaha: Do you think adding a multithreading flag to entity extractor would be a good solution here?

lib/sycamore/sycamore/transforms/extract_graph.py
RitxmSaha: fixed

lib/sycamore/sycamore/data/document.py
```diff
@@ -301,6 +303,66 @@ def split_data_metadata(all: list[Document]) -> tuple[list[Document], list[Metad
...
+        """TODO"""
+        self.data["children"] = []
+
+    @property
+    def elements(self) -> list[Element]:
+        raise ValueError("MetadataDocument does not have elements")
+
+    @elements.setter
+    def elements(self, elements: list[Element]):
+        raise ValueError("MetadataDocument does not have elements")
```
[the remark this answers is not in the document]
RitxmSaha: fixed

lib/sycamore/sycamore/data/document.py
```diff
@@ -301,6 +303,66 @@ def split_data_metadata(all: list[Document]) -> tuple[list[Document], list[Metad
...
+        self.data["children"] = children
+
+    @children.deleter
+    def children(self) -> None:
+        """TODO"""
+        self.data["children"] = []
+
+    @property
+    def elements(self) -> list[Element]:
+        raise ValueError("MetadataDocument does not have elements")
```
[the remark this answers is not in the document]
RitxmSaha: fixed

lib/sycamore/sycamore/data/document.py
```diff
@@ -301,6 +303,66 @@ def split_data_metadata(all: list[Document]) -> tuple[list[Document], list[Metad
...
+            self.children = []
+        if "elements" in self.data:
+            for element in self.data["elements"]:
+                self.children.append(HierarchicalDocument(Document(element.data)))
+
+        del self.data["elements"]
+
+    @property
+    def children(self) -> list["HierarchicalDocument"]:
+        """TODO"""
```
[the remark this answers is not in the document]
RitxmSaha: fixed

lib/sycamore/sycamore/data/document.py
```diff
@@ -301,6 +303,66 @@ def split_data_metadata(all: list[Document]) -> tuple[list[Document], list[Metad
     )
 
 
+############### EXPERIMENTAL
+class HierarchicalDocument(Document):
+    def __init__(self, document=None, **kwargs):
+        super().__init__(document)
+        if "doc_id" not in self.data:
+            self.doc_id = str(uuid.uuid4())
```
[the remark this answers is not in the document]
RitxmSaha: fixed(really cool solution now i see the power of .get)

lib/sycamore/sycamore/tests/unit/transforms/test_graph_extractor.py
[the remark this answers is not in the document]
baitsguy: Just a naming thing, we have a MetadataDocument, so this was confusing

baitsguy (approved):
Nice work!

PR #603, opened 2024-07-29
Update get_hash_context_file func

pparmar30:


alexaryn (approved):
Seems fine to me.  It's unfortunate that mypy makes duck typing so onerous.

PR #605, opened 2024-07-29
Change docs to use `ArynPartitioner` instead of `SycamorePartitioner`

MarkLindblad:


docs/source/conf.py
```diff
@@ -30,6 +30,7 @@
 
 templates_path = ["_templates"]
 exclude_patterns = []
+smartquotes = False
```
MarkLindblad, line 33: Fixes `--pull=always` looking like `--pull=always` [here](https://sycamore.readthedocs.io/en/stable/welcome_to_sycamore/get_started.html#:~:text=docker%20compose%20up-,%E2%80%93pull%3Dalways,-.%20See%20the%20.env). Downside: now all API docs use -- too.

pparmar30: Do things break if someone specifies "--" instead of '-' in the API docs?

MarkLindblad: No, it is aesthetic: for instance, it changes "...augmentor (TextAugmentor) – A TextAugmentor..." to "...augmentor (TextAugmentor) -- A TextAugmentor..." on the [Docset page](https://sycamore.readthedocs.io/en/latest/APIs/data_preparation/docset.html#:~:text=augmentor%20(TextAugmentor)%20%E2%80%93%20A%20TextAugmentor).

pparmar30 (approved):

PR #609, opened 2024-07-29
reading env variable as suggested and cosmetic changes

Soeb-aryn:
Making changes to avoid users passing api key as string from notebook.


notebooks/ArynPartitionerPython.ipynb
```diff
@@ -71,7 +71,9 @@
    "id": "4",
    "metadata": {},
    "source": [
-    "Replace the `YOUR-ARYN-API-KEY` with your key "
+    "Replace the `YOUR-ARYN-API-KEY` with your key and execute the export command\n",
```
eric-anderson, line 74: This is still flawed. It requires people to put their key in the notebook content itself. 
Think of the key as a password. You wouldn't put your password in here and commit it to github, so you shouldn't encourage people to put the password in the same place.

The safe way to get your key into the environment is something like:
```
% ARYN_API_KEY=... jupyter lab
```

notebooks/ArynPartitionerPython.ipynb
```diff
@@ -81,7 +83,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "aryn_api_key = 'YOUR-ARYN-API-KEY'"
+    "from sycamore.utils.aryn_config import ArynConfig\n",
+    "aryn_api_key = ArynConfig.get_aryn_api_key()\n",
+    "\n",
+    "# you can also read directly from env variable like this\n",
```
eric-anderson, line 89: We shouldn't need this; our APIs  automatically read the api keys. 
The main reason you need it is if you stored the key somewhere else, in which case `ArynConfig.get_aryn_api_key("/home/eric/keys/aryn/config.yaml")` would work.

notebooks/ArynPartitionerWithLangchain.ipynb
```diff
@@ -112,7 +119,7 @@
     "\n",
     "\n",
     "partitioned_docset = pdf_docset.partition(\n",
-    "    partitioner=ArynPartitioner(threshold=0.35, extract_table_structure=True, aryn_api_key = aryn_api_key, use_partitioning_service = True ) \n",
+    "    partitioner=ArynPartitioner(threshold=0.35, extract_table_structure=True, aryn_api_key = aryn_api_key) \n",
```
eric-anderson, line 122: If users do the env var or put the file as ~/.aryn/config.yaml, the aryn_api_key = aryn_api_key should be unneeded.

eric-anderson (commented):
I'd suggest something like the following cell:
```
from sycamore.utils.aryn_config import ArynConfig, _DEFAULT_PATH
assert ArynConfig.get_aryn_api_key() != "", f"Unable to find aryn API key.  Looked in {_DEFAULT_PATH}"
# if the above assertion fails, you can either set the environment variable ARYN_API_KEY and restart jupyter
# or make a yaml file at the specified path in the assertion error that looks like the next line with the leading '# ':
# aryn_token: "YOUR-ARYN-API-KEY"

# It is unsafe, but if neither of those options work, you can put it in this notebook with
# import os
# os.environ["ARYN_API_KEY"] = "UNSAFE-ARYN-API-KEY-LOCATION"
# but beware that it is easy to accidentally commit the notebook file and have it include your key.
```

notebooks/ArynPartitionerPython.ipynb
Soeb-aryn: yes, this was not clear. I will remove this entirely and add your suggestion which is much more clear and concise.

```from sycamore.utils.aryn_config import ArynConfig, _DEFAULT_PATH
assert ArynConfig.get_aryn_api_key() != "", f"Unable to find aryn API key.  Looked in {_DEFAULT_PATH}"
# if the above assertion fails, you can either set the environment variable ARYN_API_KEY and restart jupyter
# or make a yaml file at the specified path in the assertion error that looks like the next line with the leading '#':
# aryn_token: "YOUR-ARYN-API-KEY"

# It is unsafe, but if neither of those options work, you can put it in this notebook with
# import os
# os.environ["ARYN_API_KEY"] = "UNSAFE-ARYN-API-KEY-LOCATION"
# but beware that it is easy to accidentally commit the notebook file and have it include your key.```

File: notebooks/ArynPartitionerPython.ipynb (hunk 2 above)
Comment (Soeb-aryn) [reply notebooks/ArynPartitionerPython.ipynb:89]: thanks, will update the notebooks.

File: notebooks/ArynPartitionerWithLangchain.ipynb (hunk 3 above)
Comment (Soeb-aryn) [reply notebooks/ArynPartitionerWithLangchain.ipynb:122]: thanks Eric, I will update the notebook.

File: notebooks/metadata-extraction.ipynb [hunk 4]
```diff
@@ -180,7 +193,7 @@
    "source": [
     "for d in sample_docs:\n",
     "    print(f\"Path: {d.properties['path']}\")\n",
-    "    for k in [\"dateAndTime\", \"day\", \"aircraft\", \"location\"]:\n",
+    "    for k in [\"dateTime\", \"day\", \"aircraft\", \"location\"]:\n",
```
eric-anderson, line 196: Are these stable?  I remember Vinayak mentioning at one point that they aren't.  dateAndTime is mentioned above in convert_timestamp, but if one is missing and the other is present you could add code to normalize to dateTime.

notebooks/ArynPartitionerPython.ipynb
```diff
@@ -67,21 +67,35 @@
...
-   "cell_type": "markdown",
-   "id": "4",
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "5",
    "metadata": {},
+   "outputs": [],
    "source": [
-    "Replace the `YOUR-ARYN-API-KEY` with your key "
+    "from aryn_sdk.utils.aryn_config import ArynConfig, DEFAULT_PATH\n",
```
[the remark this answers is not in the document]
Soeb-aryn: Added changes in library to accommodate, the library changes are also added in this PR now.

notebooks/duckdb-writer.ipynb
```diff
@@ -81,7 +102,7 @@
     "ds = (\n",
     "    ctx.read.binary(paths, binary_format=\"pdf\", filesystem=fsys)\n",
     "    # Partition with the Aryn partitioner remotely, pulling out tables and images.\n",
-    "    .partition(partitioner=ArynPartitioner(aryn_api_key=aryn_api_key, use_partitioning_service=True, use_cache=False, extract_images=True,  extract_table_structure=True))\n",
+    "    .partition(partitioner=ArynPartitioner( extract_images=True,  extract_table_structure=True))\n",
```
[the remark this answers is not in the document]
Soeb-aryn: resolved

notebooks/metadata-extraction.ipynb
Soeb-aryn: This is working, It was stable for multiple test runs I did.

lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -5,7 +5,9 @@
 
 from typing import Dict, Any
 
-_DEFAULT_PATH = os.path.join(pathlib.Path.home(), ".aryn", "config.yaml")
+DEFAULT_PATH = os.path.join(pathlib.Path.home(), ".aryn", "config.yaml")
+_DEFAULT_PATH = DEFAULT_PATH
+## ToDO: remove after 31/08/2024
```
eric-anderson, line 10: nit:
I'll recommend writing dates as YYYY-MM-DD. Three advantages:
1. its the ISO 8601 standard
2. they sort properly in that order
3. It's sufficiently unambiguous even with days <=12.

notebooks/ArynPartitionerExample.ipynb
```diff
@@ -27,18 +30,29 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "aryn_api_key = 'YOUR-ARYN-API-KEY'"
+    "from sycamore.utils.aryn_config import ArynConfig, DEFAULT_PATH\n",
```
eric-anderson, line 33: If someone gets this version of the notebook, and the older (in pypi) version of the library it won't work.
The safer way to do this is to import _DEFAULT_PATH here, and then once we release the library with
DEFAULT_PATH switched to change the notebooks.

eric-anderson (approved):

lib/sycamore/sycamore/utils/aryn_config.py
Soeb-aryn: Makes sense, will use this format in future.

notebooks/ArynPartitionerExample.ipynb
Soeb-aryn: Done.

PR #610, opened 2024-07-29
Refactor LogicalOperators to use pydantic.

mdwelsh:
This PR leverages Pydantic and type introspection to allow LogicalOperators to be specified directly as pydantic BaseModel subclasses.

The LLM-provided description for each operator is derived from the class docstring, rather than having to write a separate method to return a string.

The input schema for each operator is directly determined from the class's typing hints.


lib/sycamore/sycamore/query/operators/logical_operator.py
```diff
@@ -1,36 +1,33 @@
...
-
-    @staticmethod
-    def output_schema() -> Dict[str, Any]:
-        raise NotImplementedError
+    description: Optional[str] = None
+    """A detailed description of why this operator was chosen for this query plan."""
+
+    input: Optional[List[int]] = None
+    """A list of node IDs that this operation depends on."""
+
```
tranade, line 24: does this need node_id as well?

lib/sycamore/sycamore/llms/llms.py
```diff
@@ -10,11 +10,11 @@ def __init__(self, model_name, cache: Optional[Cache] = None):
         self._cache = cache
 
     @abstractmethod
-    def generate(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> Any:
+    def generate(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> str:
```
baitsguy, line 13: I dunno how its implemented right now but some invocations expect a dict response, e.g. extract_schema. It might just be returning a json though

lib/sycamore/sycamore/query/execution/physical_operator.py
```diff
@@ -57,12 +59,12 @@ class MathOperator(PhysicalOperator):
...
     """
 
-    def __init__(self, logical_node: LogicalOperator, query_id: str, inputs: List[Any]) -> None:
+    def __init__(self, logical_node: Math, query_id: str, inputs: List[Any]) -> None:
         super().__init__(logical_node, query_id, inputs)
 
     def execute(self) -> Any:
-        assert self.logical_node.data is not None
-        operator = self.logical_node.data.get("type")
+        assert isinstance(self.logical_node, Math)
```
baitsguy, line 66: asserted by the construction already?

lib/sycamore/sycamore/query/execution/physical_operator.py
```diff
@@ -79,10 +81,9 @@ def execute(self) -> Any:
         return result
 
     def script(self, input_var: Optional[str] = None, output_var: Optional[str] = None) -> Tuple[str, List[str]]:
+        assert isinstance(self.logical_node, Math)
```
baitsguy, line 84: same

lib/sycamore/sycamore/query/operators/logical_operator.py
mdwelsh: No, it gets it from the Node superclass.

lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -91,10 +91,11 @@ def process_node(self, logical_node: LogicalOperator, query_id: str) -> Any:
         # Process dependencies
         if logical_node.dependencies:
             for dependency in logical_node.dependencies:
+                assert isinstance(dependency, LogicalOperator)
```
baitsguy, line 94: want to just assert this in LogicalOperator instead

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -110,20 +111,18 @@ def __init__(
         self.os_client_args = os_client_args
 
     def execute(self) -> Any:
-        assert self.logical_node.data and "index" in self.logical_node.data
-        result = self.context.read.opensearch(
-            os_client_args=self.os_client_args, index_name=self.logical_node.data["index"]
-        )
+        assert isinstance(self.logical_node, LoadData)
```
baitsguy, line 114: can this ever be false

baitsguy: same q for the other operators too

lib/sycamore/sycamore/query/operators/count.py
```diff
@@ -1,39 +1,16 @@
...
-            and justification of why you chose to use it.
-        - *input* is a list of operation ids that this operation depends on. For this operation,
-            *input* should only contain one id of an operation that returns a database (len(input) == 1).
-        - *field* is an optional parameter. non-primary database field to return a count based on.
-        - *primaryField* is an optional parameter. It is a primary field that represents what a
-              unique entry is considered for the data provided.
-        - *id* is a uniquely assigned integer that serves as an identifier.
-        """
+    field: Optional[str] = None
+    """Non-primary database field to return a count based on."""
```
baitsguy, line 13: would the generated prompt also say this is optional?

lib/sycamore/sycamore/llms/llms.py
baitsguy: Nvm I see you found summarize was relying on the dict return

baitsguy (commented):
I haven't read every docstring in query.operators, we need to test that stuff through the upcoming benchmark anyway

Couple nitty questions but looks good overall

lib/sycamore/sycamore/query/execution/sycamore_operator.py
mdwelsh: Yes, because someone might have come along and manually done `node.logical_node = some_other_thing` after it was constructed. mypy isn't happy unless we make this assertion. I am not sure it would help to make the field immutable since I don't think (not sure) that mypy understands that.

lib/sycamore/sycamore/query/execution/physical_operator.py
mdwelsh: Sadly this isn't good enough in case the field is reassigned.

lib/sycamore/sycamore/query/operators/count.py
mdwelsh: Yes, because the type hint that is included as part of the prompt says `typing.Optonal[str]`. At least GPT-4o seems to understand Python syntax when describing the types, but we may need to tweak how this is rendered to the LLM prompt if that does not work.

lib/sycamore/sycamore/query/execution/sycamore_executor.py
mdwelsh: You mean by overriding the `dependencies` property? Currently this is provided by `Node`, not `LogicalOperator`, so the only place we can assert is here.

baitsguy: I thought you can do

```
class LogicalOperator..:

def __init__:
  super()
  # assert dependencies are LogicalOperators
```

anyway not a blocker

mdwelsh: The problem is that one can reassign the values of any of the dependencies after the constructor is called, since the object as a whole is not immutable.

tranade (approved):
Codegen integ tests seem to be broken, but I'm modifying this in my PR anyways so will fix it. Looks good otherwise!

lib/sycamore/sycamore/query/execution/sycamore_executor.py
baitsguy: Ahh gotcha

baitsguy (approved):
Lint break in query-ui but ship it after

PR #619, opened 2024-07-30
Merge elements in ntsb test ingest

baitsguy:
Reduces number of elements from ~250 to 3, this results in a big reduction in LLM calls since that's what these tests currently do. The 1000 tokens number is made up, it probably results in a loss of answer quality but that's not what these tests are for.


tranade (approved):
1000 tokens seems fine to me

PR #620, opened 2024-07-30
Small update to `field_to_value`

tranade:
- Make `field_to_value` return None as a default, if the field does not exist
- Related updates in places this function is used


baitsguy (approved):

baitsguy:
Looks like an integ test failures

PR #621, opened 2024-07-30
parse html tables better

HenryL27:
when you get a table with extra stuff in the tag, e.g. from pubtabnet
```
<html> <head> <meta charset="utf-8"/> <style> table, th, td { border: 1px solid black; font-size: 10px; } </style> </head> <body> <table frame="hsides" rules="groups" width="100%"> <thead> <tr> <td> <b> Kinetic parameter </b> </td> <td> <b> ND </b> </td> <td>...
```
and you try to parse the table string part don't vomit because the string doesn't start with `<table>`


bsowell (approved):
Good catch.

PR #623, opened 2024-07-31
Add LLMFilter as a DocSet Transform

tranade:
- Move LLMFilter as DocSet transform
- Add LLMFilter prompt into `default_prompts.py`
- Update some function inputs
- Update tests


lib/sycamore/sycamore/docset.py
```diff
@@ -771,6 +772,50 @@ def process_doc(doc: Document) -> Document:
 
         return self.map(process_doc, **resource_args)
 
+    def llm_filter(
+        self,
+        client: OpenAI,
```
baitsguy, line 777: we should probably accept an Llm here

lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -169,6 +169,26 @@ def __init__(self, question: Optional[str], field: Optional[str], format: Option
             )
 
 
+class LLMFilterMessagesPrompt(OpenAIMessagesPromptBase):
+    def __init__(self, filter_question: Optional[str]):
+        super().__init__()
+
+        self.add_message(
+            "system",
+            ("You are a helpful classifier that generously filters database " "entries based on questions."),
```
baitsguy, line 178: are the double quotes after database doing anything?

lib/sycamore/sycamore/docset.py
```diff
@@ -771,6 +772,50 @@ def process_doc(doc: Document) -> Document:
...
+        new_field: str,
+        messages: List[dict],
```
baitsguy, line 779: This feels a little strange because you're expecting the caller to know how to format generic messages. What do you think about just accepting a prompt here? And that can be messages or single string

tranade: Hmm, so list[dict] or str? And based on which one it is do {"messages": messages} or {"prompt": prompt}?

lib/sycamore/sycamore/llms/prompts/default_prompts.py
tranade: oops nope

lib/sycamore/sycamore/docset.py
baitsguy: can just accept a dict like prompt_kwargs (the way extract_entity does), and that can either have messages or prompt

tranade: I think extract_entity does it in a different way, uses a prompt template

lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -125,9 +127,16 @@ def _handle_document_field_prompting(self, document: Document) -> Any:
...
-        self._messages.append({"role": "user", "content": value})
 
-        response = self._llm.generate(prompt_kwargs={"messages": self._messages}, llm_kwargs={})
+        if isinstance(self._prompt, str):
+            self._prompt += value
+            response = self._llm.generate(prompt_kwargs={"prompt": self._prompt}, llm_kwargs={})
+        else:
+            if self._prompt is None:
+                self._prompt = []
+            self._prompt.append({"role": "user", "content": value})
```
baitsguy, line 137: will need to revisit this for non openai but this is fine for now

baitsguy (approved):

PR #624, opened 2024-07-31
Jonfritz patch docsupdate

jonfritz:
Added some info to the Aryn Partitioning Service section.

### DIFF
--- a/docs/source/aryn_cloud/aryn_partitioning_service.md
+++ b/docs/source/aryn_cloud/aryn_partitioning_service.md
@@ -2,7 +2,7 @@
 
 You can use the Aryn Partitioning Service to segment PDF's into labeled bounding boxes identifying titles, tables, table rows and columns, images, and regular text. Bounding boxes are returned as JSON with their associated text for easy use. It leverages [Aryn's purpose-built AI model for document segmentation and labelling](https://huggingface.co/Aryn/deformable-detr) that was trained using DocLayNet – an open source, human-annotated document layout segmentation dataset containing tens of thousands of pages from a broad variety of document sources.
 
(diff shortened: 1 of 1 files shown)


docs/source/aryn_cloud/aryn_partitioning_service.md
```diff
@@ -102,6 +107,8 @@ The available options are listed below:
 * ```use_ocr```: If ```true```, the partitioner uses OCR to extract text from the PDF. It defaults to ```false```, where the partitioner attempts to directly extract the text from the underlying PDF in the bounding box. It currently uses Tesseract for extraction.
 * `extract_table_structure`: If `true`, the partitioner runs a separate table extraction model to extract cells from regions of the document identified as tables.
 * `extract_images`: If `true`, the partitioner crops each region identified as an image and attaches it to the associated `ImageElement`. This can later be fed into the `SummarizeImages` transform when used within Sycamore.
+* `selected_pages`: You can specify a page (like [11]) or a page range (like [25,30]) of your PDF to process.
```
HenryL27, line 110: I'd add here that the first page is 1, not 0

docs/source/aryn_cloud/aryn_partitioning_service.md
```diff
@@ -102,6 +107,8 @@ The available options are listed below:
...
+* `pages_per_call`: This is only available when using the Partition function in Sycamore. This option divides the processing of your document into batches of pages, and you specify the size of each batch. This is useful when running OCR on large documents.
```
HenryL27, line 111: If this is only applicable in sycamore, maybe put it in the sycamore section?

HenryL27 (approved):
Few nits, but won't block

docs/source/aryn_cloud/aryn_partitioning_service.md
jonfritz: Added.

docs/source/aryn_cloud/aryn_partitioning_service.md
jonfritz: People might not know where to look, and just jump to Options. I think it's most discoverable here, but will also add under the Sycamore section.

PR #625, opened 2024-07-31
Create `count_distinct` for DocSet

tranade:
- DocSet count now takes an optional `field` param, which is used for count unique


lib/sycamore/sycamore/docset.py
```diff
@@ -115,13 +115,19 @@ def _truncate(s):
 
             pprint.pp(document, stream=stream)
 
-    def count(self, include_metadata=False, **kwargs) -> int:
+    def count(self, field: Optional[str] = None, include_metadata=False, **kwargs) -> int:
```
baitsguy, line 118: Can you move the field arg after include_metadata?

that way if someone is currently doing `ds.count(True)` they won't break

baitsguy (commented):
Code looks good, will discuss offline if adding a `.count_distinct(key)` is a better approach

lib/sycamore/sycamore/docset.py
tranade: working on making a separate `count_distinct` instead

tranade (commented):
Previous commit also makes a separate count distinct

baitsguy (approved):

PR #627, opened 2024-07-31
Add Context.config

baitsguy:
Let's you set some global variables accessible in the rest of the script. The LLM and OS changes in this PR are example usages

Example code (see signature change in `write.opensearch()` and `OpenAIEntityExtractor` initialization)

```python
context = sycamore.init(
  config=Config(
      {
          Config.OPENSEARCH_CLIENT_CONFIG: os_client_args,
          Config.OPENSEARCH_INDEX_NAME: "toyindex",
          Config.OPENSEARCH_INDEX_SETTINGS: index_settings,
          Config.LLM: openai_llm,
      }
  )
)
ds = (
    context.read.binary(paths, binary_format="pdf")
    .partition(partitioner=UnstructuredPdfPartitioner())
    .extract_entity(entity_extractor=OpenAIEntityExtractor("title", prompt_template=title_context_template))
    .extract_entity(entity_extractor=OpenAIEntityExtractor("authors", prompt_template=author_context_template))
    .merge(GreedyTextElementMerger(tokenizer=tokenizer, max_tokens=300))
    .explode()
    .sketch()
    .embed(
        embedder=SentenceTransformerEmbedder(
            batch_size=100, model_name="sentence-transformers/all-MiniLM-L6-v2"
        )
    )
)
ds.write.opensearch()
```


lib/sycamore/sycamore/config.py
```diff
@@ -0,0 +1,21 @@
+from typing import Dict, Optional, Any
+
+
+class Config:
+
+    OPENSEARCH_CLIENT_CONFIG = "opensearch.client_config"
+    OPENSEARCH_INDEX_NAME = "opensearch.index_name"
+    OPENSEARCH_INDEX_SETTINGS = "opensearch.index_settings"
```
HenryL27, line 8: maybe these constants belong in a more opensearchy namespace?

baitsguy: can reorganize as they grow, for now figured having configs in 1 place is easier to manage

lib/sycamore/sycamore/config.py
```diff
@@ -0,0 +1,21 @@
+from typing import Optional, Any
+
+
+class Config:
```
mdwelsh, line 4: Why not just use a regular `@dataclass` for this, or even a Pydantic `BaseModel` subclass (which gives you serialization and validation for free)? I don't know why one would implement this as a key-value pair mapping :-)

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
```diff
@@ -97,7 +98,7 @@ def test_pdf_to_opensearch_with_llm_caching():
 
         """
 
-    s3_cache_base_path = os.environ["SYCAMORE_S3_TEMP_PATH"]
+    s3_cache_base_path = os.environ.get("SYCAMORE_S3_TEMP_PATH", "s3://aryn-sycamore-integ-temp/")
```
mdwelsh, line 101: It definitely feels like the S3 cache path should be part of the config object

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
```diff
@@ -133,12 +138,7 @@ def test_pdf_to_opensearch_with_llm_caching():
                 )
             )
         )
-
-        ds.write.opensearch(
-            os_client_args=os_client_args,
-            index_name="toyindex",
-            index_settings=index_settings,
-        )
+        ds.write.opensearch()
```
mdwelsh, line 141: I tend to think that one is more likely to read from and write to different indices, so maybe the index name is still a parameter to the read.opensearch() and write.opensearch() calls, but the OS config is in the config object.

lib/sycamore/sycamore/writer.py
```diff
@@ -97,6 +98,14 @@ def opensearch(
         from typing import Any
         import copy
 
+        if os_client_args is None:
+            os_client_args = self.context.config.get(Config.OPENSEARCH_CLIENT_CONFIG)
+        assert os_client_args is not None
+
+        if not index_name:
```
mdwelsh, line 105: See comment above, I tend to think this should be passed in on each call

lib/sycamore/sycamore/config.py
baitsguy: Haha fair enough, let me update

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
baitsguy: Yeah will migrate these in follow up PRs feature wise

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
baitsguy: They're all still parameters, they just default to reading from the config if not specified. You can still do ds.write.opensearch(index="something")

lib/sycamore/sycamore/writer.py
baitsguy: replied above

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
baitsguy: we have a couple notebooks I believe that are using the same index in a bunch of places

baitsguy: write, then query etc

lib/sycamore/sycamore/config.py
baitsguy: Discussed offline, will start with a typed BaseModel and can expand it to free-formed fields if needed later

lib/sycamore/sycamore/context.py
```diff
@@ -109,7 +117,11 @@ def deregister_rule(self, rule: Rule) -> None:
 _global_context: Optional[Context] = None
 
 
-def init(exec_mode=ExecMode.RAY, ray_args: Optional[dict[str, Any]] = None) -> Context:
+def init(exec_mode=ExecMode.RAY, ray_args: Optional[dict[str, Any]] = None, config: Optional[Config] = None) -> Context:
+    """
+    Initialized a new Context. If there is already an initialized Context, we reuse the Ray session but allow you to
+    override any Config variables.
+    """
```
HenryL27, line 124: ooh actual documentation for sycamore.init()! (nit: Initialized -> Initialize; and maybe add an Args: block)

lib/sycamore/sycamore/context.py
```diff
@@ -122,9 +134,19 @@ def init(exec_mode=ExecMode.RAY, ray_args: Optional[dict[str, Any]] = None) -> C
 
             sycamore_logger.setup_logger()
 
-            _global_context = Context(exec_mode, ray_args)
+            _global_context = Context(exec_mode, ray_args, config)
+        if config:
+            _global_context._config = config
```
HenryL27, line 139: Is this gonna do funky things to my ray workers?

lib/sycamore/sycamore/config.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@dataclass
+class Config:
+
+    opensearch_client_config: Optional[dict[str, Any]] = None
+    opensearch_index_name: Optional[str] = None
+    opensearch_index_settings: Optional[dict[str, Any]] = None
+
+    llm: Optional[LLM] = None
```
HenryL27, line 14: So if I'm a user and I want to set my own config vars how do I do that? Can I just do something like:
```python
cfg = Config()
cfg.myfieldwithlikegcscredentialsorsomething = "efngsogijaeropgijw"
```
or do I have to pull subclassing shenanigans

lib/sycamore/sycamore/context.py
HenryL27: Maybe add this module to somewhere in the api docs toctree, idk

lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -1,14 +1,14 @@
 from abc import ABC, abstractmethod
 from typing import Callable, Any, Optional, Union
 
-
+from sycamore import context
```
HenryL27, line 4: this feels weird to me, not sure why. Think I'd be more comfortable with either using it as `sycamore.context.current()` or exporting current to sycamore and having `sycamore.current()`. Eh, I guess this is fine. Disregard the above unless you want to.

lib/sycamore/sycamore/writer.py
```diff
@@ -97,6 +97,14 @@ def opensearch(
         from typing import Any
         import copy
 
+        if os_client_args is None:
+            os_client_args = self.context.config.opensearch_client_config
+        assert os_client_args is not None
```
HenryL27, line 102: Add an error message explaining what went wrong here

HenryL27 (commented):
non-blocking so if you want an approve ok but idk if there are other reviewers who disagree

lib/sycamore/sycamore/transforms/extract_entity.py
baitsguy: I think the former is fine, little more readable too - will do

lib/sycamore/sycamore/writer.py
baitsguy: Ack

HenryL27 (approved):

PR #630, opened 2024-08-01
Include match and range filter functions

tranade:
- Allow for filtering with docset.filter(f=MatchFilter(...)) and docset.filter(f=RangeFilter(...))


tranade (commented):
Wondering if docset.filter(f=match_filter(...)) and docset.filter(f=range_filter(...)) is a better approach. If so, should I include them in sycamore/transforms/basics.py or as a new file under sycamore/functions?

Edit: Resolved offline

baitsguy (approved):

PR #632, opened 2024-08-01
Added OpenAI Async client

RitxmSaha:
Added ```generate_future()``` to the openai client that can be used to request a large amount of openai calls in parallel.

Example usage:
```python
async def gather_api_calls():
    tasks = [self.llm.generate_future(prompt_kwargs={"prompt": prompt}, llm_kwargs={...}) for prompt in prompts]
    res = await asyncio.gather(*tasks)
    return res

res = asyncio.run(gather_api_calls())
```

### DIFF
--- a/lib/sycamore/sycamore/llms/llms.py
+++ b/lib/sycamore/sycamore/llms/llms.py
@@ -17,7 +17,7 @@ class LLM(ABC):
     def is_chat_mode(self) -> bool:
         pass
 
-    async def generate_future(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> Awaitable[str]:
+    async def generate_async(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> Awaitable[str]:
         raise ValueError("No implementation for llm futures exists")
 
 
--- a/lib/sycamore/sycamore/llms/llms.py
+++ b/lib/sycamore/sycamore/llms/llms.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Awaitable, Optional
 
 from sycamore.utils.cache import Cache
 
@@ -16,6 +16,9 @@ class LLM(ABC):
     @abstractmethod
     def is_chat_mode(self) -> bool:
         pass
+
+    async def generate_future(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> Awaitable[str]:
+        raise ValueError("No implementation for llm futures exists")
 
 
 class FakeLLM(LLM):
--- a/lib/sycamore/sycamore/llms/openai.py
+++ b/lib/sycamore/sycamore/llms/openai.py
@@ -125,7 +125,7 @@ class OpenAIClientWrapper:
 
             self.api_key = os.environ.get("AZURE_OPENAI_API_KEY")
 
-    def get_client(self, asynchronous: bool = False) -> OpenAIClient:
+    def get_client(self, asynchronous: bool = False) -> Union[OpenAIClient, AsyncOpenAIClient]:
         if self.client_type == OpenAIClientType.OPENAI:
             # We currently only support Helicone with OpenAI.
             base_url = self.base_url
--- a/lib/sycamore/sycamore/llms/openai.py
+++ b/lib/sycamore/sycamore/llms/openai.py
@@ -125,7 +125,7 @@ class OpenAIClientWrapper:
 
             self.api_key = os.environ.get("AZURE_OPENAI_API_KEY")
 
-    def get_client(self, asynchronous: bool = False) -> Union[OpenAIClient, AsyncOpenAIClient]:
+    def get_client(self) -> OpenAIClient:
         if self.client_type == OpenAIClientType.OPENAI:
             # We currently only support Helicone with OpenAI.
             base_url = self.base_url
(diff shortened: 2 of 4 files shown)


lib/sycamore/sycamore/llms/llms.py
```diff
@@ -17,6 +17,9 @@ def generate(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) ->
     def is_chat_mode(self) -> bool:
         pass
 
+    async def generate_future(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> Awaitable[str]:
```
eric-anderson, line 20: We should call this generate_async or generate_awaitable since it's not returning specifically a future.

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -276,7 +318,8 @@ def __init__(
                 client_wrapper.api_key = api_key
 
         self.client_wrapper = client_wrapper
-        self._client = self.client_wrapper.get_client()
+        self._client: OpenAIClient = self.client_wrapper.get_client()
```
eric-anderson, line 321: If we mark get_client and get_async_client with @functools.cache then we can get rid of both of these local variables and only calculate the client or async_client if they are used.

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -352,6 +395,63 @@ def _generate_using_openai(self, prompt_kwargs, llm_kwargs) -> str:
         completion = self._client.chat.completions.create(model=self._model_name, messages=messages, **kwargs)
         return completion.choices[0].message.content
 
+    async def generate_future(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> Awaitable[str]:
```
eric-anderson, line 398:
```
if llm_kwargs is None:
  raise ValueError(...)
```

and remove the check below.

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -352,6 +395,63 @@ def _generate_using_openai(self, prompt_kwargs, llm_kwargs) -> str:
...
+        cache_key = None
+        if self._cache:
+            cache_key = self._get_cache_key(prompt_kwargs, llm_kwargs)
+            hit = self._cache.get(cache_key)
```
eric-anderson, line 402:
```
assert hit.get("prompt_kwargs") == prompt_kwargs and hit.get("llm_kwargs") == llm_kwargs and hit.get("model_name") == self.model.name
```
No point in continuing if the cache is somehow broken and much shorter.

eric-anderson: I'd move this into a function and call it. Then generate can share the same code. So:
```
   (key, ret) = self._cache_get(prompt_kwargs, llm_kwargs)
   if ret is not None:
       return ret
  ...
   self._cache_set(key, result)
```

Then you can fix a caching bug.
```
   if llm_kwargs.get("temperature", 0) != 0:
       return (None, None)
```

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -352,6 +395,63 @@ def _generate_using_openai(self, prompt_kwargs, llm_kwargs) -> str:
...
+                "result": result,
+                "prompt_kwargs": prompt_kwargs,
+                "llm_kwargs": llm_kwargs,
+                "model_name": self.model.name,
+            }
+            self._cache.set(cache_key, item)
+        return result
+
+    async def _generate_future_using_openai(self, prompt_kwargs, llm_kwargs) -> Awaitable[str]:
+        kwargs = {
```
eric-anderson, line 434:
```
  kwargs = self._get_generate_kwargs(prompt_kwargs, llm_kwargs)
```
include messages in kwargs
and then share with _generate_using_openai

lib/sycamore/sycamore/tests/unit/transforms/test_graph_extractor.py
```diff
@@ -82,6 +83,10 @@ def __init__(self):
             super().__init__(model_name="mock_model")
 
         def generate(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None):
+            pass
+
+        async def generate_future(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None):
+            await asyncio.sleep(1)
```
eric-anderson, line 89: remove the sleep, I assume it was to help verify that it was waiting, but shouldn't be needed now and slows down tests.

lib/sycamore/sycamore/llms/llms.py
RitxmSaha: fixed!

lib/sycamore/sycamore/llms/openai.py
RitxmSaha: changed this!

lib/sycamore/sycamore/llms/openai.py
RitxmSaha: fixed

lib/sycamore/sycamore/llms/openai.py
RitxmSaha: changed -- implementation also does not store if the temperature is non zero.

lib/sycamore/sycamore/llms/openai.py
RitxmSaha: changed

lib/sycamore/sycamore/tests/unit/transforms/test_graph_extractor.py
RitxmSaha: fixed

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -295,61 +338,97 @@ def _get_cache_key(self, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None)
...
+        value = {
+            "result": ret,
+            "prompt_kwargs": prompt_kwargs,
+            "llm_kwargs": llm_kwargs,
+            "model_name": self.model.name,
+        }
+        self._cache_set(key, value)
+        return ret
+
+    async def _generate_future_using_openai(self, prompt_kwargs, llm_kwargs) -> Awaitable[str]:
```
eric-anderson, line 426: I'd rename this to _generate_awaitable_using_openai since it isn't strictly making a future.

eric-anderson (approved):

lib/sycamore/sycamore/llms/openai.py
RitxmSaha: oops! sorry missed that. changed.

PR #637, opened 2024-08-02
Refactor Context and support args in Map

baitsguy:
Remove global instance of Context, initial ray at execution, allow passing f_args in Map


baitsguy:
@eric-anderson the unit tests are currently breaking due to some CI infra issues, but figured you can look at the code at least

lib/sycamore/sycamore/context.py
```diff
@@ -1,155 +1,71 @@
...
 
-            ray.init(**ray_args)
-        elif self.exec_mode == ExecMode.LOCAL:
-            pass
-        else:
-            assert False, f"unsupported mode {self.exec_mode}"
+    """
+    Default OpenSearch configuration for a Context
+    """
+    opensearch_config: dict[str, Any] = field(default_factory=dict)
```
baitsguy, line 38: I can also make this a class, but unsure if that's overkill

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
```diff
@@ -115,12 +115,12 @@ def test_pdf_to_opensearch_with_llm_caching():
 
     try:
         context = sycamore.init(
-            config=Config(
-                opensearch_client_config=os_client_args,
-                opensearch_index_name="toyindex",
-                opensearch_index_settings=index_settings,
-                llm=openai_llm,
-            )
+            opensearch_config={
```
eric-anderson, line 118: This is where I thought it would be easier to write opensearch_config=OpenSearchConfig(os_client_args, "toyindex", index_settings)

If we had things that configure contexts having a context_varname property, then we could write this as just:
```
sycamore.init(openai_llm, OpenSearchArgs(os_client_args, "toyindex", index_settings)) 
```
and the positional args would be automatically put into kwargs via the context_varname.  Makes the code a little cleaner to use, unclear if it's too much magic.

lib/sycamore/sycamore/tests/unit/test_writer.py
```diff
@@ -120,10 +119,9 @@ def noop_map(doc: Document) -> Document:
 class TestDocSetWriter:
     def test_opensearch(self, mocker):
         context = mocker.Mock(spec=Context)
-        context.config = Config()
         docset = DocSet(context, mocker.Mock(spec=Node))
         execute = mocker.patch.object(OpenSearchWriter, "execute")
-        docset.write.opensearch(os_client_args={}, index_name="index")
+        docset.write.opensearch(os_client_args={}, index_name="index", index_settings={})
```
eric-anderson, line 124: why did you need to add index_settings here?  I feel like both that and os_client_args should default to empty dictionaries if unspecified.

lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -84,27 +81,31 @@ def __init__(
...
-    def extract_entity(self, document: Document) -> Document:
+    def extract_entity(self, document: Document, context: Optional[Context] = None) -> Document:
+        if self._llm is None and context:
+            self._llm = context.llm
+        assert self._llm is not None, "OpenAIEntityExtractor requires an LLM"
+
         if self._use_elements:
             if self._prompt_template:
-                entities = self._handle_few_shot_prompting(document)
+                entities = self._handle_few_shot_prompting(document, self._llm)
```
eric-anderson, line 91: why pass in self._llm to a method on self?

lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -170,6 +171,7 @@ def __init__(
         self,
         child: Node,
         entity_extractor: EntityExtractor,
+        context: Optional[Context] = None,
         **resource_args,
     ):
-        super().__init__(child, f=entity_extractor.extract_entity, **resource_args)
+        super().__init__(child, f=entity_extractor.extract_entity, f_args={"context": context}, **resource_args)
```
eric-anderson, line 177: I don't understand why you had to add f_args.
It looks like from the code you could have written:
```
super().__init__(child, f=entity_extractor.extract_entity, context=context, **resource_args)
```
The kwargs are being passed into the function already.  At a minimum they should be called f_kwargs, (f_args would be the positional ones), but I don't think they're needed.

I think you could have also written:
```f=lambda d: entity_extractor.extract_entity(d, context=context)```

File: lib/sycamore/sycamore/executor.py [hunk 6]
```diff
@@ -8,25 +9,69 @@
...
+
+                if "worker_process_setup_hook" not in ray_args["runtime_env"]:
+                    # logging.error("Spurious log 0: If you do not see spurious log 1 & 2,
+                    # log messages are being dropped")
+                    ray_args["runtime_env"]["worker_process_setup_hook"] = _ray_logging_setup
+
+                ray.init(**ray_args)
             return plan.execute(**kwargs)
-        if self._exec_mode == ExecMode.LOCAL:
+        elif self._exec_mode == ExecMode.LOCAL:
```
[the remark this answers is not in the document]
baitsguy: Interesting, I always stuck them in so it was explicit that only one path is being executed without having to read the other side. Will update

lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_pdf_to_opensearch.py
baitsguy: Yeah agree and that makes to me

lib/sycamore/sycamore/tests/unit/test_writer.py
baitsguy: This is a Mock artifact, without this it tries to read index_settings from the mocked context which it turn returns a mock. Since we aren't really testing the writer logic here, this was a shorthand

lib/sycamore/sycamore/transforms/extract_entity.py
baitsguy: Since the llm can now be set in __init__ and execute(..), it was a way to avoid having each llm access in the private function (_handle_xyz) be wrapped around a assert.self._llm is not None

Don't have strong opinions, this is less code

lib/sycamore/sycamore/context.py
```diff
@@ -1,155 +1,71 @@
...
-    global _global_context
-    with _context_lock:
-        if _global_context is None:
-            if ray_args is None:
-                ray_args = {}
-
-            # Set Logger for driver only, we consider worker_process_setup_hook
-            # or runtime_env/config file for worker application log
-            from sycamore.utils import sycamore_logger
+    if ray_args is None:
```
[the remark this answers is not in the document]
baitsguy: discussed offline, will nuke this in a separate PR to minimize issues

lib/sycamore/sycamore/transforms/extract_entity.py
baitsguy: Discussed offline, going to use a the 'kwargs' that already exists in BaseMap. Will update docstrings to clarify they exist in Map

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -65,9 +65,13 @@ class BaseMapTransform(UnaryNode):
...
     BaseMapTransform abstracts away MetadataDocuments from all other transforms.
 
-    If f is a class type, the class will be instantiated and run as an actor in ray.
+    If f is a class type, the class will be instantiated and run as an actor in ray. constructor_args and
+    constructor_kwargs can be used to provide arguments when initializing the class
+
     If f is an object type and resource_args["compute"] is set to ActorPoolStrategy, it will run as an actor
     Otherwise f will be run as a function.
+
+    Use args, kwargs to pass additional args to the function call.
```
eric-anderson, line 74: Either the __call__ method on the object or the function itself.

lib/sycamore/sycamore/transforms/map.py
```diff
@@ -10,6 +10,17 @@ class Map(BaseMapTransform):
     """
     Map is a transformation class for applying a callable function to each document in a dataset.
 
+    If f is a class type, constructor_args and constructor_kwargs can be used to provide arguments when
+    initializing the class
+
+    Use args, kwargs to pass additional args to the function call. The following 2 are equivalent:
```
eric-anderson, line 16: Either the direct function shown below, or the __call__ method on the class.

eric-anderson (approved):

PR #639, opened 2024-08-02
Refactor OpenSearchSchema to be more robust.

mdwelsh:
Should make retrieval of the schema more robust when there are mapping fields not present in the examples.

### DIFF
--- a/lib/sycamore/sycamore/query/planner.py
+++ b/lib/sycamore/sycamore/query/planner.py
@@ -1,4 +1,5 @@
-from typing import Any, Dict, List, Mapping, MutableMapping, Optional, Tuple, Type
+import json
+from typing import Dict, List, Mapping, MutableMapping, Optional, Tuple, Type
 
 from opensearchpy import OpenSearch
 
@@ -15,6 +16,7 @@ from sycamore.query.operators.sort import Sort
 from sycamore.query.operators.topk import TopK
 from sycamore.query.operators.limit import Limit
 from sycamore.query.operators.logical_operator import LogicalOperator
+from sycamore.query.schema import OpenSearchSchema
 from sycamore.utils.extract_json import extract_json
 
 
@@ -51,7 +53,7 @@ class LlmPlanner:
     def __init__(
         self,
         index: str,
-        data_schema: dict[str, Any],
+        data_schema: OpenSearchSchema,
         os_config: dict[str, str],
         os_client: OpenSearch,
         operators: Optional[List[Type[LogicalOperator]]] = None,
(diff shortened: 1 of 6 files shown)


lib/sycamore/sycamore/query/schema.py
```diff
@@ -21,15 +21,16 @@ def get_schema(self) -> dict[str, str]:
 
         for k, v in schema[self._index]["mappings"].items():
             if k.startswith("properties.entity") and ".keyword" not in k:
-                key = k
-                # key = k[18:]
-                result[key] = f"({type(random_sample[0]['_source']['properties']['entity'][k[18:]])}) e.g. "
+                try:
+                    key = k
+                    result[key] = f"({type(random_sample[0]['_source']['properties']['entity'][k[18:]])}) e.g. "
```
baitsguy, line 26: this will break the formatting if the first "random_sample" doesn't have the key I think?

lib/sycamore/sycamore/query/schema.py
```diff
@@ -21,15 +21,16 @@ def get_schema(self) -> dict[str, str]:
 
         for k, v in schema[self._index]["mappings"].items():
             if k.startswith("properties.entity") and ".keyword" not in k:
```
tranade, line 23: should we also add a check for if k has a "." in it? this could reduce the # of exceptions we have to catch

mdwelsh:
I reworked this PR so that it returns results in a more structured (and hence type-checkable) way, and defers how the schema is rendered into the LLM prompt to the planner code. Also added a test.

lib/sycamore/sycamore/tests/unit/query/test_schema.py
```diff
@@ -0,0 +1,67 @@
...
+    mock_query_executor = MagicMock()
+    # Note that there are no values for 'location' here.
+    mock_query_executor.query.return_value = {
+        "result": {
+            "hits": {
+                "hits": [
+                    {
+                        "_source": {
+                            "properties": {
+                                "entity": {
```
baitsguy, line 37: maybe have an entry that is missing a property

mdwelsh: Done.

mdwelsh:
OK, updated to support missing properties in some docs.

lib/sycamore/sycamore/query/planner.py
```diff
@@ -415,6 +427,7 @@ def generate_prompt(self, query):
         prompt += f"""
         USER QUESTION: {query}
         """
+        print(prompt)
```
tranade, line 430: do we want this here?

tranade (approved):

lib/sycamore/sycamore/query/planner.py
mdwelsh: Oops, good catch! Thanks!

PR #645, opened 2024-08-03
adding transforms and updating Ntsb demo notebooks

Soeb-aryn:


lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,121 @@
...
+    def standardize(self, document: Document) -> Document:
+        pass
+
+
+class LocationStandardizer(Standardizer):
+    """
+    This Class standardizes the format of US state location.
+    """
+
+    def __init__(self):
```
mdwelsh, line 22: This is fine for now but I wonder if an existing package like `address` could be used for this kind of thing.

https://pypi.org/project/address/

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,121 @@
+from abc import ABC, abstractmethod
+
+from sycamore.plan_nodes import Node
+
+from sycamore.data import Document
+from sycamore.transforms.map import Map
+from dateutil import parser
+
+
+class Standardizer(ABC):
```
mdwelsh, line 10: Needs docstring.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @abstractmethod
+    def standardize(self, document: Document) -> Document:
```
mdwelsh, line 13: Needs docstring.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def _standardize_state(self, state: str) -> str:
+        clean_state = state.lstrip().rstrip().upper()
+        return self._state_dict.get(clean_state, "")
+
+    def standardize(self, doc: Document) -> Document:
+        if "location" not in doc.properties["entity"]:
+            return doc
+        raw_location = doc.properties["entity"]["location"]
+        city, state = raw_location.split(",")
```
mdwelsh, line 85: I don't like this very much. It is going to be very brittle. "WA" can also mean Western Australia for example. And you should rarely assume that cities are always represented as "<City name>, <State abbreviation>". This might work in the very narrow case of the NTSB dataset that we have but it is far from general.

I would recommend either looking to see if there is an existing library that can do this better, or put a lot of caveats on the docstring on this class indicating it only works for locations expressed as "city, state" for US state abbreviations.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,121 @@
...
+        std_loc = f"{city}, {self._standardize_state(state)}"
```
mdwelsh, line 86: Are you OK with this throwing exceptions etc. if the state abbreviation can't be found in the dict? What about if the location string is not in the format you expect? What happens if the standardizer throws a ValueError or KeyError?

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,121 @@
...
+    This Class standardizes the format of dateTime.
+    Attributes:
+
+    """
+
+    def standardize(self, doc: Document) -> Document:
+        if "dateTime" not in doc.properties["entity"]:
+            return doc
+        raw_dateTime = doc.properties["entity"]["dateTime"]
+        raw_dateTime = raw_dateTime.replace("Local", "")
```
mdwelsh, line 102: This is also very brittle and only works in certain cases. That is probably OK for now as long as you document the limitations and ensure the code is robust to inputs not in the format you expect.

lib/sycamore/sycamore/docset.py
```diff
@@ -332,6 +333,24 @@ def spread_properties(self, props: list[str], **resource_args) -> "DocSet":
         plan = SpreadProperties(self.plan, props, **resource_args)
         return DocSet(self.context, plan)
 
+    def assign_doc_properties(self, element_type:str, property_name:str, **resource_args) -> "DocSet":
```
dtecuci, line 336: It seems that the source element is the first of the specified element_type. You should add that to the docstring.

lib/sycamore/sycamore/transforms/standardizer.py
baitsguy: +1 

I think if we're going with very specific implementations I'd rather have this as a function that a script will directly use `(docset.map(thisSpecificTypeOfStandarization)`

Top-level transforms and utilities need to be more resilient

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -52,21 +54,36 @@ def __init__(
...
             for idx, element in enumerate(document.elements):
-                elements[idx] = self._query_text_object(element)
-                if self._number_of_elements and idx >= self._number_of_elements:
-                    break
+                if self._element_type:
+                    if element.type == self._element_type:
+                        elements[idx] = self._query_text_object(element)
+                        if self._number_of_elements and element_type_count >= self._number_of_elements:
+                            break
+                        element_type_count += 1
```
karanataryn, line 69: Why do we need `element_type_count`?

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -52,21 +54,36 @@ def __init__(
...
+                    if self._number_of_elements and idx >= self._number_of_elements:
+                        break
             document.elements = elements
         elif self._number_of_elements:  # limit to a number of elements
             text_representation = self._prompt
             for idx, element in enumerate(document.elements):
-                text_representation += "\n" + element["text_representation"]
-                if idx >= self._number_of_elements:
-                    break
+                if self._element_type and element.type == self._element_type:
```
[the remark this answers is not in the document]
karanataryn: You can add a check for `_number_of_elements` in the first conditional itself

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,121 @@
...
+class Standardizer(ABC):
+
+    @abstractmethod
+    def standardize(self, document: Document) -> Document:
+        pass
+
+
+class LocationStandardizer(Standardizer):
+    """
+    This Class standardizes the format of US state location.
```
karanataryn, line 19: nit: US State abbreviations

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -85,7 +102,7 @@ def _query_text_object(self, object: Union[Document, Element]) -> Union[Document
                     .render(doc=object)
                 )
             else:
-                prompt = self._prompt + "\n" + object["text_representation"]
+                prompt = self._prompt + "\n" + object.text_representation
```
karanataryn, line 105: Is there a need for this? Would prefer it being standardized

karanataryn (commented):
Few comments, please make sure to clear the outputs in the demo notebook, I don't think it's been done yet

lib/sycamore/sycamore/transforms/assign_doc_properties.py
```diff
@@ -0,0 +1,52 @@
...
+
+class AssignDocProperties(SingleThreadUser, NonGPUUser, Map):
+    """
+    The AssignDocProperties transform is used to copy properties from specific element stored
+    within a JSON string to the parent document. This allows for the consolidation of key
+    attributes at the document level.
+
+    Args:
+        child: The source node or component that provides the hierarchical documents to be exploded.
+        resource_args: Additional resource-related arguments that can be passed to the explosion operation.
```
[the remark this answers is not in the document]
Soeb-aryn: it is not, I will update the description, thanks Matt for catching this.

lib/sycamore/sycamore/transforms/assign_doc_properties.py
```diff
@@ -0,0 +1,52 @@
...
+            return json.loads(s[: e.pos])
+
+    @timetrace("AssignProps")
+    @staticmethod
+    def assign_doc_properties(parent: Document, element_type: str, property_name: str) -> Document:
+        # element count is zero indexed
+        assert property_name is not None
+        for e in parent.elements:
+            if e.type == element_type:
+                property_str = e.properties.get(property_name)
```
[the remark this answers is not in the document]
Soeb-aryn: This Transform parses value of JSON string from an element to parent document. After discussing with Dan yesterday we have decided to repurpose this transforms and create another transform for parsing properties from table as following 
1. extract_key_value_pair -> this transform will extract value of table and save it as a property for each table element. 
2. assign_doc_property -> this transform will copy property of a particular element to parent.

lib/sycamore/sycamore/transforms/assign_doc_properties.py
```diff
@@ -0,0 +1,52 @@
...
+            property_transform = AssignDocProperties(child=source_node, list=["table"])
+            property_dataset = property_transform.execute()
+    """
+
+    def __init__(self, child: Node, parameters: List[str], **resource_args):
+        super().__init__(child, f=AssignDocProperties.assign_doc_properties, args=parameters, **resource_args)
+
+    @staticmethod
+    def _parse_json(s: str) -> Dict:
+        s = s[next(idx for idx, c in enumerate(s) if c in "{[") :]
```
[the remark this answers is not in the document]
Soeb-aryn: this might not be needed now, in the repurposed [assign_doc_properties.py](https://github.com/aryn-ai/sycamore/pull/645/files/c0561a0edcf720fa7d985a1f42e7a571520f9f13#diff-4df39ca3f1bfd8b62fc663fac55bb7bb84473d54dc2a2165ed85277387e36c2f) as in the newer implementation we are copying properties directly from child to parent.

lib/sycamore/sycamore/tests/unit/transforms/test_assign_doc_properties.py
```diff
@@ -0,0 +1,39 @@
...
+                        "type": "title",
+                        "content": {"binary": None, "text": "text2"},
+                        "properties": {'property1': '{"test1": "from element2"}', "page_number": 2},
+                    },
+                ],
+            }
+        )
+
+    def test_assign_doc_propoerties(self):
+        output = AssignDocProperties(None, ["title", "property1"]).run(self.input)
```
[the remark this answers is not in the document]
Soeb-aryn: Repurposing this transform  to not do two things of parsing json and assigning properties to parent. These tests wonts be needed.

lib/sycamore/sycamore/transforms/standardizer.py
Soeb-aryn: I have rewritten the method and handled KeyError. The transform assign full name to abbreviation, I am make the name and docString more descriptive to avoid confusion.

lib/sycamore/sycamore/tests/unit/transforms/test_standardizer.py
```diff
@@ -0,0 +1,43 @@
...
+                "properties": {"path": "s3://path",
+                               "entity":{
+                                   "location":"Mountain View, CA",
+                                   "dateTime": "March 17, 2023, 14.25 Local"
+                                   }
+                               }
+            }
+        )
+
+    def test_datetime(self):
```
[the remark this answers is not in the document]
Soeb-aryn: added more test cases.

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -52,21 +54,36 @@ def __init__(
...
 
     def execute_query(self, document: Document) -> Document:
         if self._per_element:
             elements = document.elements
+            element_type_count = 0
             for idx, element in enumerate(document.elements):
-                elements[idx] = self._query_text_object(element)
-                if self._number_of_elements and idx >= self._number_of_elements:
-                    break
+                if self._element_type:
```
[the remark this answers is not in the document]
Soeb-aryn: thnx,  I have made changes to the logic to make it less repetitive using your suggestion.

lib/sycamore/sycamore/docset.py
Soeb-aryn: thanks for catching this, Fixing it in the next push.

lib/sycamore/sycamore/transforms/llm_query.py
Soeb-aryn: text_representation is property, and thought it will be more robust to use if object is missing text representation.

lib/sycamore/sycamore/transforms/extract_key_value_pair.py
```diff
@@ -0,0 +1,131 @@
...
+import json
+from typing import List, Union
+import logging
+from sycamore.transforms.llm_query import LLMTextQueryAgent
+from sycamore.llms import OpenAI
+from sycamore.llms import LLM
+
+class ExtractKeyValuePair(SingleThreadUser, NonGPUUser, Map):
+    """
+    The ExtractKeyValuePair transform extracts the key value from tables and add it as properties to it,
```
mdwelsh, line 14: This is a little hard to parse. Maybe give a short example?

lib/sycamore/sycamore/transforms/extract_key_value_pair.py
```diff
@@ -0,0 +1,131 @@
...
+            property_transform = AssignDocProperties(child=source_node, list=["table"])
+            property_dataset = property_transform.execute()
+    """
+
+    def __init__(self, child: Node, parameters:list[Union[str, LLM]], **resource_args):
+        super().__init__(child, f=ExtractKeyValuePair.extract_table_properties, args=parameters, **resource_args)
+
+    @staticmethod
+    def extract_parent_json(input_string: str) -> str:
+        '''
```
mdwelsh, line 34: Use `"""` for docstrings. I assume black would do this?

lib/sycamore/sycamore/transforms/extract_key_value_pair.py
```diff
@@ -0,0 +1,131 @@
...
+                    assert isinstance(jsonstring_llm, str)
+                    json_string = ExtractKeyValuePair.extract_parent_json(jsonstring_llm)
+                    assert isinstance(json_string, str)
+                    keyValue = json.loads(json_string) 
+                    ele.properties[property_name] = keyValue
+                except Exception as e:
+                    logging.error(str(e))
+        return doc
+
+if __name__ == "__main__":
```
mdwelsh, line 93: I assume the rest of this code should not be here?

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -22,16 +22,17 @@ class LLMTextQueryAgent:
         llm_kwargs: (Optional) LLM keyword argument for the underlying execution engine
         per_element: (Optional) Whether to execute the call per each element or on the Document itself. Defaults to
         True.
+        element_type: (Optional) Element type based filter to make call for only specific element type. Defaults to Any
```
mdwelsh, line 25: I think we need to be more clear on what this means. It's a string. What values can it take? Should we use a python `typing.Type` for this instead?

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -44,6 +45,7 @@ def __init__(
         number_of_elements: Optional[int] = None,
         llm_kwargs: dict = {},
         per_element: bool = True,
+        element_type: str = "",
```
mdwelsh, line 48: I don't think an empty string is the right default value if you want to allow anything. Probably you want `Optional[str] = None`.

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -52,28 +54,63 @@ def __init__(
...
+                document["properties"][self._output_property] = llm_resp
         else:
             if document.text_representation:
                 document = self._query_text_object(document)
         return document
+            
+
+
+
+        # for idx, element in enumerate(document.elements):
```
mdwelsh, line 86: Remove dead code.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+from sycamore.data import Document
+from sycamore.transforms.map import Map
+from dateutil import parser
+import re
+from typing import List
+
+
+class Standardizer(ABC):
+    """
+    An Abstrant class for implementing Standarizers.
```
mdwelsh, line 14: Typo

lib/sycamore/sycamore/transforms/standardizer.py
mdwelsh: Needs more detail. What is a "Standardizer" and why is it different than a regular map?

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+
+    @abstractmethod
+    def standardize(self, doc: Document, key_path: List[str]) -> Document:
```
mdwelsh, line 18: This needs a docstring.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+            "WI": "Wisconsin",
+            "WV": "West Virginia",
+            "WY": "Wyoming",
+        }
+
+    def replace_abbreviations(self, input_string):
+        """
+        This method replaces the US State abbreviations with full names.
+        """
+        state_abbreviations = {key: value for key, value in self._state_abbreviations.items()}
```
mdwelsh, line 86: All you're doing is making a copy here. Why?

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+        pass
+
+
+class LocationStandardizer(Standardizer):
+    """
+    This Class standardizes the format of US State abbreviations.
+    """
+
+    def __init__(self):
+        self._state_abbreviations = {
```
mdwelsh, line 28: This should be a class level constant, i.e., `US_STATE_ABBREVIATIONS`.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+    def replace_abbreviations(self, input_string):
+        """
+        This method replaces the US State abbreviations with full names.
+        """
+        state_abbreviations = {key: value for key, value in self._state_abbreviations.items()}
+        abbreviations = {**state_abbreviations}
+
+        pattern = re.compile(r"\b(" + "|".join(re.escape(key) for key in abbreviations.keys()) + r")\b")
```
mdwelsh, line 89: I don't think a simple string replacement requires a huge regexp. This is also a really expensive way to do things, in particular because you're compiling the RE on every call to this function.

Your regexp is not correct, I think, because it would match "CAlifornia" and turn it into "Californialifornia", right?

A better approach is needed here. I would recommend a single regexp to identify two-letter capital sequences (`[A-Z][A-Z]`) that do not have other letters or numbers before or after them. Then extract that, do a lookup in the abbreviations dict, and do a string replace at the index where the regexp matched.

There may be better approaches as well, which is one reason I suggested seeking out a library for this so we don't end up with something brittle.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+
+        def replace_match(match):
+            return abbreviations[match.group(0)]
+
+        result_string = pattern.sub(replace_match, input_string)
+        return result_string
+
+    def standardize(self, doc: Document, key_path: List[str]) -> Document:
+        """
+        This Methods creates a new element standardises the location property of the Element
```
mdwelsh, line 99: Please re-read this docstring and fix the grammar and spelling.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class DateTimeStandardizer(Standardizer):
+    """
+    This Class standardizes the format of dateTime.
+    """
+
+    def fix_date(self, raw_dateTime):
+        """
+        This Method standardizes the datetime property of Elements by replcaing . with : and parsing date as Date
```
mdwelsh, line 123: Typo.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+
+        raw_dateTime = raw_dateTime.replace("Local", "")
+        raw_dateTime = raw_dateTime.replace(".", ":")
+        parsed_date = parser.parse(raw_dateTime)
+        extracted_date = parsed_date.date()
+        return raw_dateTime, extracted_date
+
+    def standardize(self, doc: Document, key_path) -> Document:
+        """
+        This Methods creates a new element with Date and standardises the dateTime property of the Element
```
mdwelsh, line 134: Please do a better job reviewing the grammar of your docstrings.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+
+        result_string = pattern.sub(replace_match, input_string)
+        return result_string
+
+    def standardize(self, doc: Document, key_path: List[str]) -> Document:
+        """
+        This Methods creates a new element standardises the location property of the Element
+        by replcaing all instance of US State abbreviation with correct name.
+        """
+        current = doc
```
mdwelsh, line 102: This block of code is duplicated across both standardizers. Should this be part of a parent class or something?

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+            raise KeyError(f"Key {target_key} not found in the dictionary among {current.keys()}")
+        return doc
+
+
+class DateTimeStandardizer(Standardizer):
+    """
+    This Class standardizes the format of dateTime.
+    """
+
+    def fix_date(self, raw_dateTime):
```
mdwelsh, line 121: Needs type annotations.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+            "UT": "Utah",
+            "VA": "Virginia",
+            "VT": "Vermont",
+            "WA": "Washington",
+            "WI": "Wisconsin",
+            "WV": "West Virginia",
+            "WY": "Wyoming",
+        }
+
+    def replace_abbreviations(self, input_string):
```
mdwelsh, line 82: Needs type annotations

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -52,28 +54,63 @@ def __init__(
...
-        elif self._number_of_elements:  # limit to a number of elements
-            text_representation = self._prompt
-            for idx, element in enumerate(document.elements):
-                text_representation += "\n" + element["text_representation"]
-                if idx >= self._number_of_elements:
-                    break
-            prompt_kwargs = {"prompt": text_representation}
-            llm_resp = self._llm.generate(prompt_kwargs=prompt_kwargs, llm_kwargs=self._llm_kwargs)
-            document["properties"][self._output_property] = llm_resp
+                if self._element_type and element.type != self._element_type:
```
karanataryn, line 64: You need to add the `else` block here. Otherwise you are not doing anything with the `element_type`

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -52,28 +54,63 @@ def __init__(
...
+                    continue
+                if self._per_element:
+                    document.elements[idx] = self._query_text_object(element)
+                else:
+                    final_prompt += "\n" + element["text_representation"]
+                if self._number_of_elements:
+                    element_count += 1
```
karanataryn, line 71: Why do you need `element_count` here? Can't you check this with the `idx`?

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,162 @@
...
+        return raw_dateTime, extracted_date
+
+    def standardize(self, doc: Document, key_path) -> Document:
+        """
+        This Methods creates a new element with Date and standardises the dateTime property of the Element
+        """
+
+        current = doc
+        for key in key_path[:-1]:
+            if key in current.keys():
```
karanataryn, line 139: Can optimize the number of lookups by doing `current.get(key, None)` and then checking if the value is None

lib/sycamore/sycamore/transforms/extract_key_value_pair.py
```diff
@@ -0,0 +1,131 @@
...
+            property_dataset = property_transform.execute()
+    """
+
+    def __init__(self, child: Node, parameters:list[Union[str, LLM]], **resource_args):
+        super().__init__(child, f=ExtractKeyValuePair.extract_table_properties, args=parameters, **resource_args)
+
+    @staticmethod
+    def extract_parent_json(input_string: str) -> str:
+        '''
+        Extracts the top level jsonstring from the string passed.
```
karanataryn, line 35: Typo: `JSON string`

lib/sycamore/sycamore/transforms/llm_query.py
Soeb-aryn: for use cases like: process only 4 top table elements, idx won't work

lib/sycamore/sycamore/transforms/llm_query.py
karanataryn: Nvm I see the rationale

lib/sycamore/sycamore/tests/unit/transforms/test_extract_key_value_pair.py
```diff
@@ -0,0 +1,46 @@
...
+                    "bbox": (1, 2, 3, 4.0),
+                    "properties": {"title":  {"rows": None, "columns": None}},
+                    "table": self.table('key1','val1'),
+                    "tokens": None,
+                },
+                
+            ],
+            "properties": {"int": 0, "float": 3.14, "list": [1, 2, 3, 4], "tuple": (1, "tuple")},
+        })
+        # print(self.doc)
```
mdwelsh, line 41: Remove dead code.

lib/sycamore/sycamore/transforms/extract_key_value_pair.py
mdwelsh: Ping

lib/sycamore/sycamore/transforms/standardizer.py
mdwelsh: ping

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,165 @@
...
+from typing import List, Tuple, Union
+from datetime import date
+
+
+class Standardizer(ABC):
+    """
+    An abstract class for implementing standardizers.
+    """
+
+    @abstractmethod
```
mdwelsh, line 18: Need docstring

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -0,0 +1,165 @@
...
+    def fixer(self, text: str) -> Union[str, Tuple[str, date]]:
+        pass
+
+    def standardize(self, doc: Document, key_path: List[str]) -> Document:
```
mdwelsh, line 22: Need docstring

mdwelsh (approved):
Thank you for your diligence on this. Please address the few remaining nits before submitting!

dtecuci (approved):

lib/sycamore/sycamore/transforms/standardizer.py
karanataryn: Think you might have missed this!

karanataryn (approved):
LGTM! Added a minor comment

dtecuci (approved):

PR #646, opened 2024-08-03
Fix code execution and trace display in Query UI

tranade:
- Button to execute generated code and result is printed out (this will also allow us to test/tweak codegen)
- Traces after each node displayed as scrollable/expandable/searchable pandas dataframe
- Button to download traces for a query as a zip file
- Added some error handling

If you want to switch to the branch + try it out and give feedback, that would be great! I can also demo it on Monday.

image

### DIFF
--- a/apps/query-ui/queryui/queryui.py
+++ b/apps/query-ui/queryui/queryui.py
@@ -64,11 +64,8 @@ def show_traces():
                     # print(doc.properties['LlmFilterOutput'])
 
                     for p in doc.properties:
-                        if p not in BASE_PROPS:
-                            try:
-                                doc_list[p] = doc.properties[p]
-                            except:
-                                doc_list[p] = None
+                        if p not in BASE_PROPS and p in doc.properties:
+                            doc_list[p] = doc.properties[p]
 
                     data_list.append(doc_list)
 
@@ -90,6 +87,7 @@ def generate_code(client, plan):
             language="python",
             min_lines=20,
         )
+
 
 def show_dag(plan: LogicalPlan):
     nodes = []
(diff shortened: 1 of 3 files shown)


apps/query-ui/queryui/queryui.py
```diff
@@ -18,13 +18,22 @@
...
 DEFAULT_S3_CACHE_PATH = "s3://aryn-temp/llm_cache/luna/ntsb"
-
-
-def execute(code: str):
-    try:
-        exec(code, globals(), globals())
-    except Exception as e:
-        st.exception(e)
+DEFAULT_TRACE_DIR = "luna_traces"
+DEFAULT_FULL_TRACE_DIR = f"{os.path.dirname(os.getcwd())}/query-ui/luna_traces"
```
mdwelsh, line 22: I don't want the demo UI to scribble onto the local directory. We should always use tempdir.mkdtemp() for this, which is what the current code is doing - can't you retain the use of mkdtemp and set st.session_state.trace_dir to the directory name that is generated?

apps/query-ui/queryui/queryui.py
```diff
@@ -97,14 +124,17 @@ def show_dag(plan: LogicalPlan):
 
 def run_query(query: str, index: str, plan_only: bool, do_trace: bool, use_cache: bool):
     """Run the given query."""
-    trace_dir = None
+    st.session_state.trace_dir = None
     if do_trace:
-        trace_dir = tempfile.mkdtemp()
-        st.write(f"Writing execution traces to `{trace_dir}`")
+        if not plan_only:
+            st.session_state.trace_dir = DEFAULT_TRACE_DIR
```
mdwelsh, line 130: See comment above, I think you can just continue using tempfile.mkdtemp() for this.

mdwelsh (approved):

PR #648, opened 2024-08-05
Add generic traverse

eric-anderson:
Handles both before and after transforms and properly handles adjustments to the structure.


bsowell (approved):
I'm still not entirely sure I understand all the implications here, but this seems reasonable for the materialize use case. Seems likely maybe we should add clone to the list of things to implement.

PR #649, opened 2024-08-05
A couple of small tweaks to make Sycamore more robust to missing or bogus data.

mdwelsh:


HenryL27 (approved):
sure

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -60,7 +61,10 @@ def math_operation(val1: int, val2: int, operator: str) -> Union[int, float]:
     elif operator == "subtract":
         return val1 - val2
     elif operator == "divide":
-        return val1 / val2
+        try:
+            return val1 / val2
+        except ZeroDivisionError:
+            return math.nan
```
bsowell, line 67: When is math.nan useful, or is the primary benefit suppressing the exception?

bsowell (approved):

lib/sycamore/sycamore/query/execution/operations.py
mdwelsh: Well, we want *some* return value here, and it shouldn't be 0 or math.inf I think.

PR #650, opened 2024-08-05
Neo4j Writer

RitxmSaha:


lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+        nodes = defaultdict(list)
+        relationships = defaultdict(lambda: defaultdict(list))
+        for doc in docs:
+            if "label" not in doc.data or "doc_id" not in doc.data:
+                continue
+            node = {
+                "uuid:ID": doc.doc_id,
+            }
+            if doc["type"] == "table":
+                doc["text_representation"] = doc["table"].to_csv()
```
HenryL27, line 262: sometimes tables are None, so this is a little spicy. Also there's a key ('_override_text') that gets used with TableElements when you set the text_representation (to something other than table.to_csv() - see the implementation of TableElement.text_representation to see how that works and how to make this work with that. You can also change that implementation if you have a better way to get the same behavior)

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+            for key, value in doc.data.items():
+                # add flatten properties to node
+                if key == "properties":
+                    if isinstance(value, dict) and value:
+                        properties_flat = pd.json_normalize(value, sep="_").to_dict(orient="records")[0]
```
HenryL27, line 267: There's a fn in sycamore.connectors.common called flatten_data (or something like that) that does this without sticking it in a pandas DataFrame first

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+            headers = None
+            with open(csv_path, mode="r", newline="") as file:
+                csv_reader = csv.reader(file)
+                headers = sorted(set(next(csv_reader)))
+
+            with open(csv_path, "a", newline="") as file:
+                writer = csv.writer(file)
+                for entry in rels:
+                    row = [entry.get(header, "") for header in headers]
+                    writer.writerow(row)
```
HenryL27, line 302: q: is there a race condition here that could give you weird spliced lines?

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+            with TimeTrace("UnknownWriter"):
+                return self.write_docs(docs)
+
+
+class Neo4jLoadCSV:
+    def __init__(self, client_params: Neo4jWriterClientParams, target_params: Neo4jWriterTargetParams, **kwargs):
+        self._client = Neo4jWriterClient.from_client_params(client_params)
+        self._nodes = [f for f in os.listdir(client_params.import_dir + "/sycamore/nodes")]
+        self._relationships = [f for f in os.listdir(client_params.import_dir + "/sycamore/relationships")]
+        self._labels = [f[:-4] for f in self._nodes]
```
HenryL27, line 342: where'd the :-4 come from?

HenryL27: oh. dropping `.csv`? Add a comment explaining if so and then don;t you also need to do that for the relationships?

lib/sycamore/sycamore/writer.py
```diff
@@ -562,6 +564,76 @@ def elasticsearch(
...
+        """
+        import os
+        from sycamore.connectors.neo4j import (
+            Neo4jWriterClientParams,
+            Neo4jWriterTargetParams,
+            Neo4jValidateParams,
+        )
+        from sycamore.plan_nodes import Node
+        from sycamore.connectors.neo4j import Neo4jPrepareCSV, Neo4jWriteCSV, Neo4jLoadCSV
+
```
HenryL27, line 617: add:
```python
if kwargs.get('execute') == False:
    raise NotImplementedError
```

HenryL27 (commented):
I'd love to see a IT or example and prob some unit tests too for some of the more complicated bits of logic

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+        def accumulate_row(headers, row):
+            data = extract_nodes(row)
+            if "label" not in data:
+                return headers
+            node_key = data["label"]
+            if headers["nodes"].get(node_key, None) is None:
+                headers["nodes"][node_key] = dict()
+            #### add all required keys ####
+            #### ALL DOCS HAVE uuid:ID ####
+            headers["nodes"][node_key]["uuid:ID"] = True
```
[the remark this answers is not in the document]
RitxmSaha: I tried, ray was not a fan and converted it to a list between aggregation steps. I could use a list here, but in the case that duplicate headers are found(they will), i would have to search through the entire list.

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
RitxmSaha: yep, looks like there could be a race condition. Have not run into one thus far but should protect against it. Will use filesystem lock ```fcntl.lockf``` to deal with this.

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
RitxmSaha: yeah dropping the .csv. I do this for labels specifically because for fast performance i implemented neo4j is constraints over node labels. Currently have no need to do this for relationships as I don't really need to index over uuid for relationships.

RitxmSaha: will update with comment

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+        self._client = Neo4jWriterClient.from_client_params(client_params)
+        self._client.create_target_idempotent(target_params=target_params)
+        self._check_write_permissions(client_params=client_params)
+        self._client._driver.close()
+
+    def _check_write_permissions(self, client_params: Neo4jWriterClientParams):
+        path = client_params.import_dir
+        # Check read permissions
+        if not os.access(path, os.R_OK):
+            raise OSError(f"Read permission denied for directory: {path}")
```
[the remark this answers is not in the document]
RitxmSaha: we do because we read the header from csv's after we write them.

lib/sycamore/sycamore/writer.py
RitxmSaha: fixed

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
RitxmSaha: updated

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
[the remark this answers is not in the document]
HenryL27: weird. ok

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+        self._driver = driver
+        self._import_dir = import_dir
+
+    @classmethod
+    def from_client_params(cls, params: Neo4jWriterClientParams) -> "Neo4jWriterClient":
+        try:
+            driver = GraphDatabase.driver(uri=params.uri, auth=params.auth)
+            driver.verify_connectivity()
+        except Exception:
+            raise ValueError("Invalid Neo4j URI or Authentication was used")
```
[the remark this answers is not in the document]
RitxmSaha: fixed!

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+        return Neo4jWriterClient(driver, params.import_dir)
+
+    def create_target_idempotent(self, target_params: BaseDBWriter.TargetParams):
+        assert isinstance(target_params, Neo4jWriterTargetParams)
+        try:
+            session = self._driver.session(database=target_params.database)
+            session.close()
+            return
+        except Exception as e:
+            raise e
```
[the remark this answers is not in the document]
RitxmSaha: fixed!

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
RitxmSaha: okay after looking at implementation, it will work fine without the table check at all. removed

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+            #### add all required keys ####
+            #### ALL DOCS HAVE uuid:ID ####
+            headers["nodes"][node_key]["uuid:ID"] = True
+            #### IF KEYS EXIST IN DATA ####
+            for key in ["type", "bbox", "text_representation"]:
+                if key in data:
+                    headers["nodes"][node_key][key] = True
+            #### add all keys from properties ####
+            for key in data["properties"].keys():
+                headers["nodes"][node_key][key] = True
```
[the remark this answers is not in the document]
RitxmSaha: fixed, also used the flatten_data from sycamore.connectors.commons.

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
RitxmSaha: fixed, no longer use pandas

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+            finalize=finalize,
+            name="output",
+        )
+
+        return self._dataset.aggregate(aggregation)["output"]
+
+    def _write_nodes_csv_headers(self, nodes):
+        for node_label, node_columns in nodes.items():
+            csv_path = os.path.expanduser(f"{self._import_dir}/sycamore/nodes/{node_label}.csv")
+            headers = sorted(set(column for column in node_columns.keys()))
```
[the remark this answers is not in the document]
RitxmSaha: its easiest to make this a list here. We need to mantain consistent order of the headers otherwise it will right values to the wrong column. sorted(set()) solves this but its slow, we can just use a list since all the columns will have unique values after the aggregate.

switched to using list.

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -0,0 +1,343 @@
...
+            return headers
+
+        def merge(headers1, headers2):
+            #### merge nodes together ####
+            for key, values in headers2["nodes"].items():
+                if headers1["nodes"].get(key, None) is None:
+                    headers1["nodes"][key] = values
+                else:
+                    for value in values:
+                        headers1["nodes"][key][value] = True
```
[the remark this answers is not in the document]
RitxmSaha: added

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
RitxmSaha: take it back. does not work. So whats happening is that this is actually no longer an element. To represent it in a hierarchical way, I took the element data and stuck it into a hierarchical document. I would like to extract this away from the user though, so I'll make a change in the hierarchical document for now that supports this functionality. Worth a conversation on the data model later on though.

HenryL27 (approved):

PR #652, opened 2024-08-06
Run poetry-lock-all until the dependencies became consistent.

eric-anderson:
I was seeing upgrades and downgrades recently on updating sycamore.

Add a permanent dependency on the shared config for apps/query-ui. Add a temporary dependency on the shared dependency for aryn-sdk.


HenryL27 (approved):
lg but unit tests fail with inconsistent lock file

PR #653, opened 2024-08-06
Switch Luna tracing to use materialize.

mdwelsh:
This PR migrates Luna's execution tracing to use the new materialize support.


lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -181,22 +178,24 @@ def process_node(self, logical_node: LogicalOperator, query_id: str) -> Any:
...
+
+        if self.dry_run:
+            code, imports = operation.script()
+            self.imports += imports
+            self.node_id_to_code[logical_node.node_id] = code
+            self.node_id_to_node[logical_node.node_id] = logical_node
+            result = None
+        else:
+            result = operation.execute()
+            if trace_dir and hasattr(result, "materialize"):
```
tranade, line 196: do all docsets have materialize? wondering if we can get this to write out traces after a top k operation

apps/query-ui/queryui/queryui.py
```diff
@@ -46,34 +46,40 @@ def show_schema(container: Any, schema: Dict[str, Tuple[str, Set[str]]]):
...
-                        if p not in BASE_PROPS and p in doc.properties:
-                            doc_list[p] = doc.properties[p]
-                    doc_list["text_representation"] = doc.text_representation
-                    data_list.append(doc_list)
+                    # For now, skip over MetadataDocuments.
+                    if "doc_id" not in doc:
+                        continue
+                    data_list.append(doc)
 
         df = pd.DataFrame(data_list)
```
tranade, line 64: is it possible to let each doc.properties and doc.properties.entity have its own column? helps when debugging so you can tell if something is wrong from a quick glance

lib/sycamore/sycamore/query/execution/sycamore_executor.py
mdwelsh: Yes, it is a top-level docset operation.

apps/query-ui/queryui/queryui.py
mdwelsh: Yes, we can tweak it, but I am working on a new UI for this anyway.

lib/sycamore/sycamore/query/execution/sycamore_executor.py
tranade: Hmm, it isn't writing out traces for a docset after top k (when I checked locally)

tranade (approved):

PR #655, opened 2024-08-07
Fix codegen syntax/formatting

baitsguy:
Fixed some typos, formatting, and structure. Ran the queries successfully through the benchmark

### DIFF
--- a/lib/sycamore/sycamore/query/execution/sycamore_operator.py
+++ b/lib/sycamore/sycamore/query/execution/sycamore_operator.py
@@ -444,8 +444,8 @@ class SycamoreLlmExtractEntity(SycamoreOperator):
         discrete = logical_node.discrete
         assert logical_node.dependencies is not None and len(logical_node.dependencies) == 1
 
-        input_str = input_var or get_var_name(self.logical_node.dependencies[0])
-        output_str = output_var or get_var_name(self.logical_node)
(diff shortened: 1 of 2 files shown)


tranade (approved):

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -299,34 +300,38 @@ def script(self, input_var: Optional[str] = None, output_var: Optional[str] = No
...
-                f"field='{field}',\n"
-                f"start='{start}',\n"
-                f"end='{end}',\n"
-                f"date='{date}),'\n"
-                f"**{self.get_node_args()})"
-            )
+            script = f"""
+{output_str} = {input_str}.filter(
+    f=RangeFilter(
+        field='{field}',
```
mdwelsh, line 316: Note that start and end values might be ints (for example) so wrapping them in quotes is probably not what you want to do in those cases. We should probably have a utility function to render parameters being passed into the generated code that handles things like that.

baitsguy: Realized we're going to need input/output schemas (or types) for these operators - currently there isn't a clear way to easily determine what the type of input data is. We had plans for them but seems like now is a good time to add them

PR #656, opened 2024-08-07
fix table html parsing edge case

HenryL27:
old algorithm:
for each cell, if I am a rowspan, check all the other cells and bump everything below and to the right of me.

new algorithm:
for each cell, if I need to be bumped, bump myself, and if I'm a rowspan tell the rows below me that there might need to be bumping.

old algorithm failed on a table like
```
__________
|__|  |__|
|  |  |__|
|__|__|__|
```
because the cell on the bottom right was initially in col=0, and the middle cell wouldn't bump it


karanataryn (approved):
Lgtm

PR #657, opened 2024-08-07
Code executor

baitsguy:
Allows you to run sycamore queries through generated code. This will become the default once we've tested existing workloads. 

We might also want to have the codegen mode return the result + generated code in 1 call, can think about it as we migrate


mdwelsh (approved):
noice

tranade (commented):
Does this require changes in the query UI?

baitsguy:
> Does this require changes in the query UI?

Shouldn't, the execute interface stays the same

PR #658, opened 2024-08-07
Codegen with traces in UI

tranade:
Fixes that allow generated code to produce traces in UI
Also will add updates so generated code can be tweaked and re-run


baitsguy (approved):

tranade (commented):
Made some changes, wanted to make sure I'm not breaking anything

PR #659, opened 2024-08-07
Revamp Sycamore Query demo UI.

mdwelsh:
This PR fixes a bunch of issues with the Sycamore Query UI around re-running queries, caching, etc. It should be faster and more robust.

It also makes the UI a multipage app, with one page providing the query interface, and another providing a debug trace viewer. This way we can add additional pages as needed later on.

Screenshot 2024-08-07 at 3 44 56 PM

Screenshot 2024-08-07 at 3 45 26 PM


baitsguy (approved):
Might be some nuances that get exposed by using but code looks good

PR #662, opened 2024-08-09
Fixing the title for introduction page and making the 'specifying options' section its own page for APS docs

AbhijitP-009:


jonfritz (approved):
LGTM. Thx!

PR #667, opened 2024-08-09
Add support for stdin for pdf_to_images.

austintlee:


lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -125,12 +125,18 @@ def convert_from_path_streamed_batched(filename: str, batch_size: int) -> Genera
...
 
 
-def pdf_to_image_files(pdf_path: str, file_dir: Path) -> Generator[Path, None, None]:
+def pdf_to_image_files(pdf_path: str, file_dir: Path, fh = None, pages = None) -> Generator[Path, None, None]:
     """Writes the files (streamed) into file_dir.  Caller is responsible for calling
     path.unlink() to cleanup the files.
 
     Note: model service will call this to get images for processing"""
 
+    if pdf_path == '-' and fh is None:
```
baitsguy, line 134: what's this for? we can just make the arg optional?

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -125,12 +125,18 @@ def convert_from_path_streamed_batched(filename: str, batch_size: int) -> Genera
         yield batch
 
 
-def pdf_to_image_files(pdf_path: str, file_dir: Path) -> Generator[Path, None, None]:
+def pdf_to_image_files(pdf_path: str, file_dir: Path, fh = None, pages = None) -> Generator[Path, None, None]:
```
baitsguy, line 128: can you add a type + docstring for fh and pages?

lib/sycamore/sycamore/utils/pdf.py
austintlee: https://linux.die.net/man/1/pdftoppm - pdftoppm accepts '-' as stdin which is a standard thing unix utilities accept.

baitsguy: Interesting, til

lib/sycamore/sycamore/utils/pdf.py
eric-anderson, line 128: Do pdf_path: Union[str, io.IOBase]
That way callers only have to specify a single parameter rather than two related ones.

eric-anderson (commented):
Add explanation for why you're making this change. Also make sure to note that you're also adding pages to pdf_to_images.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -195,7 +204,10 @@ def read_stderr(fh, q):
         # If we don't have the separate threads for reading stdout/stderr,
         # then if the stderr buffer fills up we could get stuck.
         args = ["pdftoppm", "-r", "200", pdf_path]
```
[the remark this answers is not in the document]
austintlee: Well, Jon or Mehul would need to weigh in, but I think we want users to select individual pages...

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -171,10 +177,13 @@ def read_stdout(fh, q):
             if len(data) < need_bytes:
                 continue
 
-            out_path = file_dir / f"image.{image_num}.ppm"
-            with open(out_path, "wb") as file:
-                file.write(data[0:need_bytes])
-            q.put(out_path)
+            if pages is None or (pages and (image_num + 1) in pages):
+                if pages:
```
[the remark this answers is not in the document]
austintlee: Ok, no logging then.

austintlee:
I'll revisit this at a later time.

PR #670, opened 2024-08-12
Add convert_file_to_pdf helper using libreoffice

baitsguy:
Tested with docx, but the library can support more. Haven't done perf testing, will be done next

*Note: Our docker installation currently doesn't install libreoffice, I will update it + docs we end up needing it*


.github/workflows/testing.yml
```diff
@@ -65,6 +65,8 @@ jobs:
         run: sudo apt-get update
       - name: Install poppler and tesseract
         run: sudo apt-get install -y poppler-utils tesseract-ocr
+      - name: Install libreoffice
```
eric-anderson, line 68: install it on the previous line. Change the name to "install apt dependencies"

lib/sycamore/sycamore/tests/unit/utils/test_fileformat_tools.py
```diff
@@ -0,0 +1,27 @@
...
+from io import BytesIO
+
+from pypdf import PdfReader
+
+import sycamore
+from sycamore.utils.fileformat_tools import convert_file_to_pdf
+from sycamore.tests.config import TEST_DIR
+
+
+def test_pdf_to_pdf():
```
eric-anderson, line 13: test_file_to_pdf?

lib/sycamore/sycamore/utils/fileformat_tools.py
```diff
@@ -0,0 +1,44 @@
...
+                "pdf",
+                source_path,
+                "--outdir",
+                output_path,
+            ]
+        )
+
+    assert doc.binary_representation is not None
+
+    current_filetype = doc.properties.get("filetype")
```
eric-anderson, line 29: .get("filetype", "unknown")

and remove the assertion. Unless libreoffice is using the extention for file type info

lib/sycamore/sycamore/utils/fileformat_tools.py
```diff
@@ -0,0 +1,44 @@
...
+    assert current_filetype is not None, "Document requires properties.filetype"
+
+    with NamedTemporaryFile(suffix=f".{current_filetype.split('/')[-1]}") as temp_file:
+        temp_file.write(doc.binary_representation)
+        temp_file.flush()
+
```
eric-anderson, line 35:
```
temp_path = Path(temp_file)
run_libreoffice(temp_file.name, temp_path.parent)
with open(temp_path.parent / temp_path.stem + ".pdf", "rb") as processed_file:
    doc.binary_representation = processed_file.read()
```

lib/sycamore/sycamore/utils/fileformat_tools.py
```diff
@@ -0,0 +1,44 @@
+import subprocess
+from tempfile import NamedTemporaryFile
+
+from sycamore.data import Document
+
+
+def convert_file_to_pdf(doc: Document) -> Document:
```
eric-anderson, line 7: binary_representation_to_pdf

I would expect convert_file_to_pdf(source_name, dest_name)

lib/sycamore/sycamore/utils/fileformat_tools.py
```diff
@@ -0,0 +1,44 @@
...
+        temp_file.write(doc.binary_representation)
+        temp_file.flush()
+
+        output_dir = temp_file.name.rsplit("/", 1)[0]
+        output_file_base = temp_file.name.rsplit(".", 1)[0]
+        run_libreoffice(temp_file.name, output_dir)
+
+        output_pdf_path = f"{output_file_base}.pdf"
+        with open(output_pdf_path, "rb") as processed_file:
+            doc.binary_representation = processed_file.read()
```
eric-anderson, line 42: doc.properties["filetype"] = "pdf"

lib/sycamore/sycamore/tests/unit/utils/test_fileformat_tools.py
baitsguy: oh yeah typo lol

lib/sycamore/sycamore/utils/fileformat_tools.py
baitsguy: it does from what I see so gonna leave this

lib/sycamore/sycamore/utils/fileformat_tools.py
baitsguy: the file extension when reading .docx was causing the Path lib to break (it includes multiple docs, e.g. 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'), so I needed to do it this way. 
I can choose to drop the file extensionbut figured its more useful when debugging

lib/sycamore/sycamore/utils/fileformat_tools.py
baitsguy: actually its a little unclear, their docs aren't very clear. Still useful for us to keep so we aren't supporting magic types

eric-anderson: ok. If it's unclear what libreoffice will do it's safer to do the clearly correct thing.

baitsguy: Which is enforcing/being aware of the extension I'd think?

eric-anderson: Per discussion, get the extension from the original filesystem.

lib/sycamore/sycamore/utils/fileformat_tools.py
eric-anderson: Per discussion this should have been application/pdf

eric-anderson (approved):

PR #677, opened 2024-08-14
Updated Entity Extractor + Infrastructure Changes

RitxmSaha:
This PR implements using pydantic models as entity schemas to be extracted from text.
**lib/sycamore/pyproject.toml**
- Added pydantic as a dependency

**lib/sycamore/sycamore/docset.py**
- Moved resolving duplicate entities to after all extractors run

**lib/sycamore/sycamore/llms/openai.py**
- Added functionality to use openai structured outputs
- Automatically select the beta client if the response format uses structured outputs

**lib/sycamore/sycamore/tests/unit/transforms/test_graph_extractor.py**
- Change tests to use normal docset inputs instead of hierarchical documents
- Changed entity extractor test to test using pydantic models

**lib/sycamore/sycamore/transforms/extract_graph.py**
- Removed resolve method and made it an importable function to be called after all graph extractors have been run
- Entity extractor now uses pydantic models to extract entity schemas
- Changes the way data schemas are modeled internally, instead of `nodes[label][value]` where value used to be the primary key, its now `nodes[label][hash]` where hash is the hash representation of the json object of a particular schema.

**lib/sycamore/sycamore/utils/pickle_pydantic.py**
- Implemented pickle workaround for pydantic objects created in ipynb notebooks to ensure they are ray serializable.
- https://github.com/pydantic/pydantic/issues/8232

### DIFF
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -549,13 +549,15 @@ class DocSet:
                     .explode()
                 )
         """
-        from sycamore.transforms.extract_graph import ExtractDocumentStructure
+        from sycamore.transforms.extract_graph import ResolveEntities, ExtractDocumentStructure
 
-        self.plan = ExtractDocumentStructure(self.plan)
         docset = self
+
+        docset.plan = ExtractDocumentStructure(docset.plan)
         for extractor in extractors:
             docset = extractor.extract(docset)
-
+        if len(extractors) > 0:
+            docset = ResolveEntities.resolve(docset=docset)
         return docset
 
     def extract_properties(self, property_extractor: PropertyExtractor, **kwargs) -> "DocSet":
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -554,10 +554,11 @@ class DocSet:
         docset = self
 
         docset.plan = ExtractDocumentStructure(docset.plan)
-        for extractor in extractors:
-            docset = extractor.extract(docset)
         if len(extractors) > 0:
+            for extractor in extractors:
+                docset = extractor.extract(docset)
             docset = ResolveEntities.resolve(docset=docset)
+
         return docset
 
     def extract_properties(self, property_extractor: PropertyExtractor, **kwargs) -> "DocSet":
--- a/lib/sycamore/sycamore/llms/openai.py
+++ b/lib/sycamore/sycamore/llms/openai.py
@@ -440,11 +440,15 @@ class OpenAI(LLM):
             completion = await self.client_wrapper.get_async_client().beta.chat.completions.parse(
                 model=self._model_name, **kwargs
             )
+            if completion.choices[0].message.content is None:
+                raise ValueError("OpenAI declined to respond to query")
+            return completion.choices[0].message.content
         else:
             completion = await self.client_wrapper.get_async_client().chat.completions.create(
                 model=self._model_name, **kwargs
             )
-        return completion.choices[0].message.content
+            return completion.choices[0].message.content
+        
 
     def _generate_using_guidance(self, prompt_kwargs) -> str:
         guidance_model = self.client_wrapper.get_guidance_model(self.model)
--- a/lib/sycamore/sycamore/llms/openai.py
+++ b/lib/sycamore/sycamore/llms/openai.py
@@ -440,17 +440,17 @@ class OpenAI(LLM):
 
     async def _generate_awaitable_using_openai(self, prompt_kwargs, llm_kwargs) -> Awaitable[str]:
         kwargs = self._get_generate_kwargs(prompt_kwargs, llm_kwargs)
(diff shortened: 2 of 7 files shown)


lib/sycamore/sycamore/docset.py
```diff
@@ -549,13 +549,15 @@ def extract_graph_structure(self, extractors: list[GraphExtractor], **kwargs) ->
...
+        from sycamore.transforms.extract_graph import ResolveEntities, ExtractDocumentStructure
 
-        self.plan = ExtractDocumentStructure(self.plan)
         docset = self
+
+        docset.plan = ExtractDocumentStructure(docset.plan)
         for extractor in extractors:
             docset = extractor.extract(docset)
-
+        if len(extractors) > 0:
```
baitsguy, line 559: nit but maybe just stick the for loop under this too so all extractors stuff is under

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -399,13 +409,18 @@ def generate(self, *, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) ->
         self._cache_set(key, value)
         return ret
 
-    def _generate_using_openai(self, prompt_kwargs, llm_kwargs) -> str:
+    def _generate_using_openai(self, prompt_kwargs, llm_kwargs) -> Any:
```
baitsguy, line 412: Can we have a separate method for the openai structured responses? We had a bug earlier because this function had an Any return type so it's safer to keep them separate

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -232,75 +149,190 @@ async def gather_api_calls():
...
+        from sycamore import Execution
+        from sycamore.reader import DocSetReader
+        from ray.data.aggregate import AggregateFn
+
+        reader = DocSetReader(docset.context)
+
+        # Get list[Document] representation of docset, trigger execute with take_all()
+        execution = Execution(docset.context, docset.plan)
+        dataset = execution.execute(docset.plan)
+        docs = dataset.take_all()
```
baitsguy, line 267: why not just use a map on the docset? or the dataset?

lib/sycamore/sycamore/transforms/extract_graph.py
```diff
@@ -232,75 +149,190 @@ async def gather_api_calls():
...
+                for hash in hashes:
+                    if nodes.get(key, {}).get(hash, {}) == {}:
+                        nodes.setdefault(key, {})
+                        nodes[key][hash] = extracted[key][hash]
+                    else:
+                        for rel_uuid, rel in extracted[key][hash]["relationships"].items():
+                            nodes[key][hash]["relationships"][rel_uuid] = rel
+            return nodes
+
+        def merge(nodes1, nodes2):
```
baitsguy, line 293: can you add documentation about how resolution is currently working?

lib/sycamore/sycamore/utils/pickle_pydantic.py
```diff
@@ -0,0 +1,40 @@
+import io
+from typing import Any, Dict, Optional, Type
+from ray import cloudpickle
+from pydantic import BaseModel
+
+
+def safe_cloudpickle(obj: Any) -> bytes:
```
baitsguy, line 7: tests

baitsguy (commented):
Seems like there's a bunch going on with hierarchical changes and openai interfaces. Seems functional but a change like this could benefit from smaller PRs

lib/sycamore/sycamore/transforms/extract_graph.py
RitxmSaha: Could do a map on the dataset, question is how do I add a document to a dataset? that's how I add the documents as nodes.

lib/sycamore/sycamore/docset.py
RitxmSaha: fixed

lib/sycamore/sycamore/llms/openai.py
RitxmSaha: fixed

lib/sycamore/sycamore/transforms/extract_graph.py
RitxmSaha: added documentation in ResolveEntities. Note, we will probably move this code to a seperate transform later on.

lib/sycamore/sycamore/transforms/extract_graph.py
RitxmSaha: Might be worth to change this in another PR when we make seperate transforms for entity/relationship resolution.

lib/sycamore/sycamore/utils/pickle_pydantic.py
RitxmSaha: added tests

baitsguy (approved):

PR #680, opened 2024-08-15
Add AutoMaterialize

eric-anderson:
automatically adds materialize nodes after each of the nodes in the plan.


lib/sycamore/sycamore/plan_nodes.py
```diff
@@ -3,20 +3,37 @@
...
     from ray import Dataset
+    from sycamore.context import Context
 
 
 class NodeTraverse:
     def __init__(
-        self, before: Optional[Callable[["Node"], "Node"]] = None, after: Optional[Callable[["Node"], "Node"]] = None
+        self,
+        before: Optional[Callable[["Node"], "Node"]] = None,
+        visit: Optional[Callable[["Node"], None]] = None,
```
baitsguy, line 13: why does this not return the node either?

lib/sycamore/sycamore/lineage.py
```diff
@@ -106,3 +111,135 @@ def doc_to_name(doc: Document) -> str:
...
+        self.directory = path.pop("root", None)
+        self._basename_to_count: dict[str, int] = {}
+
+    def once(self, context, node):
+        self._name_unique = set()
+        node = node.traverse(after=self._naming_pass())
+
+        self._setup_directory()
+
+        node = node.traverse(visit=self._cleanup_pass())
```
baitsguy, line 156: what's the reason to separate naming and cleanup?

lib/sycamore/sycamore/plan_nodes.py
baitsguy: I don't think I understand the poin tof visit? Can you add a couple comments?

eric-anderson: Simpler code.

lib/sycamore/sycamore/lineage.py
eric-anderson: Easier to understand code. I could do pass1 and pass2, but then you have to read the code to figure out what it is doing. naming_and_cleanup_pass would explain, but would mean the function isn't single purpose.

Since the performance loss is irrelevant, I prioritized clearer code.

baitsguy (approved):

PR #681, opened 2024-08-15
Change the file writer to create the output directory.

bsowell:
Based on customer feedback. This changes the default and fixes the implementation of the makedirs parameter so that the writer will create the directory to write to by default.


HenryL27 (approved):
lgtm

karanataryn (approved):

PR #686, opened 2024-08-15
Change default partitioning threshold to 0.35 from 0.4

MarkLindblad:


MarkLindblad:
Did not change the default threshold for `SycamorePartitioner` for backwards compatibility.

PR #687, opened 2024-08-15
import sycamore does not import ray

eric-anderson:
import ray is about 0.2s, import sycamore about 3s, so ~10% speedup.
First step of many, but this one was pretty mechanical.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -70,6 +70,9 @@ jobs:
       - name: Run tests
         run: poetry run pytest sycamore/tests/unit/
         working-directory: lib/sycamore
+      - name: Run more tests
+        run: poetry run python sycamore/tests/manual/test_fast_sycamore_import.py
+        working-directory: lib/sycamore
       - name: DF-6
         run: df
 
--- a/lib/sycamore/sycamore/tests/manual/test_fast_sycamore_import.py
+++ b/lib/sycamore/sycamore/tests/manual/test_fast_sycamore_import.py
@@ -1,0 +1,32 @@
+"""WARNING: this test needs to be run separately from other tests because if they
+   import sycamore and run it, then the checks for no direct depenencies will be
+   incorrect. Similarly, this test has to be run directly as python. pytest will
+   pre-import sycamore."""
+
+import inspect
+import sys
+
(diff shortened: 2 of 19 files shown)


lib/sycamore/sycamore/tests/manual/test_fast_sycamore_import.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_00_sycamore_not_yet_imported():
+    assert "sycamore" not in sys.modules, "Test run with other tests that import ray"
+
+
+def test_no_direct_ray():
+    assert "ray" not in sys.modules, "Test run with other tests that import ray"
+
+    assert "ray" not in sys.modules
```
HenryL27, line 17: why twice?

.github/workflows/testing.yml
```diff
@@ -70,6 +70,9 @@ jobs:
       - name: Run tests
         run: poetry run pytest sycamore/tests/unit/
         working-directory: lib/sycamore
+      - name: Run more tests
+        run: poetry run python sycamore/tests/manual/test_fast_sycamore_import.py
+        working-directory: lib/sycamore
```
HenryL27, line 75: heh heh this is gonna annoy some people... Maybe the error message in the tests should explain that you're not allowed to import ray in such a way that it gets imported with sycamore?

lib/sycamore/sycamore/tests/manual/test_fast_sycamore_import.py
eric-anderson: Because the line in between that did import sycamore got lost somehow. The test was uninteresting.

.github/workflows/testing.yml
eric-anderson: Updated assertion error.  I didn't want to have to keep fixing the problem.

lib/sycamore/sycamore/tests/manual/test_fast_sycamore_import.py
eric-anderson: The automatic linting stuff we had removed it.

HenryL27 (approved):

PR #690, opened 2024-08-16
Fix Override Text Bug

karanataryn:


HenryL27 (approved):

bsowell (approved):
I'm of two minds about this. I think it will probably solve the immediate problem, but I don't love bypassing the property accessor as it kind of defeats the purpose of having a single way to access this. I think the _override thing is also going to get us into trouble again in the future. 

That said, this is probably better than nothing in the immediate term, so I'm okay with getting it in with the expectation that we will need to revisit at some point.

PR #696, opened 2024-08-19
Fixed an error in file_writer_ray.py

afriedman412:
_FileDataSink would crash if you passed it an empty string as `_root` because PyArrow can't create a directory at ""

Now if `_root` is an empty string, it automatically switches to "./" before creating the directory.

### DIFF
--- a/lib/sycamore/sycamore/connectors/file/file_writer_ray.py
+++ b/lib/sycamore/sycamore/connectors/file/file_writer_ray.py
@@ -39,6 +39,8 @@ class _FileDataSink(Datasink):
         is_s3_uri = parsed_uri.scheme == "s3"
 
         if not is_s3_uri and self._filesystem.get_file_info(self._root).type is FileType.NotFound:
+            if self._root == "":
+                self._root = "./"
             self._filesystem.create_dir(self._root, recursive=True)
 
     def write(self, blocks: Iterable[Block], ctx: TaskContext) -> Any:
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/connectors/file/file_writer_ray.py
```diff
@@ -39,6 +39,8 @@ def on_write_start(self) -> None:
         is_s3_uri = parsed_uri.scheme == "s3"
 
         if not is_s3_uri and self._filesystem.get_file_info(self._root).type is FileType.NotFound:
+            if self._root == "":
+                self._root = "./"
```
HenryL27, line 43: Can we put this logic in the __init__ method? I'd just feel more comfortable keeping state as immutable as possible.

afriedman412: works for me!

lib/sycamore/sycamore/connectors/file/file_writer_ray.py
```diff
@@ -25,6 +25,8 @@ def __init__(
     ):
         (paths, self._filesystem) = _resolve_paths_and_filesystem(path, filesystem)
         self._root = paths[0]
+        if self._root == "":
```
HenryL27, line 28: so sorry, but I think you need to check for s3 uris here too, right?

afriedman412: Assuming we are using `urlparse(url).scheme == "s3"` to decide if its an s3 uri or not it shouldn't matter because both "" and "./" return False

Unless I'm missing something, this feels redundant!

HenryL27: fair enough. ok!

HenryL27 (approved):

PR #697, opened 2024-08-19
Fill in "gaps" in non-contiguous rows and columns from TATR .

bsowell:
We have seen at least one case where table transformers will produce a cell that spans a non-contiguous span of rows (e.g. [8,9,11,12]). This is clearly wrong, and we were asserting on it, but it is quite difficult to update the table transformer heuristics in a way that prevents this without having other potentially deleterious effects on quality. This change simply replaces the row and column spans with list(range(min, max)), which always produces a continuous run of rows/cols.


HenryL27 (approved):
Thanks

PR #700, opened 2024-08-20
implemented resolve graph entities

RitxmSaha:


lib/sycamore/sycamore/transforms/resolve_graph_entities.py
```diff
@@ -0,0 +1,133 @@
...
+class EntityResolver(ABC):
+    def __init__(self):
+        pass
+
+    @abstractmethod
+    def resolve(self, entities: Any) -> Any:
+        pass
+
+
+class ResolveEntities:
```
dtecuci, line 23: Could you add some high level comments to describe what these classes.methods do?

dtecuci (approved):

lib/sycamore/sycamore/transforms/resolve_graph_entities.py
RitxmSaha: comments have been added in the next PR

PR #702, opened 2024-08-20
SimplePrompt class

baitsguy:
Separate SimplePrompt and GuidanceExecutor, allows you to write a prompt and have your own execution strategy (e.g. on a native client)


lib/sycamore/sycamore/llms/guidance.py
```diff
@@ -0,0 +1,37 @@
...
+    with system():
+        lm = model + prompt.system.format(**kwargs)
+
+    with user():
+        lm += prompt.user.format(**kwargs)
+
+    with assistant():
+        lm += gen(name=prompt.var_name)
+
+    return lm[prompt.var_name]
```
bsowell, line 16: At this point `var_name` is entirely superfluous, right? I think it's a vestige of the time when we returned the dict, but now since we return the string, it's entirely wrapped in the the internal LLM call and never exposed? I don't think you need to remove it in this PR, but we probably could, right?

lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -1,64 +1,34 @@
...
-class GuidancePrompt(ABC):
-    @abstractmethod
-    def execute(self, model: GuidanceModel, **kwargs) -> str:
-        pass
+class SimplePrompt(ABC):
+    system: str = None
+    user: str = None
+    var_name: str = "answer"
+
+    def as_messages(self) -> list[dict]:
```
bsowell, line 13: Do we need to format here? I guess these messages would still have placeholders (like {filter_message}), right?

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -462,6 +464,6 @@ async def _generate_awaitable_using_openai_structured(self, prompt_kwargs, llm_k
 
     def _generate_using_guidance(self, prompt_kwargs) -> str:
```
bsowell, line 465: Would the next step be to convert `generate_with_openai` to also accept a SimplePrompt and execute that? Then eventually we could deprecate `generate_with_guidance`, if indeed we intend to do that?

bsowell (approved):
I think this looks reasonable. Left a few questions, but nothing that seems blocking. 

I hope our integ tests are sufficient to catch issues here, but I wouldn't bet money on it. Maybe run a couple of our notebooks to make sure that entity extraction/etc works as expected?

lib/sycamore/sycamore/llms/guidance.py
baitsguy: I think that makes sense? Tbh I don't remember how we got here

lib/sycamore/sycamore/llms/prompts/default_prompts.py
baitsguy: This is only being used by the new prompts (the ones that format in the constructor). For guidance (which formats in execute()), this code isn't used. Maybe I'll leave a note

baitsguy: But the 2 patterns are:

```
prompt = something # basically guidance prompts, has a .execute() implemented
prompt.execute(params=values_for_placeholders)
```

The other is:

```
prompt = Something(params=values_for_placeholders)
executor.execute(prompt)
executor.execute_w_messages_api(prompt.as_messages)
```

lib/sycamore/sycamore/llms/openai.py
baitsguy: I think so yeah, we should likely add a flag to `generate` that's more explicit about what you're executing with

PR #704, opened 2024-08-21
Return docset info

afriedman412:
Added a `.get_doc_info` function to the `DocSet` class that returns the document information (especially names of source documents) instead of pretty printing it.


HenryL27 (commented):
I'm not sure get_doc_info should be its own method. Can I recommend that if you want this behavior you do something like
```python
def truncate_elts(doc):
    doc.data['elements'] = f"<{len(doc.elements)} elements>"
    return doc

docset.map(truncate_elts).take(n)
```

eric-anderson:

PR #712, opened 2024-08-22
Add Local Mode Support for Composite Transform and Binary Read

karanataryn:
Adds the ability to be able to run most notebooks using `ExecMode.LOCAL`

### DIFF
--- a/lib/sycamore/sycamore/connectors/file/file_scan.py
+++ b/lib/sycamore/sycamore/connectors/file/file_scan.py
@@ -6,7 +6,6 @@ from typing import Any, Optional, Union, Tuple, Callable, TYPE_CHECKING
 import uuid
 import logging
 
-# from pyarrow.fs import FileSystem, LocalFileSystem
 from pyarrow.fs import FileSystem, LocalFileSystem, FileSelector
 from sycamore.data import Document
 from sycamore.plan_nodes import Scan
@@ -167,7 +166,6 @@ class BinaryScan(FileScan):
         return files.map(self._to_document, **self.resource_args)
 
     def local_source(self, **kwargs) -> list[Document]:
-        import os
 
         if isinstance(self._paths, str):
             paths = [self._paths]
--- a/lib/sycamore/sycamore/connectors/file/file_scan.py
+++ b/lib/sycamore/sycamore/connectors/file/file_scan.py
@@ -198,8 +198,12 @@ class BinaryScan(FileScan):
             documents.append(document)
 
         for path in paths:
-            for info in self._filesystem.get_file_info(FileSelector(path), recursive=True):
-                process_file(info)
+            path_info = self._filesystem.get_file_info(path)
+            if path_info.is_file:
+                process_file(path_info)
+            else:
+                for info in self._filesystem.get_file_info(FileSelector(path, recursive=True)):
+                    process_file(info)
         return documents
 
     def format(self):
--- a/lib/sycamore/sycamore/connectors/file/file_scan.py
+++ b/lib/sycamore/sycamore/connectors/file/file_scan.py
@@ -174,29 +174,32 @@ class BinaryScan(FileScan):
         documents = []
 
         def process_path(path):
-            if not self._filesystem.get_file_info(path).is_file:
+            if (
+                self._filesystem.get_file_info(path).is_file
+                and not self._filter_paths_by_extension
+                or path.endswith(self.format())
+            ):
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/connectors/file/file_scan.py
```diff
@@ -166,6 +165,47 @@ def execute(self, **kwargs) -> "Dataset":
...
+                if self._metadata_provider:
+                    document.properties.update(self._metadata_provider.get_metadata(path))
+
+                documents.append(document)
+            elif not self._filesystem.get_file_info(path).is_file:
+                file_selector = FileSelector(path)
+                for item in self._filesystem.get_file_info(file_selector):
+                    process_path(item.path)
+
+        for path in paths:
```
eric-anderson, line 204: You can simplify the code here.

```
for path in paths:
   for f in self._filesystem.get_file_info(FileSelector(path), recursive=True):
        process_file(f)

def process_file(info):
    if not info.is_file:
        return
    if self._filter_paths_by_extension and not info.path.endswith(self.format()):
        return
   with self._filesystem.open_input_file(path) as file:
   ...
```

lib/sycamore/sycamore/docset.py
```diff
@@ -835,7 +835,7 @@ def llm_filter(
         llm: LLM,
         new_field: str,
         prompt: Union[list[dict], str],
-        field: Optional[str] = "text_representation",
```
eric-anderson, line 838: This feels like an unrelated change. It's generally best to have those in separate PRs, but if necessary make sure it's included in the description.

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -297,5 +298,15 @@ def _local_process(self, in_docs: list[Document]) -> list[Document]:
 
         return docs
 
+    def local_execute(self, all_docs: list[Document]) -> list[Document]:
+        docs = [d for d in all_docs if not isinstance(d, MetadataDocument)]
+        metadata = [d for d in all_docs if isinstance(d, MetadataDocument)]
+        outputs = self._local_process(docs)
+        to_docs = [d for d in outputs if not isinstance(d, MetadataDocument)]
+        if self._enable_auto_metadata and (len(docs) > 0 or len(to_docs) > 0):
+            outputs.extend(BaseMapTransform._update_lineage(docs, to_docs))
```
eric-anderson, line 307: This works, but is a bit weird to call the private method on an unrelated class. Please move _update_lineage out of BaseMapTransform.

lib/sycamore/sycamore/connectors/file/file_scan.py
karanataryn: Included, thanks

lib/sycamore/sycamore/docset.py
karanataryn: Moved it to a separate PR: https://github.com/aryn-ai/sycamore/pull/720

lib/sycamore/sycamore/connectors/file/file_scan.py
```diff
@@ -173,37 +173,37 @@ def local_source(self, **kwargs) -> list[Document]:
...
+                return
+
+            with self._filesystem.open_input_file(info.path) as file:
+                binary_data = file.read()
+
+            document = Document()
+            document.doc_id = str(uuid.uuid1())
+            document.type = self._binary_format
+            document.binary_representation = binary_data
+            document.properties.update({"path": info.path})
```
eric-anderson, line 189: nit: you can shorten this 
`document.properties["path"] = info.path`

lib/sycamore/sycamore/docset.py
eric-anderson: It's still showing up in my diff, but maybe it's github diff weirdness.

lib/sycamore/sycamore/transforms/base.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 
 from sycamore.data import Document, MetadataDocument
+from sycamore.utils.lineage_utils import _update_lineage
```
eric-anderson, line 7: nit: import update_lineage
(it's a public function of lineage_utils)

eric-anderson (approved):

lib/sycamore/sycamore/docset.py
karanataryn: Yeah that must be it, the PR above is merged.

PR #721, opened 2024-08-24
Add Echo Guidance Flag to suppress IPython HTML Outputs 

karanataryn:
Allows us the flexibility to ensure the issue [here](https://github.com/guidance-ai/guidance/issues/506) doesn't show up, which comes up when running notebooks locally. By default set to `False`.


lib/sycamore/sycamore/llms/openai.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         # Deprecated names that we support for backwards compatibility.
         api_type: Optional[str] = None,
         api_base: Optional[str] = None,
+        echo: bool = False,
```
HenryL27, line 78: maybe put this in the non-deprecated section

HenryL27 (approved):
holy cow you actually found it! 1 tweak

PR #728, opened 2024-08-27
Dependencies 1/n: Remove need to restart colab runtime

bsowell:
This PR serves two purposes:

1) Remove/upgrade/downgrade the dependencies that require restarting the Colab runtime when installing Sycamore. After applying this change, you can `%pip install sycamore-ai` and then `import sycamore` successfully in the next cell.

Note this does have some side-effects. Most notably, I had to downgrade the PIL required version, as Colab has an old version (9.4.0). The primary impact is that this gives us less control over the font sizing we use for labeling bounding boxes. I did upgrade the code so that it will pass the size parameter when a newer version of PIL is present.

2) Beyond this specific fix, I intended this change to serve as a model for how we can decompose some of our dependencies into extras going forward. In this case I moved three dependencies that we only use for evaluation: apted, datasets, and rouge, into an 'eval' extra that is only installed when requested via `pip install sycamore-ai[eval]`. Since the eval work is mostly internal, I think this is a relatively safe place to start. I've included a few patterns to make this easier including a decorator (`requires_modules`) for declaring runtime dependencies, and the convention of conditionally importing classes only required for type-checking using the typing.TYPE_CHECKING flag.

These are similar to patterns used by other libraries, like Unstructured.


lib/sycamore/sycamore/data/table.py
```diff
@@ -421,7 +422,7 @@ def draw(self, target: U) -> U:
         return try_draw_boxes(target, self.cells, color_fn=lambda _: "red", text_fn=lambda _, i: None)
 
 
-class TableTree(apted.helpers.Tree):
+class TableTree:
```
HenryL27, line 425: the inheritance is important here I thought

bsowell: I did try it and it still seems to work. I think duck typing is sufficient here, though if you can find a counter-example I am open to it.

eric-anderson (approved):

lib/sycamore/pyproject.toml
```diff
@@ -19,42 +19,44 @@ pandas = "2.1.1"
...
+urllib3 = ">=1.26.0,<3"
 async-timeout = ">4.0.0"
-pillow = "10.2.0" # See poetry-lock/pyproject.toml
-setuptools = "69.5.1" # See poetry-lock/pyproject.toml
+pillow = ">=9.4.0" # See poetry-lock/pyproject.toml
 tenacity = "^8.4.1"
 diskcache = "^5.6.3"
 overrides = "^7.7.0"
 structlog = "^24.2.0"
+setuptools = ">70.0.0"
```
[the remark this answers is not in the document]
bsowell: Unfortunately I don't remember exactly where this failed. I ran all the unit tests successfully. I will make sure to get the integ tests running as well, at least locally.

PR #733, opened 2024-08-27
Don't require installing neo4j unless it's used.

eric-anderson:


bsowell (commented):
FYI: I have a change in flight that factors all the connectors into extras, including neo4j. Should be out in PR tonight/early tomorrow. It's fine if you want to put this in first.

RitxmSaha (approved):
Looks good to me!

bsowell (approved):

PR #734, opened 2024-08-28
Fix materialize + S3 not working.

eric-anderson:
Existing code was relying on python pathlib to do most operations rather than running them through the pyarrow filesystem.  Fix that, and add a test with a fake pyarrow filesystem to make sure we run with a path that does not exist in the local filesystem.

### DIFF
--- a/lib/sycamore/sycamore/materialize.py
+++ b/lib/sycamore/sycamore/materialize.py
@@ -17,6 +17,33 @@ if TYPE_CHECKING:
 logger = logging.getLogger("__name__")
 
 
+class _PyArrowFsHelper:
+    def __init__(self, fs: "pyarrow.FileSystem"):
+        self._fs = fs
+
+    def list_files(self, path):
+        from pyarrow.fs import FileSelector
+
+        return self._fs.get_file_info(FileSelector(str(path), allow_not_found=True, recursive=True))
+
+    def file_exists(self, path: Path) -> bool:
+        from pyarrow.fs import FileType
+
+        info = self._fs.get_file_info(str(path))
+        return info.type == FileType.File
+
+    def safe_cleanup(self, path) -> None:
+        # materialize dirs should be non-hierarchical, minimize the chance that
+        # mis-use will delete unexpected files.
+        plen = len(str(path)) + 1
+        for fi in self.list_files(path):
+            assert "/" not in fi.path[plen:], f"Refusing to clean {path}. Found unexpected hierarchical file {fi.path}"
+
+        logging.info(f"Cleaning up any materialized files in {path}")
+        self._fs.delete_dir_contents(str(path), missing_dir_ok=True)
+        self._fs.create_dir(str(path))
+
+
 class Materialize(UnaryNode):
     def __init__(
         self,
@@ -50,6 +77,8 @@ class Materialize(UnaryNode):
         else:
             assert False, f"unsupported type ({type(path)}) for path argument, expected str, Path, or dict"
 
+        self._fshelper = _PyArrowFsHelper(self._fs)
+
         if source_mode != MaterializeSourceMode.OFF:
             assert path is not None
             assert (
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/tests/unit/inmempyarrowfs.py
```diff
@@ -0,0 +1,98 @@
...
+            if not k.startswith(dir):
+                continue
+            if "/" in k[dlen:]:
+                continue
+            ret.append(FileInfo(str(k), type=FileType.File))
+
+        return ret
+
+    def move(self, src, dest):
+        assert False, "unimplemented"
```
HenryL27, line 65: there's also `raise NotImplementedError()`. Not sure what you prefer.

lib/sycamore/sycamore/tests/unit/test_materialize.py
```diff
@@ -121,6 +124,7 @@ def doc_to_name3(doc):
             self.check_files(tmpdir, ext=".test3")
 
             files = glob.glob(tmpdir + "/*")
+            logging.error(f"ERIC {files}")
```
HenryL27, line 127: cleaning

lib/sycamore/sycamore/tests/unit/test_materialize.py
```diff
@@ -250,24 +253,26 @@ def doc_to_name4(doc):
...
             a._path["clean"] = True
             ds.execute()
             files = [f for f in Path(tmpdir).rglob("*")]
-            logging.info(f"TO Found-3 {files}")
             assert len([f for f in files if ".test4" in str(f)]) == 3 + 1 + 3 + 2
 
 
+def any_id(d):
+    if isinstance(d, MetadataDocument):
+        return str(d.metadata)
```
HenryL27, line 267: this seems weird? but ok

lib/sycamore/sycamore/tests/unit/test_materialize.py
```diff
@@ -331,3 +336,24 @@ def test_materialize_read(self):
             Path(tmpdir).rmdir()
             with pytest.raises(ValueError):
                 ds.take_all()
+
+
+class TestAllViaPyarrowFS(unittest.TestCase):
```
HenryL27, line 341: Is it worth adding a ray test in here?

HenryL27 (approved):

lib/sycamore/sycamore/tests/unit/inmempyarrowfs.py
eric-anderson: switched. I have no strong preference.

lib/sycamore/sycamore/tests/unit/test_materialize.py
eric-anderson: removed.

lib/sycamore/sycamore/tests/unit/test_materialize.py
eric-anderson: ack

lib/sycamore/sycamore/tests/unit/test_materialize.py
eric-anderson: Per discussion, not needed. There is an integration test that uses ray.

lib/sycamore/sycamore/materialize.py
```diff
@@ -328,24 +338,29 @@ def before(node):
...
-        if self.directory is None:
+        directory = path.pop("root", None)
+        self._path = path
+
+        if directory is None:
             from datetime import datetime
             import tempfile
 
             now = datetime.now().replace(microsecond=0)
             dir = Path(tempfile.gettempdir()) / f"materialize.{now.isoformat()}"
```
HenryL27, line 360: does this ever leak storage?

HenryL27 (approved):
I dont see anything obviously wrong here and I'll trust you that this fixes the automat

lib/sycamore/sycamore/materialize.py
eric-anderson: It's intended to "leak". You're asking it to materialize so it's trying to leave it around.

PR #735, opened 2024-08-28
Handle None cases for element.table = <>

baitsguy:
Triggered by https://github.com/aryn-ai/sycamore/blob/eb27e29f4938c383288066fada8f8ffe1952c277/lib/sycamore/sycamore/transforms/table_structure/extract.py#L168

### DIFF
--- a/lib/sycamore/sycamore/data/element.py
+++ b/lib/sycamore/sycamore/data/element.py
@@ -174,9 +174,10 @@ class TableElement(Element):
         return self.data["table"]
 
     @table.setter
-    def table(self, value: Table) -> None:
+    def table(self, value: Optional[Table]) -> None:
         self.data["table"] = value
-        self.data["text_representation"] = self.data["table"].to_csv()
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/data/element.py
```diff
@@ -174,9 +174,10 @@ def table(self) -> Optional[Table]:
         return self.data["table"]
 
     @table.setter
-    def table(self, value: Table) -> None:
+    def table(self, value: Optional[Table]) -> None:
         self.data["table"] = value
-        self.data["text_representation"] = self.data["table"].to_csv()
+        if value:
```
HenryL27, line 179: might write out the `if value is not None` just in case we add a bool coercion to tables. Not sure such a thing ever makes sense so feel free to ignore me.

HenryL27 (approved):
Don't know of other places where this is an issue. I figure let's fix 'em as they come

PR #736, opened 2024-08-28
Fixed neo4j relationship property loading + added support for loading lists and dictionaries as properties

RitxmSaha:
**neo4j_writer.py**
- Fixed bug where relationship properties were not being added into neo4j(issue with prompt + logic in `_parse_docs`)
- Added support for lists + dictionaries are neo4j properties. Previously, they were flattened but this turns out to not be useful to the end user whilst just doing a `json.dumps()` on them allows them to be used by the user in neo4j with some commands.

**document.py**
- Now checks for text_representation in `self.data.get('text_representation')` before setting it to `self.data.get('table').to_csv()`

### DIFF
--- a/lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
+++ b/lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
@@ -274,7 +274,10 @@ class Neo4jWriteCSV(MapBatch, Write):
                 }
                 for property_key, property_value in value["properties"].items():
                     if property_key not in include_relationships:
-                        rel[property_key] = property_value
+                        if isinstance(property_value, list) or isinstance(property_value, dict):
+                            node[property_key] = json.dumps(property_value)
+                        else:
+                            node[property_key] = property_value
                 relationships[value["START_LABEL"]][value["END_LABEL"]].append(rel)
         return nodes, relationships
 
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -171,9 +176,8 @@ def accumulate_row(headers, row):
                 if key in data:
                     headers["nodes"][node_key][key] = True
             #### add all keys from properties ####
-            for key in set(pair[0] for pair in flatten_data(data["properties"])):
+            for key in data["properties"].keys():
```
HenryL27, line 179: wait what happens to nested objects now?

lib/sycamore/sycamore/data/document.py
```diff
@@ -337,8 +337,9 @@ def __init__(self, document=None, **kwargs):
 
         self.doc_id = self.data.get("doc_id", str(uuid.uuid4()))
         self.children = self.data.get("children", [])
-        if self.data.get("type", "") == "table" and self.data.get("table", None) is not None:
-            self.text_representation = self.data["table"].to_csv()
+        if self.data.get("type", None) == "table":
+            table_csv = self.data.get("table").to_csv() if self.data.get("table") else ""
+            self.text_representation = self.data.get("text_representation", table_csv)
```
HenryL27, line 342: this _should_ be redundant

HenryL27 (commented):
Is json-dumping actually what you want to do? How was it failing?

PR #738, opened 2024-08-29
Fix calls to structured outputs so that they can be cached

RitxmSaha:
Uses openai's parser to parse the pydantic object into a json schema. This now allows for the response format to be serialized, meaning llm cache now works(previously, it was broken).

### DIFF
--- a/lib/sycamore/sycamore/transforms/extract_graph_entities.py
+++ b/lib/sycamore/sycamore/transforms/extract_graph_entities.py
@@ -6,6 +6,7 @@ from sycamore.plan_nodes import Node
 from sycamore.transforms.map import Map
 from sycamore.data import HierarchicalDocument
 from sycamore.llms import LLM
+from openai.lib._parsing import type_to_response_format_param
 from pydantic import BaseModel, create_model
 
 import json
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/transforms/extract_graph_entities.py
```diff
@@ -103,7 +104,7 @@ def _deserialize_entities(self):
         return create_model("entities", __base__=BaseModel, **fields)
 
     async def _extract_from_section(self, summary: str) -> str:
-        llm_kwargs = {"response_format": self._deserialize_entities()}
+        llm_kwargs = {"response_format": type_to_response_format_param(self._deserialize_entities())}
```
baitsguy, line 107: I'm a little hesistent to use an openai function directly in the transform

Since this is openai specific formatting, is it better to let the openai.py file handle this?

RitxmSaha: I agree, added the change + an additional try except loop for openai structured outputs. Sometimes the response gets cancelled because the LLM never reaches a stop token and repeats the same thing over and over again and in this scenario, I think the best failover is to return an empty dict back. It's bound to happen over large scale data and missing a document isn't the biggest deal.

baitsguy (approved):

PR #740, opened 2024-08-29
Dependencies 2/3: Move connectors to extras.

bsowell:
This commit creates extras for the Sycamore connectors, including duckdb, elasticsearch, opensearch, neo4j, pinecone, and weaviate. The bulk of this commit consists of code changes to avoid importing these extra dependencies at the top-level of modules and provide the proper runtime warnings.


HenryL27 (approved):
lgtm. 1 thing - the pinecone client lib has an extra of its own for grpc, which we require. In this refactored world I could easily see someone installing sycamore with no extras, trying to write to pinecone, finding out they need to install pinecone, doing that directly, and then getting annoyed when they find out they actually need pinecone[grpc]. Do we care? I admit I don't see an easy solution.

lib/sycamore/sycamore/connectors/pinecone/pinecone_writer.py
```diff
@@ -42,7 +45,10 @@ class PineconeWriterClientParams(BaseDBWriter.ClientParams):
 
 
 class PineconeWriterClient(BaseDBWriter.Client):
+    @requires_modules("pinecone-client", extra="pinecone")
```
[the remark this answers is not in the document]
bsowell: Good catch. pinecone-client is the name of the pypi project, but not the module. Will update.

baitsguy (approved):
Query looks good

bsowell:
> lgtm. 1 thing - the pinecone client lib has an extra of its own for grpc, which we require. In this refactored world I could easily see someone installing sycamore with no extras, trying to write to pinecone, finding out they need to install pinecone, doing that directly, and then getting annoyed when they find out they actually need pinecone[grpc]. Do we care? I admit I don't see an easy solution.

As you said, it's not clear what we can do here. The error message we provide says to install the sycamore extra, which will include grpc. If the user insists on installing pinecone themselves, then caveat emptor, I guess. I'm not too worried about it unless people start stumbling on it.

karanataryn (approved):
Lgtm

RitxmSaha (approved):
lgtm!

lib/sycamore/sycamore/reader.py
```diff
@@ -207,6 +208,7 @@ def pandas(self, dfs: Union[DataFrame, list[DataFrame]]) -> DocSet:
         scan = PandasScan(dfs)
         return DocSet(self._context, scan)
 
+    @requires_modules("opensearchpy", extra="opensearch")
```
eric-anderson, line 211: Do we need any of the requires other than here?  I think this is the entry point so it's sufficient to have it here.

lib/sycamore/sycamore/tests/manual/test_fast_sycamore_import.py
```diff
@@ -17,21 +17,21 @@ def test_disallow_unconditional_dependencies():
     import sycamore
 
     _ = sycamore  # make sure tools don't remove the import
-    for m in ["guidance", "neo4j", "ray", "torch"]:
+    for m in ["guidance", "neo4j", "ray", "torch", "duckdb", "elasticsearch", "opensearchpy", "pinecone", "weaviate"]:
```
eric-anderson, line 20: can you switch this to a constant so we don't end up repeating it three times?  I probably should have done that before.

lib/sycamore/sycamore/writer.py
```diff
@@ -266,17 +268,17 @@ def weaviate(
             WeaviateClientParams,
             WeaviateWriterTargetParams,
         )
-        from sycamore.connectors.weaviate.weaviate_writer import CollectionConfigCreate
+        from weaviate.collections.classes.config import _CollectionConfigCreate
```
eric-anderson, line 271: `# Importing _ prefixed stuff is fairly common for users of the weaviate codebase`

eric-anderson (approved):

lib/sycamore/sycamore/tests/manual/test_fast_sycamore_import.py
bsowell: Oh, yeah. Definitely. I thought about this, and honestly thought I had done it, but now I can't find any evidence of it. Guess I'm hallucinating now too.

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -61,16 +63,22 @@ def compatible_with(self, other: BaseDBWriter.TargetParams) -> bool:
...
     @classmethod
+    @requires_modules("elasticsearch", extra="elasticsearch")
     def from_client_params(cls, params: BaseDBWriter.ClientParams) -> "ElasticsearchWriterClient":
+        from elasticsearch import Elasticsearch
+
         assert isinstance(params, ElasticsearchWriterClientParams)
         client = Elasticsearch(params.url, **params.es_client_args)
         return ElasticsearchWriterClient(client)
 
+    @requires_modules("elasticsearch.helpers", extra="elasticsearch")
```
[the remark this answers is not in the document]
bsowell: Yeah, I think you are probably right. I was kind of going back and forth on the right way to be defensive here -- either over-decorating to try to provide helpful errors in the case that the entry point changes or people manually do something weird or only decorating the entry-point to make it easier to change extras in the future. 

I probably leaned too far into the first approach. I'll try to remove a bit of redundancy.

lib/sycamore/sycamore/writer.py
bsowell: Done.

PR #742, opened 2024-08-29
Add materialize to the ntsb loader for luna

eric-anderson:
Also add a simple query script, and fix some version consistency issues.


examples/luna/ntsb_loader.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+context = sycamore.init()
+docset = context.read.binary(SOURCE_DATA_PATH, binary_format="pdf")
+tokenizer = HuggingFaceTokenizer("thenlper/gte-small")
+llm = OpenAI(OpenAIModels.GPT_3_5_TURBO.value)
+
+# partitioning docset
+partitioned_docset = (
+    docset.partition(partitioner=UnstructuredPdfPartitioner())
```
mdwelsh, line 122: Can we use ArynPartitioner please

examples/luna/ntsb_loader.py
```diff
@@ -0,0 +1,148 @@
...
+    # these are here mostly as examples; the last materialize will at this point take
+    # effect, you can use these during testing.
+    .materialize(path="/tmp/after_partition", source_mode=sycamore.MaterializeSourceMode.IF_PRESENT)
+    .map(add_property_to_schema)
+    .extract_properties(property_extractor=OpenAIPropertyExtractor(llm=llm, num_of_elements=35))
+    .materialize(path="/tmp/after_llm", source_mode=sycamore.MaterializeSourceMode.IF_PRESENT)
```
mdwelsh, line 128: Why are we doing this here - just as an example? Seems like this should be an option or the directories we're materializing to should be using `tempdir` (or at least named constants at the top of this file).

examples/luna/ntsb_loader.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env python
```
mdwelsh, line 1: We're not using the name "luna" anywhere in the code - this is Sycamore Query, so I'd recommend calling this directory `examples/query`.

examples/luna/ntsb_loader.py
```diff
@@ -0,0 +1,148 @@
...
+    .sketch()
+    .embed(embedder=SentenceTransformerEmbedder(batch_size=100, model_name="sentence-transformers/all-MiniLM-L6-v2"))
+    # comment out, to force re-evaluation of the pipeline, otherwise since this always exists it
+    # will be used.
+    .materialize(
+        path="s3://aryn-public/materialize/examples/luna/ntsb_loader",
+        source_mode=sycamore.MaterializeSourceMode.IF_PRESENT,
+    )
+    # materialize locally after reading from S3, it's a bit faster if you're running rmeotely
+    .materialize(path="/tmp/after_embed", source_mode=sycamore.MaterializeSourceMode.IF_PRESENT)
```
mdwelsh, line 142: Same here

pyproject.toml
```diff
@@ -36,6 +37,7 @@ remote-processors = { path = "lib/remote-processors", develop = true }
 integration = { path = "apps/integration", develop = true }
 sycamore-poetry-lock = { path = "lib/poetry-lock", develop = true }
 aryn-sdk = { path = "lib/aryn-sdk", develop = true }
+queryui = { path = "apps/query-ui", develop = true }
```
mdwelsh, line 40: I don't think this dependency needs to be here, right? queryui depends on sycamore, not the other way around.

eric-anderson: If it's not there, when you do a poetry install in the root you don't get the stuff to run the query ui.

examples/luna/ntsb_loader.py
eric-anderson: Done, but scares me a lot that as soon as I changed the partitioner I got different results.

examples/luna/ntsb_loader.py
eric-anderson: They're there as examples and to help while I was getting it working.

examples/luna/ntsb_loader.py
eric-anderson: Done.

examples/luna/ntsb_loader.py
eric-anderson: Also changed.

examples/luna/ntsb_loader.py
mdwelsh: Did you push your changes? This file is still called `examples/luna/ntsb_loader.py`

examples/luna/ntsb_loader.py
mdwelsh: I don't see changes here

mdwelsh (approved):

PR #743, opened 2024-08-29
Add initial version of ingest_ntsb_data script.

mdwelsh:
This simply migrates the original `ntsb_index_script.py` from the `query-engine` repo over to the Sycamore repo, so it can be used for standing up demos.


mdwelsh:
Closing in favor of #742

PR #745, opened 2024-08-30
Add PaddleOCR and Refactor Text Extraction

karanataryn:
- Refactors the Text Extraction interface to contain both PDFMiner and OCR, allowing them to act as part of the pipeline per page or supplement text to the whole document.
- Keeps the old OCR implementation under the `per_element_ocr` flag. 
- Adds PaddleOCR, Tesseract, EasyOCR, and Legacy (Easy + Tesseract) as OCR options.
- Refactors code to make it cleaner and changes necessary documentation.


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -449,20 +447,20 @@ def _partition_pdf_sequenced(
...
 
     def _partition_pdf_batched(
         self,
         file: BinaryIO,
         threshold: float = 0.4,
-        use_ocr=False,
-        ocr_images=False,
-        ocr_tables=False,
-        extract_table_structure=False,
+        use_ocr: bool = False,
```
bsowell, line 457: At some point after you get the paddle stuff in and you get the pdfminer stuff in, one of us should do a pass to clean up how we factor these methods and handle parameters. It's ridiculous how many times you have had to copy these changes, and we have seen that it makes it easy to make mistakes.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -778,137 +811,53 @@ def _get_hash_key(self, image: Image.Image, threshold: float) -> str:
...
-@requires_modules("pytesseract", extra="local-inference")
 def extract_ocr(
-    images: list[Image.Image], elements: list[list[Element]], ocr_images=False, ocr_tables=False, table_reader=None
+    images: list[Image.Image],
+    elements: list[list[Element]],
+    ocr_images: bool = False,
+    ocr_model_name: str = "easy",
 ) -> list[list[Element]]:
-    import pytesseract
+    if ocr_model_name == "paddle":
```
bsowell, line 824: Okay, if you've got this code twice, then definitely factor it into a separate method.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @timetrace("OCR Extraction")
+    def extract(self, filename: Union[str, IOBase], hash_key: str, use_cache=False, **kwargs) -> List[List[Element]]:
+        cached_result = ocr_cache.get(hash_key) if use_cache else None
+        if cached_result:
+            logger.info(f"Cache Hit for OCR. Cache hit-rate is {ocr_cache.get_hit_rate()}")
+            return cached_result
+        else:
+            with tempfile.TemporaryDirectory() as tempdirname:  # type: ignore
+                if isinstance(filename, IOBase):
```
bsowell, line 40: Why write out to a file here? Is this to communicate between the processes? I think all of the ocr models can take a file-like object.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,184 @@
...
+                        data = filename.read()
+                        temp_file.write(data)
+                        del data
+                        temp_file.flush()
+                    temp_file.close()
+                    file_name = temp_file.name
+                else:
+                    file_name = filename
+                pages = []
+                for path in pdf_to_image_files(file_name, Path(tempdirname)):
```
bsowell, line 53: I see. So this is still doing double conversion? Is it possible to do the pdf 2 image conversion once, and then share that across both the detr and text_extraction processes?

lib/sycamore/sycamore/transforms/text_extraction/text_extractor.py
```diff
@@ -0,0 +1,13 @@
...
+from sycamore.data import Element
+from io import IOBase
+
+
+class TextExtractor:
+    @abstractmethod
+    def extract(self, filename: Union[str, IOBase], hash_key: str, use_cache=False, **kwargs) -> List[List[Element]]:
+        pass
+
+    def __name__(self):
```
bsowell, line 12: What's __name__ for? Doesn't look like you are using it or overriding it anywhere.

lib/sycamore/pyproject.toml
```diff
@@ -68,6 +68,7 @@ weaviate-client = { version = ">=4.6.1,<4.7", optional = true } # typechecking f
 
 # Local inference dependencies
 easyocr = { version = "^1.7.1", optional = true }
+paddleocr = { version = "^2.8.1", optional = true }
```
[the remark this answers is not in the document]
karanataryn: No you have to install it separately, I think it's better to leave it to the user to choose to install `paddlepaddle` or `paddlepaddle-gpu`

lib/sycamore/sycamore/tests/integration/connectors/neo4j/test_docset_to_neo4j.py
```diff
@@ -48,7 +48,7 @@ class Gave(BaseModel):
         .resolve_graph_entities(resolvers=[])
         .explode()
     )
-
+    print("checker")
```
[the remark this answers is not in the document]
karanataryn: Removed, had it for debugging

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: Agreed.

lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -48,6 +50,14 @@ def partition_file(
             default: False
         ocr_images: attempt to use OCR to generate a text representation of detected images.
             default: False
+        ocr_model: model to use for OCR. Choices are "easy", "paddle", "tesseract" and "legacy", which
```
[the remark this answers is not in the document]
karanataryn: Done.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -570,25 +581,52 @@ def _partition_pdf_batched_named(
...
+                model = PaddleOCR(use_gpu=paddle.device.is_compiled_with_cuda())
+            elif ocr_model == "legacy":
+                from sycamore.transforms.text_extraction import LegacyOCR
+
+                model = LegacyOCR()
+            elif ocr_model == "tesseract":
+                from sycamore.transforms.text_extraction import Tesseract
+
+                model = Tesseract()
+            else:
```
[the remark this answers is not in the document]
karanataryn: Makes sense, done

lib/sycamore/sycamore/transforms/text_extraction/text_extractor.py
karanataryn: Good point, I overrided it so that the objects can be identified from the superclass

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
karanataryn: Changed it

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -60,9 +60,6 @@ def _can_retry(e: BaseException) -> bool:
         return False
 
 
-pdf_miner_cache = DiskCache(str(Path.home() / ".sycamore/PDFMinerCache"))
```
alexaryn, line 63: Interesting dilemma here.  Changing to page-at-a-time makes it harder to cache PdfMiner output.  Yet, PdfMiner still represents a significant chunk of work that would benefit from caching.  Is there a way to stream PdfMiner output into the cache and stream it out?  The only approach that comes to mind is disk spooling, which I expect to be faster than PdfMiner itself.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -385,108 +379,99 @@ def _partition_pdf_batched_named(
...
         extract_images=False,
         batch_size: int = 1,
         use_cache=False,
     ) -> List[List["Element"]]:
         self._init_model()
+
         if extract_table_structure and not table_structure_extractor:
             table_structure_extractor = DEFAULT_TABLE_STRUCTURE_EXTRACTOR(device=self.device)
-
+        if not use_ocr:
```
alexaryn, line 396: With these two-branch simple conditionals, I prefer to see the positive case first.  It reads more easily without the not.  And hey, one word shorter!

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -385,108 +379,99 @@ def _partition_pdf_batched_named(
...
-                pdf_page = pdf_generator.__next__()
-                pdfminer_pages.append(self.pdf_extractor.extract_page(pdf_page))
-
-            assert len(pdfminer_pages) == len(deformable_layout)
-            with LogTime("pdfminer_supplement"):
-                for d, p in zip(deformable_layout, pdfminer_pages):
+            extracted_pages = []
+            with LogTime("text_extraction"):
+                for i in range(len(batch)):
+                    page = batch[i] if use_ocr else pdf_generator.__next__()
```
alexaryn, line 469: This is a weird construct.  I'd move the testing of `use_ocr` up and have two simpler loops.  It would be good if you didn't need to call `__next__` directly.  Also, I'm not sure if you have to deal with this, but what if the generator doesn't yield the same number of pages as batch?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -508,18 +493,37 @@ def process_batch(
...
+        use_cache: bool,
+        use_ocr: bool,
+        ocr_images: bool,
+        ocr_model: str,
+        images: Optional[list[Image.Image]] = None,
+    ):
+        kwargs = {"ocr_images": ocr_images, "images": images}
+        if not use_ocr:
+            ocr_model = "pdfminer"
+        if ocr_model not in EXTRACTOR_DICT.keys():
```
alexaryn, line 509: Avoid double-lookups.  Do something like:
```
model_cls = EXTRACTOR_DICT.get(ocr_model)
if not model_cls:
    raise ...
model = model_cls()
```

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -709,65 +708,38 @@ def _get_hash_key(self, image: Image.Image, threshold: float) -> str:
...
 def extract_ocr(
-    images: list[Image.Image], elements: list[list[Element]], ocr_images=False, ocr_tables=False, table_reader=None
+    images: list[Image.Image],
+    elements: list[list[Element]],
+    ocr_images: bool = False,
+    ocr_model_name: str = "easyocr",
 ) -> list[list[Element]]:
-    import pytesseract
-
+    if ocr_model_name not in EXTRACTOR_DICT.keys():
```
alexaryn, line 717: Avoid double-lookup.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -709,65 +708,38 @@ def _get_hash_key(self, image: Image.Image, threshold: float) -> str:
...
             if elem.bbox is None:
                 continue
             if elem.type == "Picture" and not ocr_images:
                 continue
-            # elif elem.type == "table" and not ocr_tables:
-            #     continue
-            elif elem.type == "table":
+            cropped_image = crop_to_bbox(image, elem.bbox)
+            if elem.type == "table":
+                tokens = []
```
alexaryn, line 730: Maybe a function here?
```
elem.tokens = fixup_bboxes(elem, cropped_image)
```

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,216 @@
...
+class EasyOcr(OcrModel):
+    @requires_modules("easyocr", extra="local-inference")
+    def __init__(self, lang_list=["en"]):
+        import easyocr
+
+        self.reader = easyocr.Reader(lang_list=lang_list)
+
+    def get_text(self, image: Image.Image) -> str:
+        image_bytes = BytesIO()
+        image.save(image_bytes, format="PNG")
```
alexaryn, line 105: Do we need PNG?  Compression is slow and we have enough space for these transient files.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,216 @@
...
+        self.use_gpu = paddle.device.is_compiled_with_cuda()
+        self.language = language
+        self.reader = PaddleOCR(lang=self.language, use_gpu=self.use_gpu)
+
+    def get_text(
+        self,
+        image: Image.Image,
+    ) -> str:
+        bytearray = BytesIO()
+        image.save(bytearray, format="PNG")
```
alexaryn, line 198: Can we avoid PNG?

alexaryn (commented):
Looks generally OK to me.  I'd try to avoid calling PdfMiner an OCR model in the code, just to avoid confusion, but I get that is serves a similar purpose.

karanataryn:
> Looks generally OK to me. I'd try to avoid calling PdfMiner an OCR model in the code, just to avoid confusion, but I get that is serves a similar purpose.

It's a Text Extractor not an OCR Model in the code. Only the OCR Models are called as the latter.

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: This isn't used anywhere else, we can do this when we resolve the TODO.

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: Done.

lib/sycamore/sycamore/tests/integration/transforms/test_partition.py
```diff
@@ -24,14 +23,14 @@ def test_detr_ocr():
...
-        .partition(SycamorePartitioner(ARYN_DETR_MODEL, use_ocr=True))
+        .partition(ArynPartitioner(use_ocr=True))
         .explode()
         .filter(lambda doc: "page_number" in doc.properties and doc.properties["page_number"] == 1)
         .filter(lambda doc: doc.type in {"Section-header", "Title"})
         .take_all()
     )
 
-    assert "Attention Is All You Need" in set(str(d.text_representation).strip() for d in docs)
+    assert "Attention Is All You Need" or "Abstract" in set(str(d.text_representation).strip() for d in docs)
```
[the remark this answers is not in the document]
karanataryn: You're right, good catch. Looks like the test seems to work properly without it now so I've removed it.

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: Agree that there isn't a clear solution here. I think we should wait on this until after this PR, and decide on an approach for caching for both OCR and PDFMiner.

lib/sycamore/pyproject.toml
```diff
@@ -71,6 +71,7 @@ weaviate-client = { version = ">=4.6.1,<4.7", optional = true } # typechecking f
...
```
bsowell, line 74: When you install the paddleocr package, does it automatically install the cpu paddlepaddle, or do you have install it separately. (I think it's okay either way, just curious).

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -405,7 +417,8 @@ def __init__(
         threshold: Optional[Union[float, Literal["auto"]]] = None,
         use_ocr: bool = False,
         ocr_images: bool = False,
-        ocr_tables: bool = False,
+        ocr_model: str = "easyocr",
+        per_element_ocr: bool = False,
```
bsowell, line 421: I know we had some discussions about what the default should be for this. I thought you had said that maybe we would continue to do per_element_ocr by default. Is that still the case?

bsowell (commented):
I think I'm happy with this. Just want to understand what we are setting the defaults to.

lib/sycamore/pyproject.toml
karanataryn: No you'll have to install it. I wanted to leave it up to the user since they might want to install `paddlepaddle-gpu` instead depending on their machine.

lib/sycamore/sycamore/transforms/partition.py
karanataryn: That's true, good catch. I'll update it.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
+        return self.easy_ocr.get_boxes_and_text(image)
+
+    def __name__(self):
+        return "LegacyOcr"
+
+
+class PaddleOcr(OcrModel):
+    @requires_modules("paddleocr", extra="local-inference")
```
bsowell, line 182: Is there a way to declare a dependency for paddlepaddle here? Is the module name the same across CPU/GPU, or does it differ?

karanataryn: It's `paddlepaddle` for CPU and `paddlepaddle-gpu` for GPU. I can make a comment noting this here.

bsowell: That's the name of the package, though, right? The module that you import is just called `paddle`? So I think you could still do `@requires_modules(["paddleocr", "paddle"], extra="local-inference")` . That will at least fail at runtime if you don't have either version of paddle installed. The error message won't tell you exactly what to install, but at least it will fail.

bsowell (approved):
One minor comment, but otherwise I think this looks okay.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
karanataryn: Ah I see, yes that works.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+                text.text_representation = obj["text"]
+                texts.append(text)
+        return texts
+
+    @timetrace("OCR Document Extraction")
+    def extract_document(
+        self, filename: Union[str, IOBase], hash_key: str, use_cache=False, **kwargs
+    ) -> List[List[Element]]:
+        cached_result = ocr_cache.get(hash_key) if use_cache else None
+        if cached_result:
```
alexaryn, line 59: Seems like a good place to use the walrus operator:
```
if use_cache and (cached_result := ocr_cache.get(hash_key)):
    logger.info(...)
    return cached_result
```

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+            logger.info(f"Cache Hit for OCR. Cache hit-rate is {ocr_cache.get_hit_rate()}")
+            return cached_result
+        else:
+            with tempfile.TemporaryDirectory() as tempdirname:  # type: ignore
+                filename = cast(str, filename)
+                images = kwargs.get("images")
+                generator = (image for image in images) if images else pdf_to_image_files(filename, Path(tempdirname))
+                pages = []
+                for path in generator:
+                    if isinstance(path, Image.Image):
```
alexaryn, line 69: Maybe shouldn't call this `path` if it's likely to be an Image.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+                        image = path
+                    else:
+                        image = Image.open(path).convert("RGB")
+                    ocr_output = self.get_boxes_and_text(image)
+                    width, height = image.size
+                    texts: List[Element] = []
+                    for obj in ocr_output:
+                        if obj["bbox"] and not obj["bbox"].is_empty() and obj["text"] and len(obj["text"]) > 0:
```
alexaryn, line 77: See above.  Reduce redundancy.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+        val = self.pytesseract.image_to_string(image)
+        return val
+
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
+        output_list = []
+        base_dict = self.pytesseract.image_to_data(image, output_type=self.pytesseract.Output.DICT)
+        for value in zip(
+            base_dict["left"], base_dict["top"], base_dict["width"], base_dict["height"], base_dict["text"]
+        ):
+            if value[4] != "":
```
alexaryn, line 150: Should be `if value[4]:` unless you want a None value to pass the condition.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+
+        get_logger().setLevel(logging.ERROR)
+        self.use_gpu = paddle.device.is_compiled_with_cuda()
+        self.language = language
+        self.reader = PaddleOCR(lang=self.language, use_gpu=self.use_gpu)
+
+    def get_text(
+        self,
+        image: Image.Image,
+    ) -> str:
```
alexaryn, line 198: This seems like it should be on one line.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+        bytearray = BytesIO()
+        image.save(bytearray, format="BMP")
+        result = self.reader.ocr(bytearray.getvalue(), rec=True, det=True, cls=False)
+        return ans if result and result[0] and (ans := " ".join(value[1][0] for value in result[0])) else ""
```
alexaryn, line 202: This is complicated and reads weird because `ans` appears before it's assigned.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
+        bytearray = BytesIO()
+        image.save(bytearray, format="BMP")
+        result = self.reader.ocr(bytearray.getvalue(), rec=True, det=True, cls=False)
+        out = []
```
alexaryn, line 208: Can we type hint this `out: list[dict[str, Any]] = []`?

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+    def get_text(
+        self,
+        image: Image.Image,
+    ) -> str:
+        bytearray = BytesIO()
+        image.save(bytearray, format="BMP")
+        result = self.reader.ocr(bytearray.getvalue(), rec=True, det=True, cls=False)
+        return ans if result and result[0] and (ans := " ".join(value[1][0] for value in result[0])) else ""
+
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
```
alexaryn, line 204: lowercase list

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+
+    @requires_modules(["easyocr", "pytesseract"], extra="local-inference")
+    def __init__(self):
+        self.tesseract = Tesseract()
+        self.easy_ocr = EasyOcr()
+
+    def get_text(self, image: Image.Image) -> str:
+        return self.tesseract.get_text(image)
+
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
```
alexaryn, line 174: lowercase list

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+    def __init__(self):
+        import pytesseract
+
+        self.pytesseract = pytesseract
+
+    def get_text(self, image: Image.Image) -> str:
+        val = self.pytesseract.image_to_string(image)
+        return val
+
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
```
alexaryn, line 144: lowercase list

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+        image.save(image_bytes, format="BMP")
+        raw_results = self.reader.readtext(image_bytes.getvalue())
+        out_list = []
+        for res in raw_results:
+            text = res[1]
+            out_list.append(text)
+        val = " ".join(out_list)
+        return val
+
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
```
alexaryn, line 114: lowercase list

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+                    obj["bbox"].y2 / height,
+                )
+                text.text_representation = obj["text"]
+                texts.append(text)
+        return texts
+
+    @timetrace("OCR Document Extraction")
+    def extract_document(
+        self, filename: Union[str, IOBase], hash_key: str, use_cache=False, **kwargs
+    ) -> List[List[Element]]:
```
alexaryn, line 57: lowercase list

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+    @abstractmethod
+    def get_text(self, image: Image.Image) -> str:
+        pass
+
+    @abstractmethod
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
+        pass
+
+    @timetrace("OCR Page Extraction")
+    def extract_page(self, page: Union["Image.Image", "LTPage"]) -> List[Element]:
```
alexaryn, line 35: lowercase list

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+class OcrModel(TextExtractor):
+
+    @abstractmethod
+    def get_text(self, image: Image.Image) -> str:
+        pass
+
+    @abstractmethod
+    def get_boxes_and_text(self, image: Image.Image) -> List[Dict[str, Any]]:
```
alexaryn, line 31: lowercase list and dict

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -62,15 +65,16 @@ def _convert_bbox_coordinates(
         y2 = height - y2
         return x1, y1, x2, y2
 
+    @timetrace("PdfMiner Document Extraction")
```
alexaryn, line 68: Try `PdfMinerEx`

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -90,7 +94,10 @@ def extract_document(self, filename: str, hash_key: str, use_cache=False, **kwar
                 pdf_miner_cache.set(hash_key, pages)
             return pages
 
-    def extract_page(self, page: "LTPage") -> List[Element]:
+    def extract_page(self, page: Union["LTPage", "Image"]) -> List[Element]:
```
alexaryn, line 97: lowercase list

lib/sycamore/sycamore/transforms/text_extraction/text_extractor.py
```diff
@@ -0,0 +1,23 @@
...
+from io import IOBase
+
+if TYPE_CHECKING:
+    from PIL.Image import Image
+    from pdfminer.layout import LTPage
+
+
+class TextExtractor:
+    @abstractmethod
+    def extract_page(self, filename: Union["Image", "LTPage"]) -> List[Element]:
```
alexaryn, line 13: lowercase list

lib/sycamore/sycamore/transforms/text_extraction/text_extractor.py
```diff
@@ -0,0 +1,23 @@
...
+        pass
+
+    @abstractmethod
+    def extract_document(
+        self, filename: Union[str, IOBase], hash_key: str, use_cache=False, **kwargs
```
alexaryn, line 18: lowercase list

alexaryn (commented):
A number of comments.  I read it in detail.

lib/sycamore/sycamore/tests/unit/transforms/test_partition.py
```diff
@@ -180,19 +180,16 @@ def test_sycamore_partitioner_elements_reorder(self) -> None:
...
+        new_elems = [LegacyOcr().get_boxes_and_text(image=image) for image in images]
 
         assert len(new_elems) == 1
-        assert len(new_elems[0]) == 1
-
-        text = new_elems[0][0].text_representation
-        assert text is not None
-        assert text.strip() == "The quick brown fox"
+        text_list = []
+        text_list.extend(val["text"] for val in new_elems[0])
```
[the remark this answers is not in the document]
karanataryn: Done.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -414,110 +408,109 @@ def _partition_pdf_batched_named(
...
+        ocr_model: str = "easyocr",
+        per_element_ocr: bool = True,
         extract_table_structure=False,
         table_structure_extractor=None,
         extract_images=False,
         batch_size: int = 1,
         use_cache=False,
     ) -> List[List["Element"]]:
         self._init_model()
+
```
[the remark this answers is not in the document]
karanataryn: Unsure, have removed it now.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -414,110 +408,109 @@ def _partition_pdf_batched_named(
...
         if extract_table_structure and not table_structure_extractor:
             table_structure_extractor = DEFAULT_TABLE_STRUCTURE_EXTRACTOR(device=self.device)
-
+        if use_ocr:
+            text_extractor = EXTRACTOR_DICT[ocr_model]()
+            text_generator = None
+        else:
+            text_extractor = PdfMinerExtractor()
+            text_generator = PdfMinerExtractor.pdf_to_pages(filename)
```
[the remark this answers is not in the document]
karanataryn: Added.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -150,7 +145,8 @@ def partition_pdf(
         threshold: Union[float, Literal["auto"]] = DEFAULT_LOCAL_THRESHOLD,
         use_ocr=False,
         ocr_images=False,
-        ocr_tables=False,
+        ocr_model="easyocr",
```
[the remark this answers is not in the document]
karanataryn: It is an OCR Model. If `use_ocr` is off, this is ignored.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -740,68 +746,42 @@ def _get_hash_key(self, image: Image.Image, threshold: float) -> str:
 
 
 @timetrace("OCR")
-@requires_modules("pytesseract", extra="local-inference")
 def extract_ocr(
-    images: list[Image.Image], elements: list[list[Element]], ocr_images=False, ocr_tables=False, table_reader=None
+    images: list[Image.Image],
+    elements: list[list[Element]],
+    ocr_images: bool = False,
+    ocr_model_name: str = "easyocr",
```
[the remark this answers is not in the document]
karanataryn: Changed.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -740,68 +746,42 @@ def _get_hash_key(self, image: Image.Image, threshold: float) -> str:
...
+    model_cls = EXTRACTOR_DICT.get(ocr_model_name)
+    if not model_cls:
+        raise ValueError(f"Unknown OCR Model: {ocr_model_name}")
+    ocr_model: OcrModel = model_cls()
     for i, image in enumerate(images):
-        width, height = image.size
-
         page_elements = elements[i]
-
+        width, height = image.size
```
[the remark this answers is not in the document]
karanataryn: Left it in for now since the diffs are in.

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: I'll get to this in a follow-up PR, I'd like for the TODO to not be ignored.

PR #752, opened 2024-08-30
Remove unnecessary tracing code.

mdwelsh:
Since the Sycamore Query executor uses `.materialize()`, it is not necessary to separately write traces.


baitsguy (approved):
Nice, will think of a way to get rid of node_args as well

PR #754, opened 2024-08-31
Enforce the constraint that each cell is only in one spanning cell.

bsowell:
The table transformer post-processing code had some code (in nms_supercells) that attempted to remove overlapping spanning cells, but it didn't actually prevent a single cell from being put in multiple spanning cells. This could create strange issues with two cells occupying the same space, which screws up html rendering, among other things.

This change assigns each cell to at most one spanning cell. If it intersects multiple spanning cells, we pick the one with the largest intersection.


HenryL27 (approved):
[supercell!](https://images.app.goo.gl/aej9D9P9FTqbtA4A8)

PR #755, opened 2024-09-03
Dependencies 3/3: Add partitioning extras.

bsowell:
This adds the local-inference and legacy-partitioners extras. The former includes all the dependencies required for running the partitioning model locally, as well as sentence-transformer embeddings. The latter includes support for the
UnstructuredPdfPartitioner.


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -649,6 +649,8 @@ def __call__(self, image: List[Image.Image], threshold: float) -> List[List[Elem
 
 
 class DeformableDetr(SycamoreObjectDetection):
+
+    @requires_modules("transformers", extra="local-inference")
```
HenryL27, line 653: transformers gets pulled in by sentence-transformers then?

lib/sycamore/sycamore/utils/pytorch_dir.py
```diff
@@ -18,10 +15,16 @@ def get_default_build_root() -> str:
...
     """
+    # copied from the torch code.  Importing the file that contains it uses 4GB. This uses 100M
+    import torch._appdirs
+
     return os.path.realpath(torch._appdirs.user_cache_dir(appname="torch_extensions"))
 
 
+@requires_modules("torch", extra="local-inference")
 def get_pytorch_build_directory(name: str, verbose: bool) -> str:
+    import torch.version
```
HenryL27, line 26: I wonder if we can use this to avoid importing all of torch in that one branch in `choose_device` too?

HenryL27 (approved):

lib/sycamore/sycamore/transforms/detr_partitioner.py
bsowell: Yes, but why is that relevant here? We only use sentence-transformers in the embedder, right? (Not disagreeing, just trying to understand if I missed something).

lib/sycamore/sycamore/utils/pytorch_dir.py
bsowell: Sounds like a good idea. I'll try it.

lib/sycamore/sycamore/transforms/detr_partitioner.py
HenryL27: bringing it up cuz I didn't see it marked as a dep in the extra

bsowell: Oh yeah. It seems like we never had an explicit dependency on transformers, so I never moved it. Probably worth adding. Will do.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -571,6 +570,7 @@ def _run_pdfminer(pdf_path, hash_key, use_cache):
 
         return pdfminer_layout
 
+    @requires_modules("easyocr", extra="local-inference")
```
karanataryn, line 573: Instead of doing this, why don't we shift the initialization of the ocr table reader to the `extract_ocr` method? Would keep both `pytesseract` and `easyocr` together which makes more sense in my mind

karanataryn (approved):

lib/sycamore/sycamore/transforms/detr_partitioner.py
bsowell: I like that idea from a clarity perspective, but it does mean that the reader would be initialized once per doc, rather than once per ArynPdfPartitioner instance. Not sure that matters much, though I guess if it did it would be for the service which uses long-lived actors. 

I think it's reasonable to experiment with, but I might suggest we push that refactoring to a different PR just to reduce complexity here.

lib/sycamore/sycamore/utils/pytorch_dir.py
bsowell: Actually, looks like that branch isn't executed any more -- we just return cpu on the mac as it seemed to be faster. I was going to change it anyway, but it looks like importing torch.version actually imports torch? (https://github.com/pytorch/pytorch/blob/7600e9b36fd67f09c4bb8d91e6e5a78530c651ac/torch/version.py.tpl#L11). Confusing enough that I'm not going to worry about it at the moment.

PR #756, opened 2024-09-03
Dependencies: Remove need to restart colab session for aryn-sdk

bsowell:
The aryn-sdk depended on a newer version of PIL than was installed in Colab, which required a runtime restart to install. This change lowers the required PIL version to avoid this, which is similar to what we did for Sycamore.


HenryL27 (approved):

mashah:
LGTM

PR #760, opened 2024-09-03
Default llm in transforms

baitsguy:
Allow custom context object in query_client.run_plan/query(..). This allows you to pass in your own objects for query execution (e.g. custom llm).

```python
context = sycamore.init(params={OperationTypes.BINARY_CLASSIFIER: {"llm": MyFancyLLM()}})
query_client.query("How many dogs are cats?", index="catdog", context=context)
...
# llm_filter will use MyFancyLLM
```

Also includes a few more fixes/updates in the lib.

### DIFF
--- a/lib/sycamore/sycamore/context.py
+++ b/lib/sycamore/sycamore/context.py
@@ -1,7 +1,7 @@
 import functools
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Any, Callable, Optional, Union
+from typing import Any, Callable, Optional, Union, List
 
 from sycamore.plan_nodes import Node, NodeTraverse
 
@@ -52,6 +52,33 @@ class Context:
         return DocSetReader(self)
 
 
+def get_val_from_context(
+    context: "Context", val_key: str, param_names: List[str], ignore_default: bool = False
+) -> Optional[Any]:
+    """
+    GIven a Context object, return the possible value for a given val.
+    This assumes context.params is not a nested dict.
+    @param context: Context to use
+    @param val_key: Key for the value to be returned
+    @param param_names: List of parameter namespaces to look for.
+        Always uses OperationTypes.DEFAULT unless configured otherwise.
+    @param ignore_default: disable usage for OperationTypes.DEFAULT parameter namespace
+    @return: Optional value given configs.
+    """
+    if not context.params:
+        return None
+
+    for param_name in param_names:
+        cand = context.params.get(param_name, {}).get(val_key)
+        if cand is not None:
+            return cand
+
+    if not ignore_default:
+        return context.params.get(OperationTypes.DEFAULT.value, {}).get(val_key)
+
+    return None
+
+
 def context_params(*names):
     """
     Applies kwargs from the context to a function call. Requires 'context': Context, to be an argument to the method.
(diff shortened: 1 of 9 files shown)


lib/sycamore/sycamore/query/client.py
```diff
@@ -145,13 +146,14 @@ def generate_plan(self, query: str, index: str, schema: OpenSearchSchema) -> Log
         plan = planner.plan(query)
         return plan
 
-    def run_plan(self, plan: LogicalPlan, dry_run=False, codegen_mode=False) -> Tuple[str, str]:
+    def run_plan(
```
mdwelsh, line 149: Why don't we just associate a context with the SycamoreQueryClient?

lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -53,7 +54,6 @@ class SycamoreExecutor:
     def __init__(
         self,
         context: Context,
-        os_client_args: Any,
         s3_cache_path: Optional[str] = None,
```
mdwelsh, line 57: If we're gonna move the os_client_args out of here, it seems to me that s3_cache_path should be part of the context as well.

lib/sycamore/sycamore/context.py
```diff
@@ -52,6 +52,33 @@ def read(self):
         return DocSetReader(self)
 
 
+def get_val_from_context(
```
mdwelsh, line 55: This function should have its own unit test.

lib/sycamore/sycamore/query/execution/sycamore_executor.py
baitsguy: oh yeah this was a miss

lib/sycamore/sycamore/context.py
baitsguy: good catch I missed

lib/sycamore/sycamore/query/client.py
baitsguy: because that would make too much sense :)

lib/sycamore/sycamore/query/client.py
```diff
@@ -114,6 +117,15 @@ def __init__(
         self.os_client_args = os_client_args
         self.trace_dir = trace_dir
 
+        if context and os_client_args:
```
mdwelsh, line 120: Should we just get rid of these extra params and require them to be set in the context?

lib/sycamore/sycamore/query/client.py
```diff
@@ -114,6 +117,15 @@ def __init__(
...
+            raise AssertionError("If using a configured Context object, set os_client_args in context.params")
+        if context and s3_cache_path:
+            raise AssertionError("If using a configured Context object, set a cached llm in context.params")
+
+        self.context = context
```
mdwelsh, line 125: `self.context = context or self._get_default_context()`

mdwelsh (approved):
LGTM, a few nits

lib/sycamore/sycamore/query/client.py
baitsguy: I figured I'd do it separate since the UI is probably using it

lib/sycamore/sycamore/query/client.py
```diff
@@ -146,12 +158,10 @@ def generate_plan(self, query: str, index: str, schema: OpenSearchSchema) -> Log
         return plan
 
     def run_plan(self, plan: LogicalPlan, dry_run=False, codegen_mode=False) -> Tuple[str, str]:
+        assert self.context is not None, "Running a plan requires a configured Context"
```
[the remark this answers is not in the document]
baitsguy: mypy complains :( I think it's because the __init__ marks it as optional and it doesn't realize we're enforcing later in the method

PR #761, opened 2024-09-04
adding neo4j s3 proxy for aura db + split_calls flag for entity and relationship extractor.

RitxmSaha:
This PR probably should be 2 PR's but they are small enough that I think its OK to be one. Need to get this change along with another PR later with some changes for the datarobot demo tommorow.

split_calls is a flag to `extract_graph_entities()` and `extract_graph_relationships()` allowing them to split the schema among multiple calls(to allow for a better extraction as there is a less complex schema.

added flag use_auradb & s3_session to the neo4j writer that writes csv's to a aws bucket, creates temp signed urls, loads into neo4j auradb, and then deletes the bucket.

Will add test for auradb on thursday after I finish the datarobot demo.


lib/sycamore/sycamore/transforms/extract_graph_entities.py
```diff
@@ -34,9 +34,10 @@ class EntityExtractor(GraphEntityExtractor):
```
HenryL27, line 34: add doc for `split_calls`

lib/sycamore/sycamore/transforms/extract_graph_relationships.py
```diff
@@ -36,9 +36,10 @@ class RelationshipExtractor(GraphRelationshipExtractor):
 
     """
 
-    def __init__(self, llm: LLM, relationships: list[BaseModel] = []):
+    def __init__(self, llm: LLM, relationships: list[BaseModel] = [], split_calls: bool = False):
```
HenryL27, line 39: add split_calls doc

lib/sycamore/sycamore/transforms/extract_graph_relationships.py
```diff
@@ -120,9 +120,12 @@ async def _generate_relationships(self, section: HierarchicalDocument) -> dict:
...
             if start_nodes and end_nodes:
                 parsed_relations.append(relation)
-                parsed_metadata[relation.__name__] = {"start_label": start_label, "end_label": end_label}
-                parsed_nodes[start_label] |= set(start_nodes)
-                parsed_nodes[end_label] |= set(end_nodes)
+                parsed_metadata[relation.__name__] = {
+                    "start_label": start_label,
+                    "end_label": end_label,
+                    "start_nodes": set(start_nodes),
+                    "end_nodes": set(end_nodes),
```
HenryL27, line 127: wait are these really supposed to be sets? sets are unordered and stuff. what if a relation is not fully-connected?

lib/sycamore/sycamore/writer.py
```diff
@@ -563,6 +563,8 @@ def neo4j(
         uri: str,
         auth: Union[tuple[Any, Any], "Auth", "AuthManager", None],
         import_dir: str,
+        use_auradb: bool = False,
+        s3_session: Any = None,
```
HenryL27, line 567: What is s3_session suppposed to be? Ordinarily I would think you can do this all with boto3 / boto3.S3.Client ?

HenryL27 (commented):
mostly typing things

lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -58,10 +60,12 @@ def create_target_idempotent(self, target_params: BaseDBWriter.TargetParams):
...
 
-    def _write_nodes_neo4j(self, nodes: list[str], session: "Session"):
+    def _write_nodes_neo4j(self, nodes: list[tuple[str, str]], session: "Session"):
         for node_type in nodes:
-            file_url = f"file:///sycamore/nodes/{urllib.parse.quote(node_type)}"
-            node_label = node_type[:-4]
+            node_label = node_type[0]
+            file_url = node_type[1]
+            if "s3.amazonaws.com" not in file_url:
+                file_url = f"file:///sycamore/nodes/{urllib.parse.quote(file_url)}"
```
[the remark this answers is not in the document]
RitxmSaha: ohhhh yes the name is quite bad i'll change.

For the `s3.amazonaws.com` check its because signed urls are not aws uri's they look like this

`https://mybucket.s3.amazonaws.com/myfile.txt?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=YOUR_AWS_ACCESS_KEY_ID%2F20231101%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20231101T120000Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=EXAMPLE_SIGNATURE`

lib/sycamore/sycamore/writer.py
RitxmSaha: from boto.session import Session 

thats the type annotation we use now.

lib/sycamore/sycamore/transforms/extract_graph_relationships.py
RitxmSaha: yeah so this is not a big worry, its essentially just part of building context for the LLM. This stuff is passed to the prompt to generate it. I'm also going to be reworking logic around this area in the next PR. Here we create a set to of start nodes and end nodes and then pass them to the LLM as enums so it chooses a node that we have already extracted instead of hallucinating something.

lib/sycamore/sycamore/transforms/extract_graph_relationships.py
RitxmSaha: added

lib/sycamore/sycamore/transforms/extract_graph_entities.py
RitxmSaha: added

lib/sycamore/sycamore/transforms/extract_graph_relationships.py
HenryL27: oh right. You explained that to me at some point my bad

HenryL27 (approved):
I am gonna make you wait to make sure no new integration tests fail bc of this but lgtm

PR #762, opened 2024-09-04
Improve materialize

eric-anderson:
During discussion of the blog post, two things came up:
1. changing the constants used to define the source mode
2. adding clear_materialize() so people don't have to delete the source directories to easily force recomputation.


HenryL27 (approved):
lgtm

PR #765, opened 2024-09-05
Fix materialize to work even if it is re-executed on the same documents.

eric-anderson:
The groupby_count operator when run via ray results in re-execution of previous transforms.  The prior version of materialize meant this was an error because it would detect duplicate files being written. The updated code fixes this by including a sha checksum of the data in the default filename.  This change means that perfect duplicates will be dropped by materialize.

In order to preserve the safety properties, added a prepare operation on Nodes so that materialize nodes can clear their paths and then check that it was clear to make sure duplicate directories aren't being specified.


lib/sycamore/sycamore/materialize.py
```diff
@@ -50,7 +49,7 @@ def __init__(
         child: Optional[Node],
         context: Context,
         path: Optional[Union[Path, str, dict]] = None,
-        source_mode: MaterializeSourceMode = MaterializeSourceMode.OFF,
+        source_mode: MaterializeSourceMode = MaterializeSourceMode.RECOMPUTE,
```
mdwelsh, line 52: I think the docstrings documenting the different `source_mode` options in `docset.py` should be updated, yes? It would not hurt to document the different settings in the docstring for this class as well while you're at it.

lib/sycamore/sycamore/tests/unit/test_materialize.py
```diff
@@ -69,7 +69,7 @@ def test_write_str(self):
             self.check_files(tmpdir)
 
     def check_files(self, tmpdir, ext=""):
-        docs = glob.glob(tmpdir + "/doc_*" + ext)  # doc_id  is doc_#
+        docs = glob.glob(tmpdir + "/doc-*" + ext)  # doc_id  is doc_#
```
HenryL27, line 72: comment update?

lib/sycamore/sycamore/executor.py
```diff
@@ -87,6 +89,23 @@ def execute_iter(self, plan: Node, **kwargs) -> Iterable[Document]:
 
         plan.traverse(visit=lambda n: n.finalize())
 
+    def _prepare(self, plan: Node):
+        from queue import Queue
+
+        pending: Queue[Callable] = Queue()
```
HenryL27, line 95:
```
# Some prepare operations need to be executed in phases, so here's a queue for that
```

lib/sycamore/sycamore/materialize.py
```diff
@@ -92,9 +91,36 @@ def __init__(
 
         super().__init__(child, **kwargs)
 
+    def prepare(self):
```
HenryL27, line 94:
```
'''
Clean up the materialize location if necessary.
Validate that cleaning worked, but only once all materializes have finished cleaning.
This protects against multiple materializes pointing to the same location.
'''
```

HenryL27 (approved):
Here's roughly where and what comments I'd like

lib/sycamore/sycamore/materialize.py
eric-anderson: I updated that when I added the new ones. I found this as a place that I'd missed updating.  Since I'd potentially missed others I did the grep and have now fixed those as well. I fixed them to the new canonical sycamore.MATERIALIZE_RECOMPUTE and sycamore.MATERIALIZE_USE_STORED constants
I added a docstring pointer from the enum class to the DocSet materialize method. I'm not a fan of duplicate documentation (I find it gets out of date), but if you want it I can add it.

lib/sycamore/sycamore/executor.py
```diff
@@ -87,6 +89,23 @@ def execute_iter(self, plan: Node, **kwargs) -> Iterable[Document]:
...
+            f = n.prepare()
+            if f is not None:
+                pending.put(f)
+
+        plan.traverse(visit=visit)
+        while not pending.empty():
+            f = pending.get(block=False)
+            g = f()
+            if g is not None:
+                pending.put(g)
```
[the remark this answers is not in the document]
eric-anderson: Per discussion no.  If you clean and then check, the check is pointless (you just cleaned). If you check and then clean, any time you run the same script twice it will fail (because the clean file will be present from the previous run). So multiple passes are needed.

lib/sycamore/sycamore/tests/unit/test_materialize.py
eric-anderson: Fixed.

lib/sycamore/sycamore/executor.py
eric-anderson: Done, made it longer to cover the discussion and example.

lib/sycamore/sycamore/materialize.py
eric-anderson: Done.

PR #766, opened 2024-09-05
neo4j, refactor pipeline to not auto resolve entities + add support for images in pipeline.

RitxmSaha:


lib/sycamore/sycamore/connectors/neo4j/neo4j_writer.py
```diff
@@ -295,9 +295,9 @@ def _parse_docs(docs):
                 for property_key, property_value in value["properties"].items():
                     if property_key not in include_relationships:
                         if isinstance(property_value, list) or isinstance(property_value, dict):
-                            node[property_key] = json.dumps(property_value)
+                            rel[property_key] = json.dumps(property_value)
```
RitxmSaha, line 298: small bug i found

lib/sycamore/sycamore/transforms/extract_document_structure.py
```diff
@@ -16,6 +18,50 @@ def extract(document: Document) -> Union[Document, HierarchicalDocument]:
...
+    @staticmethod
+    def extract(doc: Document) -> HierarchicalDocument:
+        doc = HierarchicalDocument(doc.data)
+        doc.data["relationships"] = doc.get("relationships", {})
+        doc.data["label"] = doc.get("label", "DOCUMENT")
+
+        images = [child for child in doc.children if child.type == "Image"]
+        doc.children = []
+        for image in images:
+            initial_page = HierarchicalDocument(
```
HenryL27, line 31: You're gonna create a 'front page' doc for every image in a document? That seems not right.

lib/sycamore/sycamore/transforms/extract_document_structure.py
```diff
@@ -16,6 +18,50 @@ def extract(document: Document) -> Union[Document, HierarchicalDocument]:
...
+                "properties": {},
+                "START_ID": image.doc_id,
+                "START_LABEL": "IMAGE",
+                "END_ID": initial_page.doc_id,
+                "END_LABEL": "SECTION",
+            }
+            doc.data["relationships"][str(uuid.uuid4())] = rel_1
+            initial_page.data["relationships"][str(uuid.uuid4())] = rel_2
+            initial_page.children.append(image)
+            doc.children.append(initial_page)
```
HenryL27, line 61: so just to get this straight:
the new InitialPage doc is a SECTIONOF the document
the image is a PARTOF the new initialpage

lib/sycamore/sycamore/transforms/extract_document_structure.py
```diff
@@ -16,6 +18,50 @@ def extract(document: Document) -> Union[Document, HierarchicalDocument]:
...
 
+class StructureByImages(DocumentStructure):
+    @staticmethod
+    def extract(doc: Document) -> HierarchicalDocument:
+        doc = HierarchicalDocument(doc.data)
+        doc.data["relationships"] = doc.get("relationships", {})
+        doc.data["label"] = doc.get("label", "DOCUMENT")
+
+        images = [child for child in doc.children if child.type == "Image"]
+        doc.children = []
```
HenryL27, line 29: wait what if there's other elements?

lib/sycamore/sycamore/transforms/extract_document_structure.py
```diff
@@ -211,3 +255,13 @@ def summarize_sections(doc: HierarchicalDocument) -> HierarchicalDocument:
                 summary_list.append(elem_sum)
             section.data["summary"] = "".join(summary_list)
         return doc
+
+
+class ExtractImageSummaries(ExtractSummary):
+    @staticmethod
+    def summarize(doc: HierarchicalDocument) -> HierarchicalDocument:
+        for section in doc.children:
+            assert len(section.children) == 1 and section.children[0].type == "Image"
```
HenryL27, line 264: Maybe this can be more robust by doing something like 
```python
for child in section.children:
    if child.type == "Image":
        continue
```
?

lib/sycamore/sycamore/transforms/extract_graph_entities.py
```diff
@@ -15,6 +18,17 @@
...
 
+ENTITY_DEFAULT_PROMPT = """
+    -Instructions-
+    You are a information extraction system.
+
+    You will be given a sequence of data in different formats(text, table, Section-header) in order.
+    Your job is to extract entities from the text input that match the entity schemas provided. Each entity
+    and property extracted should directly reference part of the text input provided.
+
+    """
```
HenryL27, line 29: move to default prompts (something like `sycamore/llms/prompts/default_prompts.py`)

lib/sycamore/sycamore/transforms/extract_graph_relationships.py
```diff
@@ -16,6 +19,15 @@
...
 logger = logging.getLogger(__name__)
 
+RELATIONSHIP_DEFAULT_PROMPT = """
+    -Goal-
+    You are a helpful information extraction system.
+
+    You will be given a sequence of data in different formats(text, table, Section-header) in order.
+    Your job is to extract relationships that map between entities that have already been extracted from this text.
+
+    """
```
HenryL27, line 29: -> default prompts file

lib/sycamore/sycamore/transforms/extract_graph_entities.py
RitxmSaha: I realized this, but the thing is all those prompts are guidance prompts. These have to be test cause currently guidance is not compatible with structured outputs(which is what this stack uses). Should I still move them over?

lib/sycamore/sycamore/transforms/extract_document_structure.py
RitxmSaha: yes, thats right

lib/sycamore/sycamore/transforms/extract_document_structure.py
RitxmSaha: yeah this was because of how we organize a graph pipeline. We iterate over the children of each document. each child has a summary. So we attach images to each section(or front page doc), summarize each section(base64 encode the images), and then once its passed to openai that "summary" is passed in. I know it's weird but for this im not sure what a better abstraction would be that would support all types of summaries naturally.

lib/sycamore/sycamore/transforms/extract_document_structure.py
RitxmSaha: This is only for extracting images, which was datarobots specific use case. They wanted to extract each image as a DAG, and ignore text for now. There is probably a better abstraction that makes this more flexible but I'm not sure how it would look. I tried a different abstraction using functional programming but I was running into serialization issues left and right because of pydantic so I gave up on doing that.

lib/sycamore/sycamore/transforms/extract_document_structure.py
RitxmSaha: Sooo the thing is this function actually expects only one image per section when i was coding it up last week. This is screaming technical debt and I don't want to leave this code here. I have a meeting right now, but I'll stop by after to see if we can talk about what we can do to make these functions not finetuned to one specific use case.

lib/sycamore/sycamore/transforms/extract_graph_entities.py
HenryL27: yeah. they claim to be guidance prompts but really it's just an object with a particular string property. Someday we'll make a call on guidance and then it will be really nice to have all the prompts in one place

HenryL27 (approved):

PR #772, opened 2024-09-07
Allow custom SycamoreQueryClient in query-ui + cleanup

baitsguy:
Tested by creating a broken SycamoreQueryClient in the custom_client file and verifying things break.

Added an initialization line in generated code, that still needs iterating and some revamping.


mdwelsh (commented):
So, it seems like the only real customization happening here is setting context values when we do sycamore.init(), right? I wonder if we shouldn't have this instead be something where SycamoreQueryClient can read a local 'sycamore.yaml' file which has its various config parameters (falling back to default values if not present). WDYT?

baitsguy:
> So, it seems like the only real customization happening here is setting context values when we do sycamore.init(), right? I wonder if we shouldn't have this instead be something where SycamoreQueryClient can read a local 'sycamore.yaml' file which has its various config parameters (falling back to default values if not present). WDYT?

That makes it hard to send in custom class implementations (eg llm). Works fine for just scalar configs

mdwelsh (approved):

mdwelsh:
I don't love it but we can iterate :-)

PR #773, opened 2024-09-09
improving ExtractTableProperties and standardizer transforms

Soeb-aryn:


lib/sycamore/sycamore/transforms/extract_table_properties.py
```diff
@@ -52,42 +52,48 @@ def extract_parent_json(input_string: str) -> str:
...
+            prompt_find_table = """
+            You are given a text string where columns are separated by comma representing either a single column, 
+            or multi-column table each new line is a new row.
+            Instructions:
+            1. Parse the table and make decision if key, value pair information can be extracted from it.
+            2. if the table contains multiple cell value corresponding to one key, the key, value pair for such table 
+            cant be extracted.
+            3. return True if table cant be parsed as key value pair.
+            4. return only True or False nothing should be added in the response.
+            """
```
HenryL27, line 72: Can these default prompts be moved to the file with all the default prompts? I think it's like sycamore/llms/prompts.py or something.

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -31,9 +31,10 @@ def fixer(self, text: str) -> Union[str, Tuple[str, date]]:
         """
         pass
 
+    @abstractmethod
     def standardize(self, doc: Document, key_path: List[str]) -> Document:
         """
-        Applies the fixer method to a specific field in the document as defined by the key_path.
+        Abstract method applies the fixer method to a specific field in the document as defined by the key_path.
```
HenryL27, line 37: wait why is this an abstract method now?

HenryL27 (commented):
Ok, I think you can make the standardize method a `classmethod` - so it takes a `cls` parameter (where an instance method would take a `self` or a static method would take nothing) - then you can call the static `fixer` method from the superclass for each. e.g

```python

class A:
    @classmethod
    def foo(cls):
        cls.bar()
        print("AFoo")

    @abstractmethod
    @staticmethod
    def bar():
        pass

class B:
    @staticmethod
    def bar():
        print("BBar")

B.foo()

>>> "BBar"
>>> "AFoo"
```
I'm pretty sure this gives you the behavior you want without duplicating the path-y code everywhere and without instantiating a bunch of objects for no reason.

lib/sycamore/sycamore/transforms/standardizer.py
Soeb-aryn: standardize can have different implement for each child class depending on the required behavior, hence it is kept as Abstract method in the parent.

HenryL27 (approved):

PR #774, opened 2024-09-09
Sycamore changes to support new NTSB demo.

mdwelsh:
The main change here is to no longer require Sycamore Query to use SummarizeData as the final operator in the query chain. This allows us to get raw DocSet results out of a query plan, which can then be interpreted by the client app however it wishes.


baitsguy (approved):
Will want to validate any evaluation scripts don't break

PR #776, opened 2024-09-09
Adding query demo UI.

mdwelsh:
This PR is based off of `matt/luna-demo-sycamore` and I will rebase it on `main` once #774 lands.

This is the first cut at the new demo UI. It is not yet replacing the existing `query-ui` as I see the latter as more of an internal tool, while this demo is meant to be externally usable.

Running: `poetry run python -m streamlit run querydemo/Query_Demo.py`

The main idea here is to have the Sycamore Query client return the raw docset, and then let the LLM decide how to present the results. The prompt at the top tells the LLM to render MDX output, which allows us to embed JSX components representing things like tables, maps, etc. into the output. I then have a simple parser that walks the resulting output and spits out Markdown for the text, and specialized UI widgets for the other things.

This pattern can be generalized and extended over time to support a wider range of UI elements, but I wanted to start with this for now.

As part of this I have also refactored how the chat session works in terms of state management and control flow, which fixes some issues in the previous UI code.


baitsguy (approved):
I'm sure there's a bunch of subtleties I'm missing but better to get this in

PR #777, opened 2024-09-09
Fix Bugs in Sycamore Pipeline

karanataryn:
Fixes bug with batched partitioning (did not wait for pdfminer before extracting tables) and allows Tables to handle empty bboxes.

Without the fix, in non-ocr mode partitioning `basic_table.pdf` will fail to include the table output in the correct format (with the bounding box as a list), and instead return it as a `BoundingBox` type. This indicates that the table extraction model never saw the input tokens to help with extraction, and hence was not able to modify it to the correct format.


lib/sycamore/sycamore/data/bbox.py
```diff
@@ -161,7 +161,7 @@ def to_absolute_self(self, width, height) -> "BoundingBox":
         return self
 
     def is_empty(self):
-        return self.x1 >= self.x2 and self.y1 >= self.y2
+        return self.x1 >= self.x2 or self.y1 >= self.y2
```
eric-anderson, line 164: Suggest adding a test for this since it was a bug.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -554,7 +552,25 @@ def _partition_pdf_batched_named(
...
+        for i in batches:
+            self.process_batch_extraction(
+                i,
+                deformable_layout,
+                extract_table_structure=extract_table_structure,
+                table_structure_extractor=table_structure_extractor,
+                extract_images=extract_images,
+            )
+            assert len(parts) == len(i)
+            if tracemalloc.is_tracing():
```
eric-anderson, line 564: You can dump this tracemalloc code, it didn't turn out to be that useful.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -554,7 +552,25 @@ def _partition_pdf_batched_named(
             with LogTime("pdfminer_supplement"):
                 for d, p in zip(deformable_layout, pdfminer_layout):
                     self._supplement_text(d, p)
+        for i in batches:
```
eric-anderson, line 555: also suggest adding a test to catch the pdfminer dependency for extraction (it looks like that was the problem), to make sure it doesn't regress.

eric-anderson: Ok, I just took another look. This shouldn't be doing anything. batches is a generator, but it was consumed in the first pass.  If it's not, then we have a memory issue since you'd be keeping the entire pdf images in memory.

```
def generator(n):
    for i in range(n):
        print(f"yielding {n}")
        yield i

f = generator(10)
for i in f:
    print(f"got {i}")

print("LOOP 2")
for i in f:
    print(f"got {i}")
```

karanataryn: Fixed, thanks for catching this

karanataryn: Fixed

lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
```diff
@@ -64,3 +64,14 @@ def test_batched_sequenced(self):
             s, TEST_DIR / "resources/data/pdfs/basic_table.pdf", use_ocr=True, use_cache=False
         )
         assert len(d) == 1
+
```
eric-anderson, line 67: which of these fails without the fix?  Make sure to include that in the PR description (fixing the below to whatever is the correct statement):
```
Without the fix, in non-ocr mode partitioning basic_table.pdf will fail to include the table output.
```

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -552,7 +551,7 @@ def _partition_pdf_batched_named(
             with LogTime("pdfminer_supplement"):
                 for d, p in zip(deformable_layout, pdfminer_layout):
                     self._supplement_text(d, p)
-        for i in batches:
+        for i in convert_from_path_streamed_batched(filename, batch_size):
```
eric-anderson, line 554: `# TODO: optimize this to make pdfminer also streamed so we can process each page in sequence without having to double-convert the document`

lib/sycamore/sycamore/tests/unit/transforms/compare_detr_impls.py
```diff
@@ -11,6 +11,18 @@ def compare_batched_sequenced(partitioner, path, **kwargs):
     return batched
 
 
+def check_table_extraction(partitioner, path, **kwargs):
+    with open(path, "rb") as f:
+        sequenced = partitioner._partition_pdf_sequenced(f, **kwargs)
+        hash_key = Cache.get_hash_context(f.read()).hexdigest()
+    batched = partitioner._partition_pdf_batched_named(path, hash_key, **kwargs)
```
eric-anderson, line 18: So long as we are running both, might as well add 
`assert deep_eq(batched, sequenced)`

lib/sycamore/sycamore/tests/unit/data/test_bbox.py
```diff
@@ -2,10 +2,13 @@
...
 from math import isclose
 
-# Test bounding boxes. Box 1 and 2 intersect. Box 3 is intersects with Box 1 but not Box 2.
+# Test bounding boxes. Box 1 and 2 intersect. Box 3 is intersects with Box 1 but not Box 2. Box 4 and 5
+# are disjoint from all others and are empty.
 bbox1 = BoundingBox(20.0, 20.0, 60.0, 60.0)
 bbox2 = BoundingBox(10.0, 40.0, 40.0, 80.0)
 bbox3 = BoundingBox(50.0, 10.0, 90.0, 50.0)
+bbox4 = BoundingBox(100.0, 100.0, 100.0, 120.0)
+bbox5 = BoundingBox(100.0, 100.0, 100.0, 100.0)
```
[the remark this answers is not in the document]
karanataryn: Good point, added the assertion

lib/sycamore/sycamore/tests/unit/transforms/compare_detr_impls.py
karanataryn: Done

lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
karanataryn: Updated

lib/sycamore/sycamore/tests/unit/data/test_bbox.py
```diff
@@ -27,3 +30,11 @@ def test_union():
 def test_absolute_relative():
     assert bbox1.area == bbox1.to_relative(100, 100).to_absolute(100, 100).area
     assert isclose(bbox1.to_relative(100, 100).area, 0.16)
+
+
+def test_empty_bbox():
+    assert bbox5.is_empty()
+    assert bbox4.is_empty()
+    assert bbox4.area == 0 and bbox5.area == 0
+    assert bbox4.union_self(bbox5).area == 0 and bbox4.union_self(bbox5).is_empty()
```
bsowell, line 39: Do you want union here instead of union_self? I guess it doesn't matter since in each case the boxes are empty, but seems a bit weird to update bbox4 in place, and is counter to what you did with intersect below.

lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
bsowell: Where is this comment? Did you add it to one of the tests? Hopefully I'm not looking at an old version.

lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
```diff
@@ -64,3 +64,25 @@ def test_batched_sequenced(self):
...
+    def test_table_extraction_order(self):
+        s = ArynPDFPartitioner("Aryn/deformable-detr-DocLayNet")
+        d = check_table_extraction(
+            s,
+            TEST_DIR / "../../../../apps/crawler/crawler/http/tests/visit_aryn.pdf",
```
bsowell, line 72: minor: Maybe this is a normal thing for mono-repos, but I wonder if it might not be safer to put these pdfs in the sycamore library test directory just to avoid surprises if someone edits crawlers. I don't feel strongly, and I'm not thrilled with making the repo even bigger, so I'll leave it to you.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -554,7 +551,17 @@ def _partition_pdf_batched_named(
             with LogTime("pdfminer_supplement"):
                 for d, p in zip(deformable_layout, pdfminer_layout):
                     self._supplement_text(d, p)
-
+        # TODO: optimize this to make pdfminer also streamed so we can process each page in sequence without
+        # having to double-convert the document
```
bsowell, line 555: Just to make sure I understand correctly, this will double-convert even in cases when you aren't using pdfminer, right? That may be fine, but we need to do some performance measurements, I think. Can we get this on one of the test instances and run the load tests + measure latency?

bsowell (approved):

lib/sycamore/sycamore/tests/unit/data/test_bbox.py
karanataryn: Makes sense, have updated it

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: Will work on this after merging this in

lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
karanataryn: It is cleaner, and since it allows this to be more modular, I'll copy it over to the test directory as well

PR #784, opened 2024-09-11
A few Luna demo fixes.

mdwelsh:
This incorporates a version of Eric's workaround for Ray issues in the `eric-containers` branch into the new NTSB demo, and a few UI tweaks to make it more usable.


baitsguy (commented):
Can you port this to query-ui as well? we're diverging

mdwelsh:
Not yet since I want to see where we land with Eric's containers PR. Also, Mehul needs this for a demo tonight 😱

baitsguy (approved):

PR #787, opened 2024-09-11
Make reorder_elements more like sorted() so we can use key=

alexaryn:


lib/sycamore/sycamore/functions/elements.py
```diff
@@ -1,24 +1,32 @@
...
         comparator: A comparator function
+        key: A key as per sorted()
 
     Returns:
         Document with elements re-ordered
     """
+    if key:
+        assert not comparator
+    else:
+        assert comparator
```
HenryL27, line 26: add error message like "exactly one of comparator and key must be supplied"

HenryL27 (approved):

PR #790, opened 2024-09-12
Various Luna Demo fixes.

mdwelsh:
* Better PDF viewer.
* Show LLM prompt for query plan generation.
* Add "RAG only" toggle.

### DIFF
--- a/apps/query-demo/querydemo/Query_Demo.py
+++ b/apps/query-demo/querydemo/Query_Demo.py
@@ -476,16 +476,19 @@ def do_query():
                 with st.expander("Raw query response"):
                     st.write(tool_response)
 
-                if isinstance(tool_response, str):
-                    # We got a straight string response from the query plan, which means we can
-                    # feed it back to the LLM directly.
-                    tool_response_str = tool_response
-                elif isinstance(tool_response, sycamore.docset.DocSet):
-                    # We got a DocSet.
-                    tool_response_str = docset_to_string(tool_response)
-                else:
-                    # Fall back to string representation.
-                    tool_response_str = str(tool_response)
+                with st.spinner("Running Sycamore query..."):
+                    if isinstance(tool_response, str):
+                        # We got a straight string response from the query plan, which means we can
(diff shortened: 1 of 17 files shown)


apps/query-demo/querydemo/Query_Demo.py
```diff
@@ -364,43 +365,97 @@ def show(self):
...
-                    if self.chat_message.button("Next", use_container_width=True):
-                        st.session_state.pagenum += 1
-                with col4:
-                    self.chat_message.download_button(
-                        "Download", content, os.path.basename(self.path), "pdf", use_container_width=True
-                    )
-                pdf_viewer(content, pages_to_render=[st.session_state.pagenum], key=self.chat_message.next_key())
-
+            encoded = base64.b64encode(content).decode("utf-8")
+            pdf_display = f'<iframe src="data:application/pdf;base64,{encoded}" width="600" height="800" type="application/pdf"></iframe>'
```
baitsguy, line 369: nice

apps/query-demo/querydemo/Query_Demo.py
```diff
@@ -364,43 +365,97 @@ def show(self):
...
-        st.session_state.query_id, result = run_plan(sqclient, plan)
-    return result, plan
+def get_embedding_model_id() -> str:
+    """Get the embedding model ID from the OpenSearch instance."""
+    query = {
+        "query": {
+            "bool": {
+                "must": [
+                    {
+                        "match": {"name": "all-MiniLM-L6-v2"},
```
baitsguy, line 380: this is fine for now but we should use a better model when we actually demo for the comparison to at least be sorta fair.

baitsguy (approved):

apps/query-demo/querydemo/Query_Demo.py
mdwelsh: This is just the embedding model, though, right? IIUC it is using OpenAI for the actual RAG step, no?

PR #794, opened 2024-09-13
Add natural language result flag.

mdwelsh:
This PR adds a flag to the SycamoreQueryClient requesting a natural language response. This allows apps to request either natural language or raw data, based on what they need.

### DIFF
--- a/lib/sycamore/sycamore/query/planner.py
+++ b/lib/sycamore/sycamore/query/planner.py
@@ -107,7 +107,7 @@ class LlmPlanner:
         self._use_examples = use_examples
         self._natural_language_response = natural_language_response
 
-    def make_operator_prompt(self, operator: LogicalOperator) -> str:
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/query/planner.py
```diff
@@ -96,6 +105,7 @@ def __init__(
         self._os_client = os_client
         self._llm_client = llm_client or OpenAI(OpenAIModels.GPT_4O.value)
         self._use_examples = use_examples
+        self._natural_language_response = natural_language_response
```
baitsguy, line 108: Can we just use the same flag for the client and executor? They’re opposites i think?

mdwelsh: They are the same but I made the command line switch for the CLI the opposite (`--raw-data`) since I think that is less likely to be desired in the case of running the CLI.

baitsguy (approved):

PR #795, opened 2024-09-13
updated docs to explain the new default threshold setting for ArynPartitioner

dtecuci:


docs/source/aryn_cloud/aps_output.md
```diff
@@ -78,7 +78,7 @@ Takes the format `[x1, y1, x2, y2]` where each coordinate is given as the propor
       "page_number": 1-indexed page number the element occurs on }
 ```
 
-The `score` is the number used to determine whether to keep the element based on the `threshold`. Elements with a confidence score higher than the threshold will be kept as the model is "confident" in its prediction for this element whereas elements with a confidence score lower than this threshold will be removed, as the model is less confident in its prediction for this element.
+The `score` is the model's "confidence" in its prediction for that particular bounding box. By dafault the model makes its best prediction, but the user can control this using the `threshold` parameter (defaults to "auto"). If the user specifies a numeric value between 0 and 1, only Elements with a confidence score higher than the specified threshold value will be kept.
```
sohamkasar19, line 81: Misspelled default here.

dtecuci: I wouldn't. We're keeping that post-processing private for now.

docs/source/sycamore/transforms/partition.md
```diff
@@ -21,7 +21,7 @@ Parameters:
 
 * ```use_partitioning_service```: If ```True```, the partitioner uses the *Aryn  Partitioning Service*. Defaults to ```True```.
 * ```model_name_or_path```: The HuggingFace coordinates or model local path. It defaults to ```SYCAMORE_DETR_MODEL```, and you should only change it if you are testing a custom model. Ignored when ```use_partitioning_service``` is ```True```.
-* ```threshold```: The threshold to use for accepting the models predicted bounding boxes. A lower value will include more objects, but may have overlaps; a higher value will reduce the number of overlaps, but may miss legitimate objects. It defaults to ```0.4```.
+* ```threshold```: This represents the threshold for accepting the model's predicted bounding boxes. It defaults to "auto", where the model make its best prediction. The user can specify a numerical threshold between 0 and 1. A lower value will include more objects, but may have overlaps, a higher value will reduce the number of overlaps, but may miss legitimate objects. We suggest trying values around ``0.32``.
```
bsowell, line 24: Maybe change the recommendation at the end to use auto? You can maybe say something like "We recommend using the default "auto" threshold. If you do set the threshold manually, we recommend starting with a value of 0.32."

bsowell (approved):

docs/source/aryn_cloud/aps_output.md
sohamkasar19: Cool

docs/source/sycamore/transforms/partition.md
dtecuci: Updated. thanks!

dtecuci:
merged

docs/source/sycamore/transforms/partition.md
jonfritz: Is the model making the best prediction, or is it using the post processing logic?

PR #796, opened 2024-09-13
Add support for pushing query filters down to OpenSearch.

mdwelsh:
This PR has the Luna planner generate OpenSearch Query DSL when it can, to avoid the need for scanning the full index.

This is by no means a complete solution, since the planner does not always use the OpenSearch query functionality to the fullest extent, but it often does result in pushing down OpenSearch queries, such as the following.

We can definitely improve this with better prompting and examples but wanted to get this first version out there.

```
[
    {
        "operatorName": "QueryDatabase",
        "description": "Get all the incident reports in 2023",
        "index": "const_ntsb",
        "query": {
            "range": {
                "properties.entity.isoDateTime": {
                    "gte": "2023-01-01T00:00:00",
                    "lte": "2023-12-31T23:59:59",
                    "format": "strict_date_optional_time"
                }
            }
        },
        "node_id": 0
    },
    {
        "operatorName": "LlmFilter",
        "description": "Filter to only include incidents in Washington",
        "question": "Did this incident occur in Washington?",
        "field": "properties.entity.location",
        "input": [0],
        "node_id": 1
    },
    {
        "operatorName": "Count",
        "description": "Count the number of incidents in Washington in 2023",
        "distinct_field": "properties.entity.accidentNumber",
        "input": [1],
        "node_id": 2
    }
]```

### CODE REVIEW

**Review by baitsguy — COMMENTED** [review]
Looks good but can you add tests pls

**Comment by mdwelsh** [comment]
> Looks good but can you add tests pls

Done!

File: lib/sycamore/sycamore/tests/unit/query/execution/test_sycamore_operator.py [hunk 1]
```diff
@@ -57,6 +57,53 @@ def test_query_database(mock_sycamore_docsetreader, mock_opensearch_num_docs):
...
+        assert isinstance(result, DocSet)
+
+        # Validate result
+        assert result.count() == mock_opensearch_num_docs
+
+        # Validate that the correct query would be passed to OpenSearch.
+        assert result.plan._query_params.query == {"query": os_query_plan}
+
+
+#        assert mock_impl.query_params == "4949494"
```
baitsguy, line 98: probably delete before pushing

baitsguy (approved):

PR #799, opened 2024-09-13
Docs for nms change (take 2)

dtecuci:
changed the wording to include Jon's comments


docs/source/aryn_cloud/aps_output.md
```diff
@@ -78,7 +78,7 @@ Takes the format `[x1, y1, x2, y2]` where each coordinate is given as the propor
       "page_number": 1-indexed page number the element occurs on }
 ```
 
-The `score` is the number used to determine whether to keep the element based on the `threshold`. Elements with a confidence score higher than the threshold will be kept as the model is "confident" in its prediction for this element whereas elements with a confidence score lower than this threshold will be removed, as the model is less confident in its prediction for this element.
+The `score` is the model's "confidence" in its prediction for that particular bounding box. By dafault the model makes its best prediction, but the user can control this using the `threshold` parameter (defaults to "auto"). If the user specifies a numeric value between 0 and 1, only Elements with a confidence score higher than the specified threshold value will be kept.
```
bsowell, line 81: typo: dafault -> default

I find the wording "By dafault the model makes its best prediction" a bit awkward. I might say "By default, we automatically select bounding boxes to achieve good coverage with high prediction accuracy, but the user can control this ..." or something similar. 

I don't feel super strongly, so if you and Jon are happy with this, then I'm happy.

bsowell (approved):

docs/source/aryn_cloud/aps_output.md
jonfritz: I'm fine with Ben's update.

dtecuci: updated. thanks or catching these two.

dtecuci:
merged

bsowell (approved):

dtecuci:
done

PR #802, opened 2024-09-14
Verify .map can run parallel classes

eric-anderson:
* Add integration test confirming it works.
* Add documentation explaining how.
* Add support for viewing the readthedocs documentation before committing.


baitsguy (approved):
Thanks this helps. Curious why we chose to only support this in Map classes and not methods?

PR #806, opened 2024-09-16
Update type signatures for threshold in aryn_sdk.

bsowell:
The aryn_sdk already does the right thing in terms of what we are sending the service, but the type annotations were not compatible with specifying "auto". This just updates those annotations.


HenryL27 (approved):

HenryL27 (commented):
actually documentation

bsowell:
> actually documentation

@HenryL27 Ack, I'll fix the docstring. Is there documentation somewhere else we should update as well?

HenryL27:
I think the only other place to check would be the readme but pretty sure it doesn't need anything

PR #807, opened 2024-09-16
Make PDFMiner Pipelined

karanataryn:
Removes the sequenced partition_pdf method and makes the batched method pipelined.


lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -0,0 +1,108 @@
...
+        from pdfminer.converter import PDFPageAggregator
+        from pdfminer.layout import LAParams
+        from pdfminer.pdfinterp import PDFPageInterpreter, PDFResourceManager
+
+        rm = PDFResourceManager()
+        param = LAParams()
+        self.device = PDFPageAggregator(rm, laparams=param)
+        self.interpreter = PDFPageInterpreter(rm, self.device)
+
+    def _open_pdfminer_pages_generator(self, fp: BinaryIO):
```
bsowell, line 27: Isn't this kind of redundant with pdf_to_pages?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -578,10 +439,79 @@ def _partition_pdf_batched_named(
     def _run_pdfminer(pdf_path, hash_key, use_cache):
```
bsowell, line 439: Once this goes in, where does this get called? I guess from the managed-service before we update that?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -578,10 +439,79 @@ def _partition_pdf_batched_named(
...
+            pdfminer_pages = []
+            for _ in range(len(batch)):
+                pdf_page = pdf_generator.__next__()
+                pdfminer_pages.append(self.pdf_extractor.extract_page(pdf_page))
+
+            assert len(pdfminer_pages) == len(deformable_layout)
+            with LogTime("pdfminer_supplement"):
+                for d, p in zip(deformable_layout, pdfminer_pages):
+                    self._supplement_text(d, p)
+        # else pdfminer happens in parent since it is whole document.
```
bsowell, line 492: I think this comment no longer applies?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -578,10 +439,79 @@ def _partition_pdf_batched_named(
...
+        if extract_table_structure:
+            with LogTime("extract_table_structure_batch"):
+                if table_structure_extractor is None:
+                    table_structure_extractor = DEFAULT_TABLE_STRUCTURE_EXTRACTOR(device=self.device)
```
bsowell, line 496: minor, and we can maybe leave this to another cleanup PR, but we already initialize this in _extract_pdf_batched as well. Seems redundant to do it both places.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -0,0 +1,108 @@
...
+        pdf coordinates are different, bottom left is origin, also two diagonal points defining a rectangle is
+        (bottom left, upper right), for details, refer
+        https://www.leadtools.com/help/leadtools/v19/dh/to/pdf-topics-pdfcoordinatesystem.html
+        """
+        x1, y2, x2, y1 = rect
+        y1 = height - y1
+        y2 = height - y2
+        return x1, y1, x2, y2
+
+    def extract_document(
```
bsowell, line 51: Is the idea that this is legacy and will get removed when we don't have a dependency from the managed service? Otherwise, it seems like there is quite a bit of duplication. You could probably implement this as a loop over pdf_to_pages.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -0,0 +1,108 @@
...
+                            if text.text_representation:
+                                texts.append(text)
+
+                    pages.append(texts)
+                if use_cache:
+                    logger.info("Cache Miss for PDFMiner. Storing the result to the cache.")
+                    pdf_miner_cache.set(hash_key, pages)
+                return pages
+
+    def extract_page(self, page: Any, **kwargs) -> List[Element]:
```
bsowell, line 88: Why are you using Any as the type of a page here? If it's just for dependency stuff, you can do 

    if typing.TYPE_CHECKING:
         from pdfminer.layout import LTPage

and then put `"LTPage"` in quotes in the type signature.

bsowell: Also, why kwargs? Doesn't seem like you are using them.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -234,3 +235,24 @@ def read_stderr(fh, q):
 
         t_out.join()
         t_err.join()
+
+
+@requires_modules(["pdfminer", "pdfminer.utils"], extra="local-inference")
+def pdf_to_pages(file_name: str) -> Generator[Any, None, None]:
```
bsowell, line 241: Might suggest putting this in pdf_miner.py. Seems pretty tied to that as opposed to generally useful for PDFs.

lib/sycamore/sycamore/tests/unit/transforms/compare_detr_impls.py
```diff
@@ -1,22 +1,18 @@
 from sycamore.utils.cache import Cache
```
bsowell, line 1: This no longer compares impls, right? Should we rename?

karanataryn: Done

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: Yes, exactly. I want to be able to update Sycamore in managed-service without breaking it.

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: Yep, my bad, removed.

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: Makes sense, I've removed it

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
karanataryn: Good idea, have changed it. I initially added kwargs in case the user might want to pass arguments to PDFMiner, but think removing would be better for now.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
karanataryn: Removed.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
karanataryn: Yep, exactly. Kept it here and will remove once the managed service has moved.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -0,0 +1,108 @@
...
+        pdf coordinates are different, bottom left is origin, also two diagonal points defining a rectangle is
+        (bottom left, upper right), for details, refer
+        https://www.leadtools.com/help/leadtools/v19/dh/to/pdf-topics-pdfcoordinatesystem.html
+        """
+        x1, y2, x2, y1 = rect
+        y1 = height - y1
+        y2 = height - y2
+        return x1, y1, x2, y2
+
+    def extract_document(self, filename: str, hash_key: str, use_cache=False, **kwargs) -> List[List[Element]]:
```
bsowell, line 64: Can you add a TODO to clean this up once we have the implementation in the service?

bsowell (commented):
In theory I like this, but how do we convince ourselves that is not going to regress performance in mysterious ways? I know part of that is in the managed service, but maybe it makes sense to run the partitioning from sycamore on a big pdf before and after this change and see the difference?

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
karanataryn: Done.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -0,0 +1,109 @@
...
+from pathlib import Path
+from sycamore.utils.import_utils import requires_modules
+import logging
+
+if TYPE_CHECKING:
+    from pdfminer.layout import LTPage
+
+logger = logging.getLogger(__name__)
+
+pdf_miner_cache = DiskCache(str(Path.home() / ".sycamore/PDFMinerCache"))
```
bsowell, line 13: Add a todo to fix the cache.

bsowell (approved):

PR #811, opened 2024-09-17
Transform to convert Document into Markdown.

alexaryn:


lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+                s += f"- {text}\n"
+        elif type in ("caption", "footnote"):
+            s += f"\n{text}\n\n"
+        else:
+            s += text + "\n"
+    return s
+
+
+def render_table(elem: TableElement) -> str:
+    table = elem.table
```
HenryL27, line 64: there is `table.to_pandas().to_markdown()` you know

lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+    matrix = [[""] * ncol for _ in range(nrow)]
+    hdr_max = -1
+    for cell in cells:
+        hdr = cell.is_header
+        for row in cell.rows:
+            if hdr:
+                hdr_max = max(hdr_max, row)
+            for col in cell.cols:
+                if cell.content:
+                    matrix[row][col] = cell.content
```
HenryL27, line 79: This looks like it'll handle spanning cells weird. I guess how does markdown do spanning cells?

lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+    bullet = ""
+    for pos in range(0, lng):
+        ch = ary[0][pos]
+        same = True
+        for idx in range(1, num):
+            if ch != ary[idx][pos]:
+                same = False
+                break
+        if (not same) or ch.isalnum():
+            break
```
HenryL27, line 128: What about lists like 
1. foo
2. bar

lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+            s += f"\n## {text}\n\n"
+        elif type == "list-item":
+            tup = elem.data.get("_listctx")
+            if tup:
+                indent, bullet = tup
+                n = len(indent) + len(bullet)
+                t = text[n:]
+                if not t[0].isspace():
+                    t = " " + t
+                s += f"{indent}-{t}\n"
```
HenryL27, line 53: not `f"{indent}{bullet or '-'}{t}\n"`?

lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+SKIP_TYPES = {"page-header", "page-footer", "image"}
+
+
+def elements_to_markdown(elems: list[Element]) -> str:
+    """
+    This is the main function of interest.
+    Assumes elements are sorted as per bbox_sort.bbox_sort_document().
+    """
+    label_lists(elems)
+    s = ""
```
HenryL27, line 23: eric would probably point out that building a string like this is n^2 - and it might actually make a difference here if I try to turn a textbook into markdown

HenryL27 (commented):
mostly lgtm

examples/markdown.py
```diff
@@ -0,0 +1,16 @@
...
+    sycamore.init()
+    .read.binary(sys.argv[1:], binary_format="pdf")
+    .partition(partitioner=ArynPartitioner(extract_table_structure=True,
+                                           use_partitioning_service=False))
+    .markdown()
+    .explode()
+    .take()
+)
+
+print(docs[1].text_representation)  # doc zero isn't real
```
bsowell, line 16: I'm confused. I don't understand what the explode does here. If I don't do the explode, is the document text_representation set to the markdown?

lib/sycamore/sycamore/transforms/markdown.py
```diff
@@ -0,0 +1,53 @@
...
+        if pn is not None:
+            pageset.add(pn)
+    pages = sorted(pageset)
+    if not pages:
+        pages.append(1)
+    doc.elements = [
+        Element(
+            {
+                "type": "Text",
+                "bbox": (0.0, 0.0, 1.0, 1.0),  # best guess
```
bsowell, line 43: Why this as opposed to None?

lib/sycamore/sycamore/utils/markdown.py
bsowell: Yes, you are better off using StringIO, I would think.

lib/sycamore/sycamore/utils/markdown.py
bsowell: Yeah, can you run the table test cases we have through this and ensure you get something sensible? That PR I put out for markdown tables has the tests, though you might need to update the string comparison to remove whitespace.

lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+            else:
+                s += f"- {text}\n"
+        elif type in ("caption", "footnote"):
+            s += f"\n{text}\n\n"
+        else:
+            s += text + "\n"
+    return s
+
+
+def render_table(elem: TableElement) -> str:
```
bsowell, line 63: I would love to have a to_markdown method in the table object (which can call this). I feel like for tables specifically that is where I would look, since that's where to_html/to_csv/etc. are.

lib/sycamore/sycamore/transforms/markdown.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def make_markdown(doc: Document) -> Document:
+    elems = doc.elements
+    text = elements_to_markdown(elems)
+    pageset: set[int] = set()
+    for elem in elems:
+        pn = elem.properties.get("page_number")
+        if pn is not None:
+            pageset.add(pn)
```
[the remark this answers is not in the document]
alexaryn: Raises an exception if there's no None in the set.  Can be handled, but the original code works, too.

lib/sycamore/sycamore/transforms/markdown.py
```diff
@@ -0,0 +1,53 @@
...
+    pages = sorted(pageset)
+    if not pages:
+        pages.append(1)
+    doc.elements = [
```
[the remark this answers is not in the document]
alexaryn: I don't believe Markdown has inline-image functionality.

lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -0,0 +1,152 @@
...
+def elements_to_markdown(elems: list[Element]) -> str:
+    """
+    This is the main function of interest.
+    Assumes elements are sorted as per bbox_sort.bbox_sort_document().
+    """
+    label_lists(elems)
+    s = ""
+    last = [-1, -1, -1, -1]
+    for elem in elems:
+        type = elem_type(elem).lower()
```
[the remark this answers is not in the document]
alexaryn: How does one differentiate them.  If you say "by case", I'm going to ask for a more robust and future-proof answer.  Maybe I need `isinstance()`?

lib/sycamore/sycamore/utils/markdown.py
alexaryn: There are enough weird edge cases in Markdown that I'd rather have the logic and control here in our code.

lib/sycamore/sycamore/utils/markdown.py
alexaryn: Markdown has no official support for spanning cells, but there are many variants of MD.

lib/sycamore/sycamore/utils/markdown.py
alexaryn: We don't currently remove common prefixes from ordered lists.  Basically, these render with bullets in front of the numbers.  Similar things happen with hierarchical lists.  They actually don't look too bad.  Not a lot of great options for hierarchical lists.  We can add recognition/rendering for simple numerical lists, but I'd rather do this later.  I'd expect it to be a bit complicated and also to be very susceptible to OCR and partitioning errors.

lib/sycamore/sycamore/utils/markdown.py
alexaryn: Markdown requires the dash, not arbitrary bullet characters.  In make cases, especially with OCR, you get "bullets" that are like `e` or `=` or some Unicode thing.  Also, I prefer to put only dead-simple things inside the curlies of f-strings.

lib/sycamore/sycamore/utils/markdown.py
alexaryn: I forgot that Python doesn't have growable strings.  I can tweak this.  I have already run the 582-page SLO budget through without obvious issues, but have not tried the calculus text yet.

examples/markdown.py
alexaryn: If you don't explode, the text_representation is in the Element.  I did it this way because I didn't think converting to Markdown should also do an implicit explode().  I think people are used to seeing explode in pipelines.  It would be easy to change, if needed.

lib/sycamore/sycamore/transforms/markdown.py
alexaryn: It seems there's a fair amount of code that expects to see bounding boxes.  I could leave it unset.  Not sure if there would be any consequences, good or bad.

lib/sycamore/sycamore/utils/markdown.py
alexaryn: I was actually surprised to see all those methods there.  I felt like the other classes (Document, Element) were more about data representation and I liked the separation from parsing and rendering, which can become kitchen sinks of ad hoc methods.

lib/sycamore/sycamore/utils/markdown.py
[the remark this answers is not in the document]
HenryL27: This is what we do with textract tables:
```python
            if table.title:
                element.text_representation = table.title.text + "\n"
            else:
                element.text_representation = ""

            element.text_representation = element.text_representation + table.to_csv() + "\n"
            element.bbox = BoundingBox(*bbox_to_coord(table.bbox))

            for footer in table.footers:
                element.text_representation = element.text_representation + footer.text + "\n"
```
sycamore/transforms/extract_table.py:L88

HenryL27 (approved):
I'm happy if you're happy

PR #813, opened 2024-09-17
Add QueryVectorDatabase to SycamoreQuery

baitsguy:
1. Some bug fixes regarding default vars in context_params
2. Integ test fix for SycamoreQuery missing `natural_language_response=True`
3. Added QueryVectorDatabase operator

Will continue prompt tuning as we play with it. Will also want to port this as a native DocSetReader function at some point.

### DIFF
--- a/lib/sycamore/sycamore/connectors/opensearch/utils.py
+++ b/lib/sycamore/sycamore/connectors/opensearch/utils.py
@@ -7,6 +7,10 @@ from sycamore.transforms import Embedder
 
 @context_params("opensearch")
 def get_knn_query(text_embedder: Embedder, query_phrase: str, k: int = 500, context: Optional[Context] = None):
+    """
+    Given a query string and an Embedder, create a simple OpenSearch Knn query. Uses a default k value of 500.
+    This is only the base query, if you need to add filters or other specifics you can extend the object.
+    """
     embedding = text_embedder.generate_text_embedding(query_phrase)
     query = {
         "query": {
--- a/lib/sycamore/sycamore/connectors/opensearch/utils.py
+++ b/lib/sycamore/sycamore/connectors/opensearch/utils.py
@@ -1,0 +1,21 @@
+from typing import Optional
+
+from sycamore import Context
+from sycamore.context import context_params
+from sycamore.transforms import Embedder
+
+
+@context_params("opensearch")
+def get_knn_query(text_embedder: Embedder, query_phrase: str, k: int = 500, context: Optional[Context] = None):
+    embedding = text_embedder.generate_text_embedding(query_phrase)
+    query = {
+        "query": {
+            "knn": {
+                "embedding": {
+                    "vector": embedding,
+                    "k": k,
+                }
+            }
+        }
+    }
+    return query
--- a/lib/sycamore/sycamore/query/operators/query_database.py
+++ b/lib/sycamore/sycamore/query/operators/query_database.py
@@ -93,6 +93,3 @@ class QueryVectorDatabase(LogicalOperator):
 
     The full range of OpenSearch Query DSL parameters for a filter query are supported.
     """
-
-
-
--- a/lib/sycamore/sycamore/query/operators/query_database.py
+++ b/lib/sycamore/sycamore/query/operators/query_database.py
@@ -45,32 +45,32 @@ class QueryDatabase(LogicalOperator):
 
 
 class QueryVectorDatabase(LogicalOperator):
-    """OpenSearch knn query to load records that are similar to a given query string. Returns the top k records
-    according to the cosine similarity between the query string and record's text content.
-    Use this if we need to filter on a field that can't suffice with an exact match, but also isn't complex enough to
+    """Vector search to load records that are similar to a given query string. Returns the top k records
+    according to the vector similarity between the query string and record's text content.
+    Use this if you need to filter on a field that can't suffice with an exact match, but also isn't complex enough to
     require a llm_filter, i.e. a record level LLM call.
 
     Here are some instructions about using QueryVectorDatabase instead of QueryDatabase:
     1. Use QueryVectorDatabase if the query plan uses an LLM operator (LLMFilter or LLMExtractEntity),
         and the final answer doesn't require you to scan the full dataset (e.g. find me an instance of a transcript
-        that talked about AI)
+        that talked about AI).
     2. Use QueryVectorDatabase for RAG-style questions where you need to retrieve a set of candidate records where the
         answer might exist.
-    3. Do not use QueryVectorDatabase where examining each record is essential (e.g. did all transcripts talk about AI?)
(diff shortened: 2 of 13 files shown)


lib/sycamore/sycamore/connectors/opensearch/utils.py
```diff
@@ -0,0 +1,21 @@
+from typing import Optional
+
+from sycamore import Context
+from sycamore.context import context_params
+from sycamore.transforms import Embedder
+
+
+@context_params("opensearch")
+def get_knn_query(text_embedder: Embedder, query_phrase: str, k: int = 500, context: Optional[Context] = None):
```
mdwelsh, line 9: Docstring please :-)

lib/sycamore/sycamore/connectors/opensearch/utils.py
```diff
@@ -0,0 +1,21 @@
...
+    embedding = text_embedder.generate_text_embedding(query_phrase)
+    query = {
+        "query": {
+            "knn": {
```
mdwelsh, line 13: I've been using the "neural" feature:
https://github.com/aryn-ai/sycamore/blob/01f42f124b972211474882d0bf3392f134e799b3/apps/query-demo/querydemo/Query_Demo.py#L447

How is this difference? Should we settle on one way to do this?

lib/sycamore/sycamore/query/operators/query_database.py
```diff
@@ -42,3 +42,54 @@ class QueryDatabase(LogicalOperator):
     Whenever possible, use the query parameter to filter data at the source, as this is more
     efficient than filtering data in subsequent data filtering operators.
     """
+
+
+class QueryVectorDatabase(LogicalOperator):
+    """OpenSearch knn query to load records that are similar to a given query string. Returns the top k records
+    according to the cosine similarity between the query string and record's text content.
+    Use this if we need to filter on a field that can't suffice with an exact match, but also isn't complex enough to
```
mdwelsh, line 50: Since these docstrings end up in the LLM prompt, use "you" instead of "we".

lib/sycamore/sycamore/query/operators/query_database.py
```diff
@@ -42,3 +42,54 @@ class QueryDatabase(LogicalOperator):
     Whenever possible, use the query parameter to filter data at the source, as this is more
     efficient than filtering data in subsequent data filtering operators.
     """
+
+
+class QueryVectorDatabase(LogicalOperator):
+    """OpenSearch knn query to load records that are similar to a given query string. Returns the top k records
```
mdwelsh, line 48: I think it's best if the planner does not need to know about OpenSearch or implementation details. That is, from the planner's perspective, this is just a "SimilaritySearch" operation --maybe you don't even want to call it "QueryVectorDatabase". I would not say things about "knn" or "cosine similarity" since the planner need not understand those details.

lib/sycamore/sycamore/query/operators/query_database.py
```diff
@@ -42,3 +42,54 @@ class QueryDatabase(LogicalOperator):
...
+    Use this if we need to filter on a field that can't suffice with an exact match, but also isn't complex enough to
+    require a llm_filter, i.e. a record level LLM call.
+
+    Here are some instructions about using QueryVectorDatabase instead of QueryDatabase:
+    1. Use QueryVectorDatabase if the query plan uses an LLM operator (LLMFilter or LLMExtractEntity),
+        and the final answer doesn't require you to scan the full dataset (e.g. find me an instance of a transcript
+        that talked about AI)
+    2. Use QueryVectorDatabase for RAG-style questions where you need to retrieve a set of candidate records where the
+        answer might exist.
+    3. Do not use QueryVectorDatabase where examining each record is essential (e.g. did all transcripts talk about AI?)
```
mdwelsh, line 59: Can you make sure each of these list items ends in a period? This is SUPER anal, I know, but I have a superstitious belief that grammatical queries do a better job than ones with missing punctuation, etc. (This is a completely unfounded belief, but it seems to me that being grammatical is always better when it comes to prompting LLMs.)

lib/sycamore/sycamore/query/operators/query_database.py
```diff
@@ -42,3 +42,54 @@ class QueryDatabase(LogicalOperator):
...
+    4. Do not use QueryVectorDatabase where you need to potentially return large datasets (e.g. give me all transcripts
+        that talked about AI)
+
+    """
+
+    index: str
+    """The index to load data from."""
+
+    query_to_embed: str
+    """The string to convert to a vector and perform vector search"""
```
mdwelsh, line 69: Maybe call this "query_term" or something like that.

lib/sycamore/sycamore/query/operators/query_database.py
```diff
@@ -42,3 +42,54 @@ class QueryDatabase(LogicalOperator):
...
+
+    filter: Optional[Dict] = None
+    """
+    A filter query in OpenSearch Query DSL format which represents a filter in a knn query.
```
mdwelsh, line 73: Same as above, I think the wording here should probably not mention "knn query"

mdwelsh (approved):
Nice! Some suggestions about docstrings and LLM prompts.

lib/sycamore/sycamore/connectors/opensearch/utils.py
baitsguy: Neural search expects you to have an embedding model setup in your opensearch cluster. So the input to a neural search is a string, which opensearch will then create a vector of using that model, and then perform knn.

Knn doesn't use opensearch to create the query embedding, but instead accepts a vector as an input. So to use this we need to embed the query string outside of the opensearch call.

Knn is simpler for us since it requires less configuration on the opensearch instance, and access to the embedding model can be tricky to do in opensearch. From what I see, most vector dbs expect the latter model, so this approach unifies the logic in that sense as well.

I'd prefer knn if we think the UI will have enough juice to run an embedding model (almost certainly yes at this point)

lib/sycamore/sycamore/query/operators/query_database.py
baitsguy: I'll remove knn/cosine but need opensearch in there since we're still using it to generate filters in opensearch dsl

baitsguy: Or will remove as much of it as I can while maintaining filter support

lib/sycamore/sycamore/query/operators/query_database.py
baitsguy: Haha sure

lib/sycamore/sycamore/query/operators/query_database.py
baitsguy: will change to query_phrase

lib/sycamore/sycamore/query/operators/query_database.py
baitsguy: I updated this param to say 'opensearch_filter', and left opensearch specific wording in here

PR #815, opened 2024-09-17
Ritam add example notebook

RitxmSaha:


examples/simple_neo4j.py
```diff
@@ -0,0 +1,67 @@
...
+    is_student: bool
+    total_flight_hours: Optional[int]
+
+
+class INVOLVED_AIRCRAFT(BaseModel):
+    start: Report
+    end: Aircraft
+
+
+class FLOWN_BY(BaseModel):
```
HenryL27, line 38: why the UPPERCASE NAMES?

examples/simple_neo4j.py
```diff
@@ -0,0 +1,67 @@
...
+class Report(BaseModel):
+    ID: str = Field(
+        description="This ID can be found under either Accident Number, Incident Number, or Occurance Number"
+    )
+
+
+class Aircraft(BaseModel):
+    registration: str
+    make: str
+    model: str
```
HenryL27, line 25: should all these guys have descriptions too?

examples/simple_neo4j.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class Pilot(BaseModel):
+    is_student: bool
+    total_flight_hours: Optional[int]
+
+
+class INVOLVED_AIRCRAFT(BaseModel):
+    start: Report
+    end: Aircraft
```
HenryL27, line 35: is the start, end mandatory? Can I stick other fields on it?

HenryL27 (approved):

examples/simple_neo4j.py
RitxmSaha: start and end are mandatory. You can stick other fields on it as well and they are saved as properties!

examples/simple_neo4j.py
RitxmSaha: They don't have to, but they can. It just helps the LLM generate context for what to extract, usually I only add them for stuff that is unclear(i.e. the report id since it could be found under several different fields)

examples/simple_neo4j.py
RitxmSaha: Default way neo4j does it, not required but thought it would just be consistent

PR #816, opened 2024-09-17
Improved NTSB ingestion pipeline for Luna demo.

mdwelsh:
This PR improves the loading script for the Luna NTSB demo. I've added the script to the `query-demo` directory since I want to ensure that the demo UI and loader code remain consistent.

In addition to changes to the loader script, this PR has a few improvements to bits and pieces of Sycamore that it depends on:

* Fix prompts for ExtractTableProperties
* Ensure that AssignDocProperties does not overwrite 'entity' (use .update instead)
* Rename LocationStandardizer -> USStateStandardizer to properly convey what it is doing
* Update DateTimeStandardizer to support different formatting options, and more robust date parsing using the `dateparse` library.
* A few other minor tweaks.


lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -16,8 +15,8 @@ class Standardizer(ABC):
     transforming specific fields within a document according to certain rules.
     """
 
-    @abstractmethod
-    def fixer(self, text: str) -> Union[str, Tuple[str, date]]:
+    @staticmethod
```
bsowell, line 18: Does this mean that all standardizer implementations have to be static? I guess that's true now, but not sure it's always true. That said, I guess this abstraction doesn't provide us all that much anyway, so whatever.

bsowell (approved):

apps/query-demo/querydemo/loaddata.py
```diff
@@ -0,0 +1,161 @@
...
+                doc, element_type="table", property_name="table_props"
+            )
+        )
+        .extract_properties(property_extractor=OpenAIPropertyExtractor(llm=llm, num_of_elements=35))
+        .merge(GreedyTextElementMerger(tokenizer, 300))
+        .map(lambda doc: USStateStandardizer.standardize(doc, key_path=["properties", "entity", "location"]))
+        .map(lambda doc: DateTimeStandardizer.standardize(doc, key_path=["properties", "entity", "dateAndTime"]))
+        .spread_properties(["entity", "path"])
+        .explode()
+        .sketch()
```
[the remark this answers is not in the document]
mdwelsh: Yes, but for now it's harmless so keeping it in.

lib/sycamore/sycamore/transforms/standardizer.py
mdwelsh: Yes, for now. I could not figure out how to get mypy to be happy with the method being non-static in the base class but static in the subclass.

PR #825, opened 2024-09-18
Fix poetry.lock with missing dependency.

mdwelsh:


baitsguy (approved):
Tests are running now so this fixed the issue

karanataryn (approved):

PR #826, opened 2024-09-19
Get local mode working 1/n

eric-anderson:
Working for a few unit tests (file writer), examples (html_ingest, ndd_debug), and notebooks (default-prep-script).
* Update the file writers to work in local mode
* Refactor the code for the database writers to remove duplicate lines.
* Refactor infer_fs so we can use it in multiple places and a variant that will cross check that the user specified fs and the inferred one are the same, in which case we use the inferred path to be more similar to ray.
* Add constants for exec mode to sycamore namespace. Remove materialize constants from default imports.
* Add materialize in a bunch of places to speed up execution.
* Discover we have a strong ray dependency for specifing parallelism. Add TODO to fix this problem.

### DIFF
--- a/lib/sycamore/sycamore/__init__.py
+++ b/lib/sycamore/sycamore/__init__.py
@@ -3,6 +3,8 @@ from sycamore.docset import DocSet
 from sycamore.executor import Execution
 from sycamore.materialize_config import MaterializeSourceMode
 
+EXEC_RAY = ExecMode.RAY
+EXEC_LOCAL = ExecMode.LOCAL
 MATERIALIZE_RECOMPUTE = MaterializeSourceMode.RECOMPUTE
 MATERIALIZE_USE_STORED = MaterializeSourceMode.USE_STORED
 
@@ -14,6 +16,4 @@ __all__ = [
     "Execution",
     "ExecMode",
     "MaterializeSourceMode",
-    "MATERIALIZE_RECOMPUTE",
-    "MATERIALIZE_USE_STORED",
 ]
--- a/lib/sycamore/sycamore/materialize.py
+++ b/lib/sycamore/sycamore/materialize.py
@@ -225,9 +225,9 @@ class Materialize(UnaryNode):
 
     @staticmethod
     def infer_fs(path: str) -> Tuple["pyarrow.FileSystem", Path]:
-        from sycamore.utils.pyarrow import infer_fs
+        from sycamore.utils.pyarrow import infer_fs as util_infer_fs
 
-        (fs, path) = infer_fs(path)
+        (fs, path) = util_infer_fs(path)
         return (fs, Path(path))
 
     def save(self, doc: Document) -> None:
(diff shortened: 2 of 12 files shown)


lib/sycamore/sycamore/writer.py
```diff
@@ -742,6 +711,13 @@ def json(
...
+
+        self._maybe_execute(node, True)
+
+    def _maybe_execute(self, node: Node, execute: bool) -> Optional[DocSet]:
+        ds = DocSet(self.context, node)
+        if not execute:
+            return ds
+
+        ds.execute()
+        return None
```
HenryL27, line 723: nice

notebooks/default-prep-script.ipynb
```diff
@@ -312,7 +312,16 @@
...
    "execution_count": null,
    "id": "8",
-   "metadata": {},
+   "metadata": {
+    "nbmake": {
+     "post_cell_execute": [
+      "if sycamore.EXEC_LOCAL != sycamore.ExecMode.RAY:",
+      "    import sys",
+      "    # TODO: eric - remove the True",
+      "    assert True or 'ray' not in sys.modules"
```
HenryL27, line 321: what does this do?

HenryL27 (approved):
lgtm. A little confused on what dropping those constants from sycamore __ all__ did.

lib/sycamore/sycamore/__init__.py
```diff
@@ -14,6 +16,4 @@
     "Execution",
     "ExecMode",
     "MaterializeSourceMode",
-    "MATERIALIZE_RECOMPUTE",
-    "MATERIALIZE_USE_STORED",
```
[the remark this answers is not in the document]
eric-anderson: It works fine. __all__ affects what happens if you do import sycamore.*

lib/sycamore/sycamore/materialize.py
```diff
@@ -224,22 +224,11 @@ def finalize(self):
...
-            # normalize everything to a URI.
-            p = Path(path)
-            if p.is_absolute():
-                path = p.as_uri()
-            else:
-                path = p.absolute().as_uri()
-
-        from pyarrow import fs
+    def infer_fs(path: str) -> Tuple["pyarrow.FileSystem", Path]:
+        from sycamore.utils.pyarrow import infer_fs
```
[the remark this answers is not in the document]
eric-anderson: done.

notebooks/default-prep-script.ipynb
eric-anderson: Lets you add "tests" to notebooks. If we're running in local mode, then ray should not be in sys.modules.
I was planning on having the test script run the notebooks and then replace EXEC_LOCAL with EXEC_RAY and then running them a second time to make sure the notebooks work in both modes.  However, the check I wanted to add doesn't work yet.

PR #832, opened 2024-09-20
Sundry improvements while using luna in a customer.

eric-anderson:
setup.sh: Remove the , that was added. We need the , to get multiple simultaneous installs in a single bunch, and need to remove it for the script.

.force_directory_creation: Empty file to make sure the cache_dir exists to allow running the container

configuration.py: Change the way we hook a custom client into the query ui -- the existing approach
  was fragile. If the API for specifying the custom client changes, we want people to get a git
  conflict when they try to merge because they probably have to adjust how they configure the
  client. By always running the code we can be sure that it parses, it was very easy to have a bug
  in parsing cause it to not work. For example, the sample file had the wrong function name. Remove
  some parameters that are not useful, a custom client can't take an llm because it would not be able
  to replace the llm with whatever customization is needed. Similarly os_client_args don't make sense
  because the customization may have to specify those values. Luckily nothing used those parameters.
  Add an example of how to use the configuration to override things -- we should eventually add for
  overriding, but it probably makes sense to only do that for the query-demo UI since query-ui is
  mostly obsolete now.

test_configuration.py: Make sure that if people are changing stuff they intend it.

simple_ntsb.py: plan.show() no longer works, remove it.

client.py: Add documentation on the rules for overriding parameters, and have the assertions
  point people at that documentation. I had to read the code to understand the existing assertion messages.
  I think the text_embedder is also needed if you specify opensearch so document that as well.
  Fix bug where not specifying os_client_args in the context, but specifying the llm didn't work.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -73,6 +73,9 @@ jobs:
       - name: Run more tests
         run: poetry run python sycamore/tests/manual/test_fast_sycamore_import.py
         working-directory: lib/sycamore
+      - name: Run query-ui test
+        run: poetry run bash -c 'cd ../../apps/query-ui && PYTHONPATH=. pytest .'
+        working-directory: lib/sycamore
       - name: DF-6
         run: df
 
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -74,7 +74,8 @@ jobs:
         run: poetry run python sycamore/tests/manual/test_fast_sycamore_import.py
         working-directory: lib/sycamore
       - name: Run query-ui test
-        run: poetry run bash -c 'cd ../../apps/query-ui && PYTHONPATH=. pytest .'
+        run: poetry run bash -c 'cd ../../apps/query-ui/queryui && PYTHONPATH=. pytest .'
+        # run here because that's where the poetry venv is set up.
         working-directory: lib/sycamore
       - name: DF-6
         run: df
--- a/apps/query-ui/queryui/test_configuration.py
+++ b/apps/query-ui/queryui/test_configuration.py
@@ -7,4 +7,4 @@ def test_configuration_sha():
         sha = sha256(bytes).hexdigest()
         # If the change was intentional, update the hash
         # Think about whether you have to make the change since everyone with a custom config will need to update it
-        assert sha == "c8805d3e9a641a6e29a947e179a8eb7441567ecbbc238bc7a39c8ea2e2c811cf", f"hash mismatch got {sha}"
+        assert sha == "1c12bf1d3437a494d24910bf2073b7461ac6dfa4c37c25d9929878fc35be9d40", f"hash mismatch got {sha}"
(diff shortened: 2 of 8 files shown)


.github/workflows/testing.yml
```diff
@@ -73,6 +73,9 @@ jobs:
       - name: Run more tests
         run: poetry run python sycamore/tests/manual/test_fast_sycamore_import.py
         working-directory: lib/sycamore
+      - name: Run query-ui test
+        run: poetry run bash -c 'cd ../../apps/query-ui && PYTHONPATH=. pytest .'
+        working-directory: lib/sycamore
```
mdwelsh, line 78: Why not just set `working-directory` to `apps/query-ui` and then do `poetry run pytest .` ?

apps/query-ui/queryui/test_configuration.py
```diff
@@ -0,0 +1,10 @@
+def test_configuration_sha():
+    from hashlib import sha256
+    import configuration
+
+    with open(configuration.__file__, "rb") as f:
+        bytes = f.read()
+        sha = sha256(bytes).hexdigest()
+        # If the change was intentional, update the hash
+        # Think about whether you have to make the change since everyone with a custom config will need to update it
+        assert sha == "c8805d3e9a641a6e29a947e179a8eb7441567ecbbc238bc7a39c8ea2e2c811cf", f"hash mismatch got {sha}"
```
mdwelsh, line 10: This seems kind of odd to me. Why is this API so brittle that we need an explicit test to prevent people from changing this file? Or is this just to prevent accidental modifications to the file from being checked in? If the latter, we should not instruct people to be modifying that file to customize the client; they should use some other mechanism, like creating their own client file.

lib/sycamore/sycamore/query/client.py
```diff
@@ -100,6 +100,18 @@ class SycamoreQueryClient:
         os_config (optional): OpenSearch configuration. Defaults to DEFAULT_OS_CONFIG.
         os_client_args (optional): OpenSearch client arguments. Defaults to DEFAULT_OS_CLIENT_ARGS.
         trace_dir (optional): Directory to write query execution trace.
+
+    Notes:
+        If you override the context. You cannot override the s3_cache_path or os_client_args; you need
```
mdwelsh, line 105:
```suggestion
        If you override the context, you cannot override the s3_cache_path or os_client_args; you need
```

mdwelsh, line 105: This seems a little confusing. Maybe we should just get rid of these parameters and require a custom context being passed in if any of these parameters needs to be modified.

.github/workflows/testing.yml
eric-anderson: Because the poetry environment is in lib/sycamore.  Look at the install-sycamore step above.

apps/query-ui/queryui/test_configuration.py
eric-anderson: Lets meet and discuss.

lib/sycamore/sycamore/query/client.py
eric-anderson: I agree, but I didn't want to break existing code.

eric-anderson: Done.

mdwelsh (approved):

PR #836, opened 2024-09-21
Some improvements to query plans for Luna demo.

mdwelsh:
This PR makes it possible for a Luna app to provide its own few-shot examples for query plan generation, based on the
specific knowledge of the data and types of queries likely to be issued.

It also fixes a small bug in docset.llm_cluster_entity where a doc that does not have the field used to perform clustering would cause this operation to crash.

### DIFF
--- a/apps/query-demo/querydemo/Query_Demo.py
+++ b/apps/query-demo/querydemo/Query_Demo.py
@@ -23,7 +23,6 @@ NUM_DOCS_GENERATE = 60
 NUM_DOCS_PREVIEW = 10
 NUM_TEXT_CHARS_GENERATE = 2500
 
-# Set OpenAI API key from Streamlit secrets
 openai_client = OpenAI(api_key=os.environ["OPENAI_API_KEY"])
 
 OPENSEARCH_INDEX = "const_ntsb"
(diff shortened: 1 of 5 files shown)


apps/query-demo/querydemo/Query_Demo.py
```diff
@@ -491,12 +491,22 @@ def show(self):
         show_query_traces(st.session_state.trace_dir, self.query_id)
 
 
+@st.dialog("Send Feedback", width="large")
+def send_feedback():
+    feedback_form_code = (
```
baitsguy, line 496: is there a way to pass in context here? i.e. questions asked etc etc

lib/sycamore/sycamore/query/planner.py
```diff
@@ -99,7 +402,7 @@ def __init__(
         os_client: "OpenSearch",
         operators: Optional[List[Type[LogicalOperator]]] = None,
         llm_client: Optional[LLM] = None,
-        use_examples: bool = True,
+        examples: Optional[str] = None,
```
baitsguy, line 405: exciting

lib/sycamore/sycamore/query/planner.py
```diff
@@ -74,6 +74,307 @@
     "8. The last step of each plan *MUST* be a **SummarizeData** operation that returns a natural language response."
 )
 
+PLANNER_EXAMPLES = """
```
baitsguy, line 77: can you convert this to a list? the immediate use case is using different formatting based on llm. I want xml <example> for claude

baitsguy (approved):
Have a request to make examples a list of examples but not blocking

PR #837, opened 2024-09-23
Update requires_modules type annotations to work with mypy.

bsowell:
Decorators require some fancy generic annotations to play nicely with mypy and type checking. This commit adds the annotations and verifies that the types are detected correctly.


lib/sycamore/pyproject.toml
```diff
@@ -80,6 +80,7 @@ transformers = { version = "^4.43.1", optional = true }
 # Legacy partitioner dependencies
 unstructured = { version = "0.10.20", optional = true }
 python-pptx = {version = "^0.6.22", optional = true }
+typing-extensions = "^4.12.2"
```
HenryL27, line 83: not sure we care overly much about pyproject.toml organization but this is in legacy partitioner dependencies segment

HenryL27 (approved):
cool!

lib/sycamore/pyproject.toml
bsowell: Good call. Will move.

PR #839, opened 2024-09-24
Lazily Set Table Text Representation

karanataryn:
To help fix memory issues.


lib/sycamore/sycamore/data/element.py
```diff
@@ -159,8 +159,7 @@ def __init__(
         self.data["properties"]["rows"] = rows
         self.data["table"] = table
         self.data["tokens"] = tokens
-        if self.data["table"]:
-            self.data["text_representation"] = self.data["table"].to_csv()
+        self.data["text_representation"] = None
```
alexaryn: Leaving this unset is fine, too, and matches the previous code if table is not provided.  If we use self.data.get("text_representation") everywhere, there's no difference.  Personally, I like to err on the side of less code, but there's also an argument for "full initialization".

lib/sycamore/sycamore/data/element.py
```diff
@@ -191,10 +190,7 @@ def table(self) -> Optional[Table]:
     @table.setter
     def table(self, value: Optional[Table]) -> None:
         self.data["table"] = value
-        if value is not None:
-            self.data["text_representation"] = self.data["table"].to_csv()
-        else:
-            self.data["text_representation"] = None
+        self.data["text_representation"] = None
```
alexaryn, line 192: Maybe a comment "invalidate cache"?

Could also `del self.data["text_representation"]` which may more clearly express the intent.

lib/sycamore/sycamore/data/element.py
```diff
@@ -204,6 +200,16 @@ def tokens(self) -> Optional[list[dict[str, Any]]]:
     def tokens(self, tokens: list[dict[str, Any]]) -> None:
         self.data["tokens"] = tokens
 
+    @property
+    def text_representation(self) -> Optional[str]:
+        if not self.data["text_representation"] and self.data["table"]:
+            self.data["text_representation"] = self.data["table"].to_csv()
+        return self.data["text_representation"]
```
alexaryn, line 209: I'd prefer to see the code more like the following.  I think the logic is clearer and there are fewer hash lookups.  It handles missing keys better.  Also, the walrus operator.
```
tr = self.data.get("text_representation")
if not isinstance(tr, str):
    if (tbl := self.data.get("table")):
        tr = tbl.to_csv()
	self.data["text_representation"] = tr
return tr
```

lib/sycamore/sycamore/data/element.py
```diff
@@ -204,6 +200,16 @@ def tokens(self) -> Optional[list[dict[str, Any]]]:
...
+
+    @text_representation.setter
+    def text_representation(self, text_representation: str) -> None:
+        self.data["text_representation"] = text_representation
```
alexaryn, line 213: Can we just inherit this from Element?

alexaryn (approved):
Code as written will work, but I think it can be improved a bit.

lib/sycamore/sycamore/data/element.py
karanataryn: Have removed it to match previous behavior

lib/sycamore/sycamore/data/element.py
karanataryn: We'll need to redefine a setter since we're overriding the getter method.

lib/sycamore/sycamore/data/element.py
karanataryn: Sure, I've included a version of this.

lib/sycamore/sycamore/data/element.py
karanataryn: I'd like to keep the assignment to avoid missing keys, added the comment.

lib/sycamore/sycamore/data/element.py
karanataryn: We'll need to override both to keep the property as read-write instead of read-only

PR #840, opened 2024-09-24
Add a simple logical query plan compare function

baitsguy:
Helps you get some sense of how 2 query plans compare, will let you do regression testings, planner comparisons. We'll probably iterate on the actual comparisons we do as we use this

Note: This doesn't validate functional equality. For e.g. 2 filter operations reversed in order might mean the same thing but this function will detect a difference.


lib/sycamore/sycamore/query/compare.py
```diff
@@ -0,0 +1,66 @@
+from typing import Any, Set
```
mdwelsh, line 1: Why not just have this in logical_plan.py? Seems like we don't need a separate source file for this.

lib/sycamore/sycamore/query/compare.py
```diff
@@ -0,0 +1,66 @@
...
+
+from sycamore.query.logical_plan import LogicalPlan, Node
+
+
+def compare_graphs(node_a: Node, node_b: Node, visited_a: Set[int], visited_b: Set[int]) -> dict[str, Any]:
```
mdwelsh, line 6: Could we just use https://dictdiffer.readthedocs.io/en/latest/ ?

lib/sycamore/sycamore/query/compare.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+    @param node_a: graph node a
+    @param node_b: graph node b
+    @param visited_a: helper to track traversal in graph a
+    @param visited_b: helper to track traversal in graph b
+    @return: Dict with comparison metrics in the format diff_type_name: [(node_a, node_b)...]
+    """
+    diff_result: dict[str, Any] = {
+        "node_type_diff_result": [],
```
mdwelsh, line 21: I don't love this pattern, maybe this should be a dataclass or a tuple.

lib/sycamore/sycamore/query/operators/llm_extract_entity.py
```diff
@@ -24,3 +24,5 @@ class LlmExtractEntity(LogicalOperator):
     discrete: bool = False
     """True if the new field has a known finite number of possible values (e.g. number, letter,
     continent, color). False otherwise (e.g., for any free text outputs)."""
+
+    _keys_to_exclude_for_comparison = {"question", "new_field"}
```
mdwelsh, line 28: Could you instead use a field decorator? Seems cleaner.

lib/sycamore/sycamore/query/planner.py
```diff
@@ -76,6 +76,59 @@
 )
 
 
+def process_json_plan(
```
mdwelsh, line 79: What's the advantage of lifting this out of the Planner class?

lib/sycamore/sycamore/tests/unit/query/test_compare.py
```diff
@@ -0,0 +1,112 @@
...
+    result_node, nodes = process_json_plan(plan)
+    plan = LogicalPlan(result_node=result_node, nodes=nodes, query="", llm_prompt="", llm_plan="")
+    return plan
+
+
+def test_compare_plans(llm_filter_plan, vector_search_filter_plan):
+    diff = compare_logical_plans_from_query_source(
+        get_logical_plan(llm_filter_plan), get_logical_plan(vector_search_filter_plan)
+    )
+    assert len(diff["node_type_diff_result"]) > 0
```
mdwelsh, line 83: I wish these tests were more precise; it would be easy for your code to be somewhat broken and these tests still pass.

lib/sycamore/sycamore/query/compare.py
baitsguy: Was trying to keep it separate since its an eval framework but I can move it

lib/sycamore/sycamore/query/compare.py
baitsguy: reading the link the PR seems to give a little more control on what we're comparing. I also prefer being able to access the class.compare() function

lib/sycamore/sycamore/query/compare.py
baitsguy: Ack

lib/sycamore/sycamore/query/operators/llm_extract_entity.py
baitsguy: Let me give it a shot

lib/sycamore/sycamore/query/planner.py
baitsguy: I'm using it as a util in the tests. Easier to create a plan and just have this convert it into the class

lib/sycamore/sycamore/tests/unit/query/test_compare.py
baitsguy: Do you mean in terms of asserting the specific diffs? I can do that

lib/sycamore/sycamore/query/operators/llm_extract_entity.py
baitsguy: updated, not super clean but less likely to mess up at least

mdwelsh (approved):

PR #842, opened 2024-09-25
Improve luna property handling

eric-anderson:
* Include all properties instead of the ones just under properties.entity Keys are named with the full path, so it's easy to tell which properties are entity properties.

* Handle properties with varying types, for int/float it takes the most general type. For other cases it takes the first and logs a message since there are likely to be problems.

### DIFF
--- a/lib/sycamore/sycamore/query/schema.py
+++ b/lib/sycamore/sycamore/query/schema.py
@@ -40,7 +40,6 @@ class OpenSearchSchemaFetcher:
         query["index"] = self._index
         query["query"] = {"query": {"match_all": {}}, "size": self.NUM_EXAMPLES}
         random_sample = self._query_executor.query(query)["result"]["hits"]["hits"]
-        # logger.debug(f"Random sample came back with {random_sample}")
         result: OpenSearchSchema = {}
         result["text_representation"] = ("<class 'str'>", {"Can be assumed to have all other details"})
 
--- a/lib/sycamore/sycamore/query/schema.py
+++ b/lib/sycamore/sycamore/query/schema.py
@@ -1,14 +1,18 @@
+import logging
 import typing
 from typing import Dict, Set, Tuple
 
 from sycamore.transforms.query import OpenSearchQueryExecutor
 from sycamore.data import OpenSearchQuery
+from sycamore.utils.nested import dotted_lookup
 
 if typing.TYPE_CHECKING:
     from opensearchpy.client.indices import IndicesClient
 
 OpenSearchSchema = Dict[str, Tuple[str, Set[str]]]
 """Represents a mapping from field name to field type and a set of example values."""
+
+logger = logging.getLogger(__name__)
 
 
 class OpenSearchSchemaFetcher:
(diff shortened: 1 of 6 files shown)


lib/sycamore/sycamore/query/schema.py
```diff
@@ -31,32 +35,55 @@ def get_schema(self) -> OpenSearchSchema:
         schema = self._client.get_field_mapping(fields=["*"], index=self._index)
         query = OpenSearchQuery()
 
+        logger.debug(f"Getting schema for index {self._index}")
         # Fetch example values.
         query["index"] = self._index
         query["query"] = {"query": {"match_all": {}}, "size": self.NUM_EXAMPLES}
         random_sample = self._query_executor.query(query)["result"]["hits"]["hits"]
+        # logger.debug(f"Random sample came back with {random_sample}")
```
baitsguy, line 43: delete?

lib/sycamore/sycamore/utils/nested.py
```diff
@@ -0,0 +1,18 @@
...
+            return None
+        if keys[0] not in d:
+            return None
+        d = d[keys[0]]
+        keys = keys[1:]
+
+    return d
+
+
+def dotted_lookup(d: Any, keys: str) -> Any:
```
baitsguy, line 17: can you consolidate with https://github.com/aryn-ai/sycamore/blob/23747b1854523d38c785950f4ac9ec78307eec4b/lib/sycamore/sycamore/data/document.py#L217 ? 2 methods doing similar stuff

lib/sycamore/sycamore/tests/unit/query/test_schema.py
```diff
@@ -73,6 +96,16 @@ def test_opensearch_schema():
         }
     }
 
+    # Verify we can handle schems where int and float are both used
```
mdwelsh, line 99:
```suggestion
    # Verify we can handle schemas where int and float are both used
```

lib/sycamore/sycamore/utils/nested.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any
```
mdwelsh, line 1: Seems odd that this would need to be its own file.

mdwelsh (approved):

lib/sycamore/sycamore/query/schema.py
eric-anderson: Done.

lib/sycamore/sycamore/utils/nested.py
eric-anderson: Done. I thought I had a todo to do that. Turned out to be not too hard.

lib/sycamore/sycamore/tests/unit/query/test_schema.py
eric-anderson: Done.

lib/sycamore/sycamore/utils/nested.py
eric-anderson: I agree, but utils seems like the right place and none of the files in there looked like a good choice.

PR #843, opened 2024-09-25
Allow few-shot examples for query planner to be read from a file.

mdwelsh:
This PR moves us towards a configuration-file based approach to customizing Luna behavior for individual customers. Instead of the few-shot planner examples being hardcoded in Python, they can be specified as a YAML file that can be loaded by the app and passed to the Luna planner.

Currently this file is only read by the `query-ui` app, but if we like this approach, I can do a follow up to incorporate this functionality into the `LlmPlanner` class directly.

To make this simpler, I also switched up some of the data types being used for planner examples and OpenSearch schema to be Pydantic models so we could get automatic JSON and YAML support. This reduces the amount of ad hoc serialization/deserialization code we need to write.

The idea here is that for a given customer deployment, we would write a (new) `lunaconfig.yml` file containing the schema and few-shot example query plans, and when running the Luna UI we would point it at this file.


apps/query-ui/queryui/util.py
```diff
@@ -85,86 +143,136 @@ def docset_to_string(docset: DocSet) -> str:
...
+            # Keep only the columns we care about.
+            # Limit size of each to avoid blowing out memory.
+            def format_value(value):
+                MAX_CHARS = 1024
+                if isinstance(value, str):
+                    if len(value) > MAX_CHARS:
+                        return value[:MAX_CHARS] + "..."
+                    else:
+                        return value
+                if isinstance(value, dict):
```
eric-anderson, line 210: I would think you would want return format_value(str(value)). If you end up with a big class in the value it seems like you would want to truncate it.
I'd probably then just rewrite to:
```
  if not isinstance(value, str):
       value = str(value)
  MAX_CHARS = 1024 
  if len(value) > MAX_CHARS:
       return value[:MAX_CHARS] + "..."
  else:
       return value
```

apps/query-ui/queryui/util.py
```diff
@@ -85,86 +143,136 @@ def docset_to_string(docset: DocSet) -> str:
...
+                    return format_value(str(value))
+                if isinstance(value, list):
+                    return format_value(str(value))
+                return value
+
+            row = {k: format_value(doc.get(k)) for k in doc.keys() if k in self.COLUMNS or k.startswith("properties.")}
+            return row
+
+    def readdata(self):
```
eric-anderson, line 219: readdir or readtraces?

apps/query-ui/queryui/util.py
```diff
@@ -85,86 +143,136 @@ def docset_to_string(docset: DocSet) -> str:
...
+        directory = os.path.join(self.trace_dir, self.node_id)
```
eric-anderson, line 220: I like using the pathlib library for stuff like this:
```
directory = Path(self.trace_dir) / self.node_id
docs = [d for d in [self.readfile(f) for f in directory.glob("*") if f.is_file()] if d is not None]
```
does the same as the entire for filename loop

apps/query-ui/queryui/util.py
```diff
@@ -85,86 +143,136 @@ def docset_to_string(docset: DocSet) -> str:
...
+        "text_representation",
+        "doc_id",
+        "parent_id",
+    ]
+
+    def __init__(self, trace_dir: str, node_id: str):
+        self.trace_dir = trace_dir
+        self.node_id = node_id
+        self.df = None
+        self.readdata()
```
eric-anderson, line 176: I like calling internal methods with an _ to make it clear they aren't part of the public API. Here I think thats _readdata() and _readfile().

mdwelsh:
I'm closing this PR in favor of #852 which provides similar customization by modifying a Python file instead, which I like better.

PR #844, opened 2024-09-25
Another try at fixing path prompting

eric-anderson:
The example was just wrong (matching *.pdf) will never work unless there is a file with the name "/path/to/data/*.pdf". In my testing it caused openai to try matching properties.path.keyword = "*22*" when asked about item 22.

Be more explicit in the prompt about using keyword for exact matches and including full paths.

### DIFF
--- a/lib/sycamore/sycamore/query/operators/query_database.py
+++ b/lib/sycamore/sycamore/query/operators/query_database.py
@@ -33,7 +33,7 @@ class QueryDatabase(LogicalOperator):
                     },
                     {
                         "match": {
-                            "properties.path.keyword": "/path/to/data/*.pdf",
+                            "properties.path": "/path/to/data/*.pdf",
                             "properties.entity.location": "Georgia"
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/query/operators/query_database.py
```diff
@@ -33,7 +33,7 @@ class QueryDatabase(LogicalOperator):
                     },
                     {
                         "match": {
-                            "properties.path.keyword": "/path/to/data/*.pdf",
+                            "properties.path": "/path/to/data/*.pdf",
```
mdwelsh, line 36: I'm confused by this. We want is to use the `.keyword` version of the field so the wildcard match will work. I realize more few-shots may be needed to get this dialed in but I don't think you want to take `.keyword` out here.

eric-anderson:

eric-anderson:
obsolete

PR #849, opened 2024-09-26
Ignore bad tables

MarkLindblad:


lib/sycamore/sycamore/transforms/table_structure/table_transformers.py
```diff
@@ -753,8 +753,8 @@ def objects_to_structures(objects, tokens, class_thresholds):
 
     tables = [obj for obj in objects if obj["label"] == "table"]
 
-    assert len(tables) <= 1
-    if len(tables) == 0:
+    # assert len(tables) <= 1
```
eric-anderson, line 756: I'd recommend:
```
if len(tables) == 0:
    return {}
if len(tables) > 1:
   logger.warning("Got multiple tables in document. Using only the first one")
```

MarkLindblad: Done.

eric-anderson (approved):

bsowell (approved):

alexaryn:
Seems like a good quick hack.  We should revisit soon.

PR #852, opened 2024-09-27
Fold in query-demo capability to query-ui.

mdwelsh:
This PR migrates the functionality of the `query-demo` UI into the `query-ui` app under the "Chat" pane.

(I will follow up with a PR to remove `query-demo`, but we are still running it on the Luna demo instance.)

In addition to migrating the functionality over, this PR makes a number of cleanups:
  * The `Chat.py` application code is simplified and helper functions are migrated to the new file `chat.py`.
  * The NTSB-specific customizations are moved to `ntsb.py`. This makes it easier to tweak the UI for different use cases.
  (Note that in #843 I am migrating much of this to a separate configuration file instead of a Python file. After #843 lands I will update this to use the config file instead.)
  * I have made the way the UI is launched to always use `poetry run python queryui/main.py` which fires up Ray before launching the Streamlit app as a subprocess. Running Streamlit+Ray together in one process is not supported.
  * I have added some command-line flags to `main.py` so the OpenSearch index, trace dir, LLM cache dir, etc. can be specified on the commandline.

I realize this is a largeish PR but the functionality is the same as the existing `query-demo`.


apps/query-ui/queryui/main.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def main():
+    argparser = argparse.ArgumentParser()
+    argparser.add_argument("--chat", action="store_true", help="Only show the chat demo pane.")
+    argparser.add_argument(
+        "--index", help="OpenSearch index name to use. If specified, only this index will be queried."
+    )
+    argparser.add_argument(
+        "--llm-cache-dir", type=str, default="llm_cache", help="LLM query cache dir. Defaults to ./llm_cache."
```
baitsguy, line 43: probably want to use cache-dir from https://github.com/aryn-ai/sycamore/pull/850/files#diff-d6138cc332b0d2bb7e774f837ecceb3056c656c38b1faafff1b1703fbdb6b340 ?

apps/query-ui/queryui/pages/Trace_Viewer.py
```diff
@@ -7,24 +7,30 @@
...
     query_dirs_dt.sort(key=lambda x: x[1], reverse=True)
-    for query_id in query_dirs_dt:
-        with st.expander(f"Query `{query_id[0]}` ({query_id[1]})", expanded=False):
-            st.write(f"Query {query_id}")
-            show_query_traces(trace_dir, query_id[0])
 
+    def trace_summary(query_dir):
+        return f"{query_dir[0]} ({query_dir[1]})"
+
+    query_id = st.selectbox(
```
baitsguy, line 20: nice

apps/query-ui/queryui/util.py
```diff
@@ -66,105 +80,196 @@ def docset_to_string(docset: DocSet) -> str:
...
+            # For now, skip over MetadataDocuments.
+            if "metadata" in doc.keys():
+                return None
+
+            # Flatten properties.
+            if "properties" in doc:
+                for prop in doc["properties"]:
+                    if isinstance(doc["properties"][prop], dict):
+                        for nested_property in doc["properties"][prop]:
+                            doc[".".join(["properties", prop, nested_property])] = doc["properties"][prop][
```
baitsguy, line 188: for some cases I needed to convert the value to a string as well (multiple inferred types for the same key). So I'd do a str(doc["properties"][prop][next_property]) since this is for displaying only

apps/query-ui/queryui/util.py
```diff
@@ -66,105 +80,196 @@ def docset_to_string(docset: DocSet) -> str:
...
+                                nested_property
+                            ]
+                    else:
+                        doc[".".join(["properties", prop])] = doc["properties"][prop]
```
baitsguy, line 192: same str casting

baitsguy (commented):
couple questions

apps/query-ui/queryui/main.py
mdwelsh: Yes, done. I had sent the PR for that concurrently to this one :-)

apps/query-ui/queryui/util.py
mdwelsh: So, the problem with that is that casting to str doesn't let streamlit do its magic with rendering, e.g., lists as individual items. Is the issue only for multiple types with the same key? I'm a little confused why we would end up with different types for the same key though.

mdwelsh (commented):
Thanks for the comments!

baitsguy (approved):
Tests broken

PR #853, opened 2024-09-30
Define parallelism on nodes

eric-anderson:
* Allows us to remove a dependency on importing ray.
* tests/integration/test_map.py confirms that the parallelism setting is working
* notebooks/default-prep-script.ipynb verifies we don't import ray in local mode
* moved serializability checks into the ray execution function. For the writer, I could just remove the serializeability check since it's done by basemap (manually verified)
* add check_serializable test as an example of something that can't be serialized.
* a fair bit of ugly code to patch in ray stuff in cases where we bypass the Executor in tests.
* fix handling of parallelism in file readers, they use -1 == unlimited, but ActorPoolStrategy asserts on size <= 0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ traces/
 apps/jupyter/bind_dir/poetry_cache
 apps/query-ui/cache_dir
 docs/build
+notebooks/default-prep-data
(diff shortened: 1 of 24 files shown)


bsowell (approved):
LGTM. Discussed a bit with Eric to understand the details. Seems like in the future we may want to support parallelism for non-actor-based operations, but that's not something we have had much need for right now.

PR #855, opened 2024-10-01
Super basic FastAPI wrapper to Sycamore Query.

mdwelsh:
This is a trivial FastAPI wrapper to Sycamore Query, providing a simple REST API for running Luna queries.

This is by no means complete, and lacks a bunch of features we'll eventually want, but provides a starting point that we can build on.

Missing obvious features:
  * Logging (beyond what FastAPI already does)
  * Status updates on long-running queries
  * Asynchronous query execution
  * Configurability (e.g., specifying OpenSearch location or indices to use, etc.)
  * Tests
  
Even so this gives us a skeleton we can add these things into.


baitsguy (approved):
Awesome

PR #857, opened 2024-10-01
Implement output_format in Aryn SDK partition_file().

alexaryn:


HenryL27 (approved):
can you also bump the sdk version?
```
cd lib/aryn-sdk
poetry version patch
```

PR #861, opened 2024-10-02
Get Local Mode working 2/n

eric-anderson:
* Switch examples/markdown.py and examples/s3_ingest.py over to local mode.
* Fix s3_ingest to have the correct path adjustment. Fix it to only setup the filesystem if the AWS env vars are not present (enabled verifying that auto-fs detection worked) Switch it from taking a manifest to taking a path. I was unable to find an example of using the manifest in files, and given the python path bug, it means no one was using the script.

* Fix file scan to be able to work when the filesystem isn't specified.
* Add logging for when we run steps in local mode -- it was hard to debug if things were having problems or just being slow.
* docset.py, regex_replace.py: Fix escape sequences in docstrings that get warnings on 3.12.

### DIFF
--- a/examples/s3_ingest.py
+++ b/examples/s3_ingest.py
@@ -1,14 +1,16 @@
 import sys
 import boto3
 import pyarrow.fs
+import os
+
+# Usage: poetry run python s3_ingest.py s3://<something> [s3://another-thing ...]
 
 # ruff: noqa: E402
-sys.path.append("../sycamore")
+sys.path.append("../lib/sycamore")
 
 import sycamore
 from sycamore.functions.tokenizer import HuggingFaceTokenizer
 from sycamore.llms import OpenAIModels, OpenAI
-from sycamore.connectors.file.file_scan import JsonManifestMetadataProvider
 from sycamore.transforms import COALESCE_WHITESPACE
 from sycamore.transforms.merge_elements import MarkedMerger
 from sycamore.transforms.partition import UnstructuredPdfPartitioner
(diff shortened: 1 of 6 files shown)


examples/s3_ingest.py
```diff
@@ -1,14 +1,16 @@
 import sys
 import boto3
 import pyarrow.fs
+import os
+
+# Usage: poetry run python s3_ingest.py s3://<something> [s3://another-thing ...]
```
HenryL27, line 6: maybe:
```python
if len(sys.argv) <= 1:
    print("Usage is blah")
    exit()
```

lib/sycamore/sycamore/executor.py
```diff
@@ -135,15 +138,30 @@ def visit(n):
...
+        def get_name(f):
+            try:
+                return f._name  # handle the case of basemap transforms
+            except AttributeError:
+                pass
+            try:
+                return f.__name__
+            except AttributeError:
+                pass
+            return f.__class__.__name__
```
HenryL27, line 150: maybe use `hasattr` / `getattr`?

HenryL27 (approved):

examples/s3_ingest.py
eric-anderson: Done.

lib/sycamore/sycamore/executor.py
eric-anderson: Done.

PR #864, opened 2024-10-02
Add similarity scoring and rerank transform

baitsguy:
Allows you to do:

```python
similarity_scorer = HuggingFaceTransformersSimilarityScorer(ignore_doc_structure=True)
doc_set = context.read.binary(paths).rerank(
    similarity_scorer=similarity_scorer, query="is this a cat?", score_property_name=score_property_name, limit=5
)
result = doc_set.take()
```
A key callout here is if we're doing a rerank on a non-exploded DocSet (i.e. not rag on chunks style), we will order Documents by their highest matching chunk, even though another Document might have slightly lower scoring but more chunks that are similar.

This PR also adds a local mode for BaseDBReader, will update our integration tests touching DBs to test both local and ray modes.

### DIFF
--- a/lib/sycamore/sycamore/data/element.py
+++ b/lib/sycamore/sycamore/data/element.py
@@ -23,6 +23,16 @@ class Element(UserDict):
             self.data["properties"] = {}
 
     @property
+    def id(self) -> Optional[int]:
+        """A unique identifier for the element within a Document. Represents an order within the document"""
+        return self.data.get("id")
+
+    @id.setter
+    def id(self, value: int) -> None:
+        """Set the unique identifier of the element within a Document."""
+        self.data["id"] = value
+
+    @property
     def type(self) -> Optional[str]:
         return self.data.get("type")
 
--- a/lib/sycamore/sycamore/data/element.py
+++ b/lib/sycamore/sycamore/data/element.py
@@ -23,14 +23,14 @@ class Element(UserDict):
             self.data["properties"] = {}
 
     @property
-    def id(self) -> Optional[int]:
+    def seq_no(self) -> Optional[int]:
         """A unique identifier for the element within a Document. Represents an order within the document"""
-        return self.data.get("id")
+        return self.data.get("seq_no")
 
-    @id.setter
-    def id(self, value: int) -> None:
+    @seq_no.setter
+    def seq_no(self, value: int) -> None:
         """Set the unique identifier of the element within a Document."""
-        self.data["id"] = value
+        self.data["seq_no"] = value
 
     @property
     def type(self) -> Optional[str]:
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -16,12 +16,13 @@ from sycamore.llms.prompts.default_prompts import (
 from sycamore.plan_nodes import Node, Transform
 from sycamore.transforms.augment_text import TextAugmentor
 from sycamore.transforms.embed import Embedder
-from sycamore.transforms import DocumentStructure
+from sycamore.transforms import DocumentStructure, Sort
 from sycamore.transforms.extract_entity import EntityExtractor, OpenAIEntityExtractor
 from sycamore.transforms.extract_graph_entities import GraphEntityExtractor
 from sycamore.transforms.extract_graph_relationships import GraphRelationshipExtractor
 from sycamore.transforms.extract_schema import SchemaExtractor, PropertyExtractor
 from sycamore.transforms.partition import Partitioner
+from sycamore.transforms.similarity import SimilarityScorer
 from sycamore.transforms.resolve_graph_entities import EntityResolver, ResolveEntities
 from sycamore.transforms.summarize import Summarizer
 from sycamore.transforms.llm_query import LLMTextQueryAgent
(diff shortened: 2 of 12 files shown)


lib/sycamore/sycamore/transforms/similarity.py
```diff
@@ -0,0 +1,224 @@
...
+    @timetrace("TransformersSimilarity")
+    def score(self, inputs: list[tuple[str, str]]) -> list[float]:
+        import torch
+
+        if not self._model or not self._tokenizer:
+            logger.info(f"Initializing transformers model: {self.model_name}")
+            from transformers import AutoModelForSequenceClassification, AutoTokenizer
+
+            self._tokenizer = AutoTokenizer.from_pretrained(self.model_name)
+            self._model = AutoModelForSequenceClassification.from_pretrained(self.model_name).to(self.device)
```
HenryL27, line 164: Do you need to do any explicit stateful actor things bc of this? I guess since it's a class you're good. idk I kinda lost track of how you differentiate them in all of eric's ray-removal

lib/sycamore/sycamore/tests/unit/transforms/test_similarity.py
```diff
@@ -0,0 +1,98 @@
...
+        result = similarity_scorer.generate_similarity_scores(
+            docs, query=query, score_property_name=score_property_name
+        )
+        result.sort(key=lambda doc: doc.properties.get(score_property_name, float("-inf")), reverse=True)
+        assert [doc.doc_id for doc in result] == [2, 1, 3, 4, 5]
+
+        assert result[0].properties[score_property_name + "_source_element_id"] == 7
+
+    def test_transformers_similarity_scorer_no_doc_structure(self):
+        similarity_scorer = HuggingFaceTransformersSimilarityScorer(ignore_doc_structure=True)
```
HenryL27, line 53: maybe use a small model for unit/integ tests? perhaps cross-encoder/ms-marco-TinyBERT-L-2

lib/sycamore/sycamore/data/element.py
```diff
@@ -22,6 +22,16 @@ def __init__(self, element=None, /, **kwargs):
...
 
+    @property
+    def id(self) -> Optional[int]:
+        """A unique identifier for the element within a Document. Represents an order within the document"""
+        return self.data.get("id")
+
+    @id.setter
+    def id(self, value: int) -> None:
+        """Set the unique identifier of the element within a Document."""
+        self.data["id"] = value
```
HenryL27, line 33: where is this actually set? also if it's just for ordering I might call it something like "seq_no"
might need to get plumbed through explode as well.

lib/sycamore/sycamore/tests/integration/transforms/test_rerank.py
```diff
@@ -0,0 +1,90 @@
...
+        {  # drop because of limit
+            "doc_id": 6,
+            "elements": [
+                {"id": 1},
+            ],
+        },
+    ]
+    docs = [Document(item) for item in dicts]
+
+    context = sycamore.init()
```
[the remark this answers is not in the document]
baitsguy: I'll do this as a follow up, will be updating the integ tests to all do both boths

lib/sycamore/sycamore/tests/integration/transforms/test_rerank.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    result = doc_set.take()
+
+    assert len(result) == len(docs)
+    assert [doc.doc_id for doc in result] == [2, 1, 3, 5, 4]
+
+    for doc in result:
+        if doc.doc_id == 4:
+            continue
+        assert float(doc.properties.get(score_property_name))
```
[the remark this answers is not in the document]
baitsguy: Lol let me see what I can do

lib/sycamore/sycamore/transforms/similarity.py
baitsguy: No clue, @eric-anderson anything to note here? We're initializing the model once per class instance

lib/sycamore/sycamore/tests/unit/transforms/test_similarity.py
baitsguy: Ack

lib/sycamore/sycamore/data/element.py
baitsguy: I'll be setting it in the following PR in the partitioners which create Elements. Can rename to seq_no

lib/sycamore/sycamore/tests/integration/transforms/test_rerank.py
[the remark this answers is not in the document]
baitsguy: actually I'm going to skip that, we expect predictable scores from this test anyway (asserting the order)

lib/sycamore/sycamore/docset.py
```diff
@@ -1101,6 +1102,40 @@ def query(self, query_executor: QueryExecutor, **resource_args) -> "DocSet":
...
+            query: The query string to compute similarity against.
+            score_property_name: The name of the key where the score will be stored in document.properties
+            limit: Limit scoring and sorting to fixed size.
+        """
+        from sycamore.transforms import ScoreSimilarity, Limit
+
+        if limit:
+            plan = Limit(self.plan, limit)
+        else:
+            plan = self.plan
```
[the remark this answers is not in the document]
baitsguy: Looked at our limit and sort implementations, it doesn't seem trivial. Skipping for now but we can add functionality if we find a use case

lib/sycamore/sycamore/tests/unit/transforms/test_similarity.py
baitsguy: Ended up using cross-encoder/ms-marco-MiniLM-L-2-v2 because the TinyBERT results were bad and felt wrong to use

HenryL27 (approved):
lgtm

PR #867, opened 2024-10-03
Do some escaping of special Markdown characters.

alexaryn:
I'm not touching the "styling" characters (italic, bold, etc.) because we may get then from OCR.


HenryL27 (approved):
lgtm

PR #868, opened 2024-10-03
Clean Up Text Extraction

karanataryn:
Addresses Alex's comments in https://github.com/aryn-ai/sycamore/pull/745 and fixes some other bugs.


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -501,9 +501,12 @@ def process_batch(
...
             with LogTime("text_extraction"):
-                for page in extractor_list:
-                    extracted_pages.append(text_extractor.extract_page(page))
-
+                for page_value in extractor_list:
+                    if isinstance(page_value, list):
+                        page = text_extractor.extract_page(page_data=page_value)
+                    else:
+                        page = text_extractor.extract_page(page_value)
+                    extracted_pages.append(page)
```
alexaryn, line 509: This smells funny.  On the one hand, these two calls involve subclasses of the same class and have the same method name.  On the other hand, the calling conventions and inputs are totally different.  At the very least, the `page_data` thing looks like a hack.  I think the cleaner thing to do here is to have different method names when things should be called differently.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -28,69 +28,60 @@ def get_text(self, image: Image.Image) -> str:
...
-        ocr_output = self.get_boxes_and_text(page)
-        width, height = page.size
-        texts: List[Element] = []
+    def parse_ocr_output(self, ocr_output: list[dict[str, Any]], width, height) -> list[Element]:
+        texts: list[Element] = []
         for obj in ocr_output:
-            if obj["bbox"] and not obj["bbox"].is_empty() and obj["text"] and len(obj["text"]) > 0:
+            obj_bbox = obj.get("bbox")
+            obj_text = obj.get("text")
+            if obj_bbox and not obj_bbox.is_empty() and obj_text:
```
alexaryn, line 39: nit: put the expensive test (is_empty) last

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -28,69 +28,60 @@ def get_text(self, image: Image.Image) -> str:
...
         return texts
 
-    @timetrace("OCR Document Extraction")
+    @timetrace("OCRPageEx")
+    def extract_page(self, page: Optional[Union["PDFPage", "Image.Image"]]) -> list[Element]:
+        assert isinstance(page, Image.Image)
+        ocr_output = self.get_boxes_and_text(page)
+        width, height = page.size
+        texts: list[Element] = self.parse_ocr_output(ocr_output, width, height)
+        return texts
```
alexaryn, line 58: If you just return `self.parse_ocr_output(...)` you can sidestep the unsightly annotation with no loss of type checking.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -28,69 +28,60 @@ def get_text(self, image: Image.Image) -> str:
...
                 )
-                text.text_representation = obj["text"]
+                text.text_representation = obj_text
                 texts.append(text)
         return texts
 
-    @timetrace("OCR Document Extraction")
+    @timetrace("OCRPageEx")
+    def extract_page(self, page: Optional[Union["PDFPage", "Image.Image"]]) -> list[Element]:
+        assert isinstance(page, Image.Image)
```
alexaryn, line 54: It's unexpected to see this immediately after the function is declared with a very broad type hint.  Python inheritance with mypy is a pain.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -28,69 +28,60 @@ def get_text(self, image: Image.Image) -> str:
...
-
-                    pages.append(texts)
-                if use_cache:
-                    logger.info("Cache Miss for OCR. Storing the result to the cache.")
-                    ocr_cache.set(hash_key, pages)
-                return pages
+        with tempfile.TemporaryDirectory() as tempdirname:  # type: ignore
+            filename = cast(str, filename)
+            images = kwargs.get("images")
+            generator = (image for image in images) if images else pdf_to_image_files(filename, Path(tempdirname))
```
alexaryn, line 70: There's a lot loaded into this line.  I'd probably write it out as a normal if-then-else.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -28,69 +28,60 @@ def get_text(self, image: Image.Image) -> str:
...
-                            text.text_representation = obj["text"]
-                            texts.append(text)
-
-                    pages.append(texts)
-                if use_cache:
-                    logger.info("Cache Miss for OCR. Storing the result to the cache.")
-                    ocr_cache.set(hash_key, pages)
-                return pages
+        with tempfile.TemporaryDirectory() as tempdirname:  # type: ignore
+            filename = cast(str, filename)
```
alexaryn, line 68: It looks like pdf_to_image_files() doesn't accept an IOBase.  Should we assert that filename is an instance of str?

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -28,69 +28,60 @@ def get_text(self, image: Image.Image) -> str:
...
+            images = kwargs.get("images")
+            generator = (image for image in images) if images else pdf_to_image_files(filename, Path(tempdirname))
+            pages = []
+            for image_file in generator:
+                if isinstance(image_file, Image.Image):
+                    image = image_file
+                else:
+                    image = Image.open(image_file).convert("RGB")
```
alexaryn, line 76: A little cleaner:
```
for image in generator:
    if isinstance(image, Path):
        image = Image.open(image).convert("RGB")
    ...
```

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -24,31 +24,22 @@ def __init__(self):
         from pdfminer.layout import LAParams
         from pdfminer.pdfinterp import PDFPageInterpreter, PDFResourceManager
 
-        rm = PDFResourceManager()
-        param = LAParams()
-        self.device = PDFPageAggregator(rm, laparams=param)
-        self.interpreter = PDFPageInterpreter(rm, self.device)
+        self.rm = PDFResourceManager()
+        self.param = LAParams()
```
alexaryn, line 28: Do we need to store these two in `self`?

alexaryn (commented):
Overall looks good, but there are a couple of things that could use changing.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
karanataryn: Agreed :/

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
karanataryn: Could be more than path, so adjusted it for that.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
karanataryn: Done.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
karanataryn: removed.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
karanataryn: Doesn't the cast deal with this already? We potentially get a wider range of acceptable entries this way.

lib/sycamore/sycamore/transforms/detr_partitioner.py
karanataryn: Changed.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -73,31 +73,23 @@ def extract_document(self, filename: str, hash_key: str, use_cache=False, **kwar
             return pages
 
     @timetrace("PdfMinerPageEx")
-    def extract_page(
-        self, page: Optional[Union["PDFPage", "Image"]] = None, page_data: list[tuple[Any, Any, Any, Any]] = []
-    ) -> list[Element]:
+    def extract_page(self, page: Optional[Union["PDFPage", "Image"]] = None) -> list[Element]:
```
alexaryn, line 76: So, I guess `page_data` wasn't needed after all?

alexaryn (approved):
Looks OK.  Some changes I think are needed before landing.  Approving now so as not to block you.

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
karanataryn: Not exactly. The functionality to support it was moved into a `TextExtractor` function.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
karanataryn: Changed.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -192,10 +177,39 @@ def get_text(self, image: Image.Image) -> str:
             return " ".join(text_values)
         return ""
 
+    def set_slicing_parameters(self, image_width, image_height) -> dict[str, Any]:
+        slicing_params = {}
+        slice_threshold = self.slice_kwargs.get("slice_threshold", 1000)  # Only slice big images
+        if image_width * image_height > slice_threshold ^ 2:
+            merge_threshold = self.slice_kwargs.get("image_threshold", 1000)
+            default_horizontal_stride = self.slice_kwargs.get("horizontal_stride", 300)  # pixels
```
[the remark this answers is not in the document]
karanataryn: Done.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -192,10 +177,39 @@ def get_text(self, image: Image.Image) -> str:
             return " ".join(text_values)
         return ""
 
+    def set_slicing_parameters(self, image_width, image_height) -> dict[str, Any]:
+        slicing_params = {}
+        slice_threshold = self.slice_kwargs.get("slice_threshold", 1000)  # Only slice big images
+        if image_width * image_height > slice_threshold ^ 2:
```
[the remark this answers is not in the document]
karanataryn: Good catch, math vs cs always a pain.

PR #876, opened 2024-10-04
Get llm_filter to support document structure + similarity sorting for elements

baitsguy:
**1. Support document model in llm_filter and some performance improvements**
llm_filter will work on a DocSet made of original Documents (i.e. pre-exploded state) when using `use_elements=True`
* Element iteration "early stopping" - a record (document) passes the filter if any of the document's elements pass the llm_filter condition, we don't evaluate every element. This helps for documents which would be a hit, but doesn't for documents which would not (because we would still iterate over every element)
* Optionally, it sorts elements by a similarity score to get potential llm hits sooner
* `keep_none` flag to dictate behavior for a missing property value

None of these are perfect but they align with the use cases we have so far and can keep iterating.

**2. Sorted elements returned by OpenSearchReader in reconstructed documents**
 Use element.element_index field to sort elements when reconstructing a document. Does an explicit sort after each reconstruction which is okay based on the magnitude of data.

**Note:** the `ignore_doc_structure` flag is to prevent current usage from breaking, once we update other code to use the document model we can remove/flip it

### DIFF
--- a/lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
+++ b/lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
@@ -143,8 +143,7 @@ class OpenSearchReaderQueryResponse(BaseDBReader.QueryResponse):
 
             # sort elements per doc
             for doc in result:
-                num_elements = len(doc.elements)
-                doc.elements.sort(key=lambda e: e.element_index if e.element_index is not None else num_elements)
+                doc.elements.sort(key=lambda e: e.element_index if e.element_index is not None else float("inf"))
 
         return result
 
--- a/lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
+++ b/lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
@@ -80,7 +80,7 @@ class OpenSearchReaderQueryResponse(BaseDBReader.QueryResponse):
 
     def to_docs(self, query_params: "BaseDBReader.QueryParams") -> list[Document]:
         assert isinstance(query_params, OpenSearchReaderQueryParams)
-        result = []
+        result: list[Document] = []
         if not query_params.reconstruct_document:
             for data in self.output:
                 doc = Document(
@@ -140,6 +140,11 @@ class OpenSearchReaderQueryResponse(BaseDBReader.QueryResponse):
                 parent.elements.append(Element(doc.data))
 
             result = list(unique_docs.values())
+
+            # sort elements per doc
+            for doc in result:
+                num_elements = len(doc.elements)
+                doc.elements.sort(key=lambda e: e.element_index if e.element_index is not None else num_elements)
 
         return result
 
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -4,6 +4,7 @@ from pathlib import Path
 import pprint
 import sys
 from typing import Callable, Optional, Any, Iterable, Type, Union, TYPE_CHECKING
+import re
 
 from sycamore.context import Context, context_params, OperationTypes
 from sycamore.data import Document, Element, MetadataDocument
@@ -956,6 +957,7 @@ class DocSet:
         return self.map(process_doc, **resource_args)
 
     @context_params(OperationTypes.BINARY_CLASSIFIER)
+    @context_params(OperationTypes.TEXT_SIMILARITY)
     def llm_filter(
         self,
         llm: LLM,
@@ -963,6 +965,9 @@ class DocSet:
         prompt: Union[list[dict], str],
         field: str = "text_representation",
         threshold: int = 3,
+        ignore_doc_structure: bool = True,
+        sort_elements_by_similarity: bool = False,
+        similarity_scorer: Optional[SimilarityScorer] = None,
         **resource_args,
     ) -> "DocSet":
         """
@@ -975,28 +980,41 @@ class DocSet:
             prompt: LLM prompt.
             field: Document field to filter based on.
             threshold: Cutoff that determines whether or not to keep document.
+            ignore_doc_structure: ignores document.element mapping.
+            sort_elements_by_similarity: sort elements by similarity to the prompt.
+            similarity_scorer: scorer to generate similarity if 'sort_elements_by_similarity' is True.
             **resource_args
 
         Returns:
             A filtered DocSet.
         """
-
-        def threshold_filter(doc: Document, threshold) -> bool:
-            try:
-                return_value = int(doc.properties[new_field]) >= threshold
-            except Exception:
-                # accounts for llm output errors
-                return_value = False
-
-            return return_value
-
-        docset = self.filter(lambda doc: doc.field_to_value(field) is not None and doc.field_to_value(field) != "None")
-
         entity_extractor = OpenAIEntityExtractor(
             entity_name=new_field, llm=llm, use_elements=False, prompt=prompt, field=field
         )
-        docset = docset.extract_entity(entity_extractor=entity_extractor)
-        docset = docset.filter(lambda doc: threshold_filter(doc, threshold), **resource_args)
+
+        def threshold_filter(doc: Document, threshold) -> bool:
+            if ignore_doc_structure:
+                if doc.field_to_value(field) is None:
+                    return False
+                doc = entity_extractor.extract_entity(doc)
+                return int(re.findall(r"\d+", doc.properties[new_field])[0]) >= threshold
+            else:
+                if sort_elements_by_similarity:
+                    assert similarity_scorer is not None
+                    score_property_name = f"{field}_similarity_score"
+                    doc = similarity_scorer.generate_similarity_scores(
+                        doc_batch=[doc], query=prompt, score_property_name=score_property_name
+                    )[0]
+                    doc.elements.sort(key=lambda e: e.properties.get(score_property_name, float("-inf")), reverse=True)
+                for element in doc.elements:
+                    e_doc = Document(element.data)
+                    e_doc = entity_extractor.extract_entity(e_doc)
+                    element.properties[new_field] = e_doc.properties[new_field]
+                    if int(re.findall(r"\d+", element.properties[new_field])[0]) >= threshold:
+                        return True
+            return False
+
+        docset = self.filter(lambda doc: threshold_filter(doc, threshold), **resource_args)
 
         return docset
 
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -966,7 +966,6 @@ class DocSet:
         field: str = "text_representation",
         threshold: int = 3,
         ignore_doc_structure: bool = True,
-        sort_elements_by_similarity: bool = False,
         similarity_query: Optional[str] = None,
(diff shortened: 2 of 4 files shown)


lib/sycamore/sycamore/docset.py
```diff
@@ -975,28 +980,48 @@ def llm_filter(
...
         )
-        docset = docset.extract_entity(entity_extractor=entity_extractor)
-        docset = docset.filter(lambda doc: threshold_filter(doc, threshold), **resource_args)
+
+        def threshold_filter(doc: Document, threshold) -> bool:
+            if ignore_doc_structure:
+                if doc.field_to_value(field) is None:
+                    return False
+                doc = entity_extractor.extract_entity(doc)
+                return int(re.findall(r"\d+", doc.properties[new_field])[0]) >= threshold
```
mdwelsh, line 1000: This is a strange operation. You're just returning the integer value of the first string of digits, but why? That seems incredibly brittle. It seems to me we want to enforce that the LLM output of the entity extractor is a numeric value in a certain range, not try to hack around it here if it's not. Can we push some logic down to `entity_extractor` with a validator to ensure the result is something we expect, and set it to, e.g., None, if not?

If you're trying to avoid brittleness in the entity_extractor here, you should also be using `doc.properties.get(new_field)` and using a default value in case it did not populate the field at all.

lib/sycamore/sycamore/docset.py
```diff
@@ -975,28 +980,48 @@ def llm_filter(
...
+
+            # If we're looking at a document level custom property (i.e. not text_representation),
+            # only evaluate the Document and not elements.
+            if field != "text_representation" and doc.field_to_value(field) is not None:
```
mdwelsh, line 1004: It would be nice if we did not have `text_representation` hard-coded deep in this part of the code as meaning an element-level property that can never (?) appear at a doc level.

I suppose I would MUCH rather it be explicit when `.llm_filter()` is called whether you want to operate at the doc level or the element level. That is, rather than have something like `ignore_doc_structure`, it seems that we should have a flag `elementwise` to explicitly say we want the LLM filter to operate on the document's elements, not the top-level doc. Better yet, it seems that this should really be a separate transform that does a 'for each element do X' operation instead of baking this logic into `.llm_filter` itself. 

So for example, if I want to filter documents based on the result of an elementwise boolean operation, I could do:
```
docset.with_property('weather_is_good', lambda doc: all(doc.elementwise(lambda e: filter_weather(e))))
   .filter(lambda doc: doc.properties['weather_is_good'] == False)
   ...
```

Or something to that effect. I think we're mixing doc- and element-level operations a bit too much in the API right now and it's confusing.

My point is that it kinda feels that we're hacking in logic here for this specific case, but what we're really bumping up against is the need to perform some ops at the doc level, and others at the element level.

lib/sycamore/sycamore/docset.py
```diff
@@ -975,28 +980,48 @@ def llm_filter(
...
+            if similarity_query or similarity_scorer:
+                assert similarity_scorer is not None, "Similarity sorting requires a scorer"
+                assert similarity_query is not None, "Similarity sorting requires a string query"
+                score_property_name = f"{field}_similarity_score"
+                doc = similarity_scorer.generate_similarity_scores(
+                    doc_batch=[doc], query=similarity_query, score_property_name=score_property_name
+                )[0]
+                doc.elements.sort(key=lambda e: e.properties.get(score_property_name, float("-inf")), reverse=True)
+            for element in doc.elements:
+                e_doc = Document(element.data)
```
mdwelsh, line 1017: This is really the way to lift an element into a Document?

lib/sycamore/sycamore/docset.py
```diff
@@ -975,28 +980,48 @@ def llm_filter(
...
+                e_doc = entity_extractor.extract_entity(e_doc)
+                element.properties[new_field] = e_doc.properties[new_field]
+                if int(re.findall(r"\d+", element.properties[new_field])[0]) >= threshold:
```
mdwelsh, line 1020: See comment above, I don't really like this approach...

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -141,6 +141,11 @@ def to_docs(self, query_params: "BaseDBReader.QueryParams") -> list[Document]:
 
             result = list(unique_docs.values())
 
+            # sort elements per doc
+            for doc in result:
+                num_elements = len(doc.elements)
```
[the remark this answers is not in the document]
baitsguy: element_index is sort of up to the partitioner currently and I don't know if we have full coverage. It also doesn't exist on already ingested data

lib/sycamore/sycamore/docset.py
baitsguy: >. Can we push some logic down to entity_extractor with a validator to ensure the result is something we expect, and set it to, e.g., None, if not?
This is the right solution but I don't want to combine it in this PR, there are quiet a few places we need to update. 

>If you're trying to avoid brittleness in the entity_extractor here, you should also be using doc.properties.get(new_field) and using a default value in case it did not populate the field at all.
We had discussed this earlier in the error handling discussion where we don't want to put missing data handling in operations (I believe it was missing dates from ntsb data), staying consistent with that for now. I can add a ignore_nulls style flag though

lib/sycamore/sycamore/docset.py
baitsguy: Yeah 🙃

lib/sycamore/sycamore/docset.py
baitsguy: >It would be nice if we did not have text_representation hard-coded deep in this part of the code as meaning an element-level property that can never (?) appear at a doc level.

That's the goal, but currently we have no way of enforcing a document model on the docset. The hope is to get there incrementally. In the current world its possible for a document object to have a text_representation and also have elements that have a text_representation, but in our logical document model we don't expect that to be the case

>I suppose I would MUCH rather it be explicit when .llm_filter() is called whether you want to operate at the doc level or the element level....

That works for me, we have a `use_elements` flag in entity extraction that basically decides whether to use the Document's information or document.elements. I don't think it's perfect but for consistency's sake maybe I just add that here? 

I dunno if we want to built the elementwise style approach yet, I think we should have a better idea of how we generally apply the data model. It seems useful but we might want to support different element projections which would impact that

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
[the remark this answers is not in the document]
baitsguy: will do the `key=lambda e: e.element_index or float('inf')`

lib/sycamore/sycamore/docset.py
mdwelsh: Sure, I think a decent intermediate point would be to use an "elementwise" flag or something on this function.

lib/sycamore/sycamore/docset.py
```diff
@@ -974,29 +980,51 @@ def llm_filter(
...
         )
-        docset = docset.extract_entity(entity_extractor=entity_extractor)
-        docset = docset.filter(lambda doc: threshold_filter(doc, threshold), **resource_args)
+
+        def threshold_filter(doc: Document, threshold) -> bool:
+            if not use_elements:
+                if doc.field_to_value(field) is None:
+                    return keep_none
+                doc = entity_extractor.extract_entity(doc)
+                return int(re.findall(r"\d+", doc.properties[new_field])[0]) >= threshold
```
mdwelsh, line 1003: Nit: Could you put a TODO above this line and line 1020 below explaining what this does and indicating that we should move the data validation down into the entity extractor?

lib/sycamore/sycamore/docset.py
```diff
@@ -974,29 +980,51 @@ def llm_filter(
             new_field: The field that will be added to the DocSet with the outputs.
             prompt: LLM prompt.
             field: Document field to filter based on.
-            threshold: Cutoff that determines whether or not to keep document.
+            threshold: Cutoff that determines whether to keep document.
+            keep_none:  whether to keep records with a None value.
+                        Warning: using this might hide data corruption issues.
+            use_elements: use contents of a document's elements to filter as opposed to document level contents.
+            similarity_query: query string to compute similarity against.
+            similarity_scorer: scorer to generate similarity if 'sort_elements_by_similarity' is True.
```
mdwelsh, line 988: What is `sort_elements_by_similarity` and where is it set?

lib/sycamore/sycamore/docset.py
```diff
@@ -974,29 +980,51 @@ def llm_filter(
             new_field: The field that will be added to the DocSet with the outputs.
             prompt: LLM prompt.
             field: Document field to filter based on.
-            threshold: Cutoff that determines whether or not to keep document.
+            threshold: Cutoff that determines whether to keep document.
+            keep_none:  whether to keep records with a None value.
```
mdwelsh, line 984: Does this mean "keep records with a None value for the provided field" ?

lib/sycamore/sycamore/docset.py
```diff
@@ -974,29 +980,51 @@ def llm_filter(
             new_field: The field that will be added to the DocSet with the outputs.
             prompt: LLM prompt.
             field: Document field to filter based on.
-            threshold: Cutoff that determines whether or not to keep document.
+            threshold: Cutoff that determines whether to keep document.
```
mdwelsh, line 983: Can you make this more clear? Maybe: "If the value of the computed result is an integer value greater than or equal to this threshold, the document will be kept."

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -141,6 +141,10 @@ def to_docs(self, query_params: "BaseDBReader.QueryParams") -> list[Document]:
 
             result = list(unique_docs.values())
 
+            # sort elements per doc
+            for doc in result:
+                doc.elements.sort(key=lambda e: e.element_index if e.element_index is not None else float("inf"))
```
mdwelsh, line 146: Does the shorter form not work?

`doc.elements.sort(key=lambda e: e.element_index or float("inf"))`

mdwelsh (approved):
A few nits but LGTM

lib/sycamore/sycamore/docset.py
baitsguy: Ack

lib/sycamore/sycamore/docset.py
baitsguy: Ah leftover comment

lib/sycamore/sycamore/docset.py
baitsguy: Yeah, let me clarify

lib/sycamore/sycamore/docset.py
baitsguy: Ack

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
baitsguy: element_index == 0 gets dropped that way

baitsguy: or pushed to the end rather

PR #880, opened 2024-10-07
Add Table Merger

dhruvkaliraman7:
The ``Table merger`` handles 3 operations
    1. If a text element (Caption, Section-header, Text...) contains the regex pattern anywhere in a page
     it is attached to the text_representation of the table on the page.
    2. LLMQuery is used for adding a table_continuation property to table elements. If the table is
     a continuation from a previous table the property is stored as true, else false.
    3. After LLMQuery, table elements which are continuations are merged as one element.

### DIFF
--- a/lib/sycamore/sycamore/transforms/llm_query.py
+++ b/lib/sycamore/sycamore/transforms/llm_query.py
@@ -72,7 +72,7 @@ class LLMTextQueryAgent:
                     if not self._table_cont:
                         document.elements[idx] = self._query_text_object(element)
                     else:
-                        if prev_table > 0:
+                        if prev_table >= 0:
                             document.elements[idx] = self._query_text_object(element, document.elements[prev_table])
                         else:
                             document.elements[idx] = self._query_text_object(element)
@@ -94,7 +94,7 @@ class LLMTextQueryAgent:
 
     @timetrace("LLMQueryText")
     def _query_text_object(
-        self, object: Union[Document, Element], objectPrev: Element = None
+        self, object: Union[Document, Element], objectPrev: Optional[Element] = None
     ) -> Union[Document, Element]:
         if object.text_representation:
             if self._format_kwargs:
--- a/lib/sycamore/sycamore/transforms/llm_query.py
+++ b/lib/sycamore/sycamore/transforms/llm_query.py
@@ -48,6 +48,7 @@ class LLMTextQueryAgent:
         llm_kwargs: dict = {},
         per_element: bool = True,
         element_type: Optional[str] = None,
+        table_cont: Optional[bool] = False,
     ):
         self._llm = llm
         self._prompt = prompt
@@ -57,16 +58,25 @@ class LLMTextQueryAgent:
         self._format_kwargs = format_kwargs
         self._number_of_elements = number_of_elements
         self._element_type = element_type
+        self._table_cont = table_cont
 
     def execute_query(self, document: Document) -> Document:
         final_prompt = self._prompt
         element_count = 0
+        prev_table = -1
         if self._per_element or self._number_of_elements:
             for idx, element in enumerate(document.elements):
                 if self._element_type and element.type != self._element_type:
                     continue
                 if self._per_element:
-                    document.elements[idx] = self._query_text_object(element)
+                    if not self._table_cont:
+                        document.elements[idx] = self._query_text_object(element)
+                    else:
+                        if prev_table > 0:
+                            document.elements[idx] = self._query_text_object(element, document.elements[prev_table])
+                        else:
+                            document.elements[idx] = self._query_text_object(element)
+                        prev_table = idx
                 else:
                     final_prompt += "\n" + element["text_representation"]
                 if self._number_of_elements:
@@ -83,7 +93,9 @@ class LLMTextQueryAgent:
         return document
 
     @timetrace("LLMQueryText")
-    def _query_text_object(self, object: Union[Document, Element]) -> Union[Document, Element]:
+    def _query_text_object(
+        self, object: Union[Document, Element], objectPrev: Element = None
+    ) -> Union[Document, Element]:
         if object.text_representation:
             if self._format_kwargs:
                 prompt = (
@@ -92,10 +104,16 @@ class LLMTextQueryAgent:
                     .render(doc=object)
                 )
             else:
-                prompt = self._prompt + "\n" + object.text_representation
+                if objectPrev and objectPrev.text_representation:
+                    prompt = self._prompt + "\n" + objectPrev.text_representation + "\n\n" + object.text_representation
+                else:
+                    prompt = self._prompt + "\n" + object.text_representation
             prompt_kwargs = {"prompt": prompt}
             llm_resp = self._llm.generate(prompt_kwargs=prompt_kwargs, llm_kwargs=self._llm_kwargs)
-            object["properties"][self._output_property] = llm_resp
+            if self._table_cont:
+                object["properties"]["table_continuation"] = llm_resp
+            else:
+                object["properties"][self._output_property] = llm_resp
         return object
 
 
--- a/lib/sycamore/sycamore/transforms/merge_elements.py
+++ b/lib/sycamore/sycamore/transforms/merge_elements.py
@@ -429,10 +429,14 @@ class TableMerger(ElementMerger):
 
             llm = OpenAI(OpenAIModels.GPT_4O, api_key = '')
 
-            prompt = "Analyze two CSV tables that may be parts of a single table split across pages. Determine if the second table is a continuation of the first with 100% certainty. Check either of the following:\
-            1. Column headers: Must be near identical in terms of text(the ordering/text may contain minor errors because of OCR quality) in both tables. If the headers are almost the same check the number of columns, they should be roughly the same.\
-            2. Missing headers: If the header/columns in the second table are missing, then the first row in the second table should logically be in continutaion of the last row in the first table.\
-            Respond with only 'true' or 'false' based on your certainty that the second table is a continuation. Certainty is determinedx if either of the two conditions is true."
+            prompt = "Analyze two CSV tables that may be parts of a single table split across pages. Determine if the second table\
+                      is a continuation of the first with 100% certainty. Check either of the following:\
+            1. Column headers: Must be near identical in terms of text(the ordering/text may contain minor errors because of OCR quality)\
+               in both tables. If the headers are almost the same check the number of columns, they should be roughly the same.\
+            2. Missing headers: If the header/columns in the second table are missing, then the first row in the second table should logically\
+               be in continutaion of the last row in the first table.\
+            Respond with only 'true' or 'false' based on your certainty that the second table is a continuation. \
+            Certainty is determined if either of the two conditions is true."
 
             regex_pattern = r"table \d+"
 
--- a/lib/sycamore/sycamore/transforms/merge_elements.py
+++ b/lib/sycamore/sycamore/transforms/merge_elements.py
@@ -493,6 +493,9 @@ class TableMerger(ElementMerger):
 
     def merge(self, elt1: Element, elt2: Element) -> Element:
 
+        # Check if both elements are TableElements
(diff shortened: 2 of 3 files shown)


lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -48,6 +48,7 @@ def __init__(
         llm_kwargs: dict = {},
         per_element: bool = True,
         element_type: Optional[str] = None,
+        table_cont: Optional[bool] = False,
```
HenryL27, line 51: I think this makes more sense as a subclass? hmm, not sure. worth thinking through the factoring at least

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -57,16 +58,25 @@ def __init__(
...
         if self._per_element or self._number_of_elements:
             for idx, element in enumerate(document.elements):
                 if self._element_type and element.type != self._element_type:
                     continue
                 if self._per_element:
-                    document.elements[idx] = self._query_text_object(element)
+                    if not self._table_cont:
+                        document.elements[idx] = self._query_text_object(element)
+                    else:
+                        if prev_table > 0:
```
HenryL27, line 75: >=

lib/sycamore/sycamore/transforms/llm_query.py
```diff
@@ -83,7 +93,9 @@ def execute_query(self, document: Document) -> Document:
         return document
 
     @timetrace("LLMQueryText")
-    def _query_text_object(self, object: Union[Document, Element]) -> Union[Document, Element]:
+    def _query_text_object(
+        self, object: Union[Document, Element], objectPrev: Element = None
```
HenryL27, line 97: object_prev: Optional[Element] = None

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+
+            merger = TableMerger(llm_prompt = prompt, llm=llm)
+
+            context = sycamore.init()
+            pdf_docset = context.read.binary(paths, binary_format="pdf", regex_pattern= regex_pattern)
+                .partition(partitioner=ArynPartitioner())
+                .merge(merger=merger)
+    """
+
+    def __init__(self, regex_pattern=None, llm_prompt=None, llm=None, *args, **kwargs):
```
HenryL27, line 446: type annotations please

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+            else:
+                new_table_elements.append(element)
+        other_elements.extend(new_table_elements)
+        document.elements = other_elements
+        return document
+
+    def should_merge(self, element1: Element, element2: Element) -> bool:
+        if "true" in element2["properties"]["table_continuation"].lower():
+            return True
+        return False
```
HenryL27, line 476: return "true" in blah

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+
+        table_elements = [ele for ele in document.elements if ele.type == "table"]
+        if len(table_elements) < 1:
+            return document
+        if self.regex_pattern:
+            document.elements = self.customTableHeaderAdditionFilter(document.elements)
+        if not self.llm_prompt or len(table_elements) < 2:
+            return document
+        document = self.process_llm_query(document)
+        table_elements = [ele for ele in document.elements if ele.type == "table"]
```
HenryL27, line 461: repeated?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+        other_elements = [ele for ele in document.elements if ele.type != "table"]
+        new_table_elements = [table_elements[0]]
+        for element in table_elements[1:]:
+            if self.should_merge(new_table_elements[-1], element):
+                new_table_elements[-1] = self.merge(new_table_elements[-1], element)
+            else:
+                new_table_elements.append(element)
+        other_elements.extend(new_table_elements)
+        document.elements = other_elements
```
HenryL27, line 470: prob want to add a sort step in here

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+
+    def should_merge(self, element1: Element, element2: Element) -> bool:
+        if "true" in element2["properties"]["table_continuation"].lower():
+            return True
+        return False
+
+    def merge(self, elt1: Element, elt2: Element) -> Element:
+
+        new_elt = Element()
+        new_elt.type = "table"
```
HenryL27, line 481: also prob need to find a way to merge the cells in the underlying table objects.
something like
```
translate cells in table2 down by table1.nrows rows
concat the cell lists and construct a new table
```

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+
+        dic = defaultdict(str)
+
+        # First pass: capture headers
+        for ele in elements:
+            if ele.type in ["table", "Image", "Formula"]:
+                continue
+            elif ele.type in ["Text", "Title", "Page-header", "Section-header", "Caption"]:
+                if ele.text_representation is not None:
+                    text_rep = ele.text_representation.strip().lower()
```
HenryL27, line 521: Case-sensitivity is tricky. What if I want to filter "Table \d+\." but not "TABLE \d+\." for whatever reason? Idk, if there's a way to encode case-sensitivity in re then I'd drop the .lower() but if there isn't then just make it super well documented.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+                    continue
+                if re.search(self.regex_pattern, text_rep):
+                    dic[ele["properties"]["page_number"]] = text_rep + " "
+
+        # Second pass: update table elements with headers, done in separate loops since
+        # table headers can be within table elements as well or after them
+        for ele in elements:
+            if ele.type == "table" and isinstance(ele["table"], Table):
+                ele.text_representation = dic[ele["properties"]["page_number"]] + ele.text_representation
+                ele["properties"]["table_header"] = dic[ele["properties"]["page_number"]]
```
HenryL27, line 532: pretty sure the Table object has a `caption` field or some such that might be handy here

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+
+        return elements
+
+    def process_llm_query(self, document):
+        llm_query_agent = LLMTextQueryAgent(prompt=self.llm_prompt, element_type="table", llm=self.llm, table_cont=True)
+        llm_results = llm_query_agent.execute_query(document)
+        return llm_results
```
HenryL27, line 539: how's the performance of this? I know ritam did a bunch of stuff to enable asynchronous llm queries, are we using that here / do we need to?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -412,6 +415,136 @@ def merge_elements(self, document: Document) -> Document:
...
+
+    def preprocess_element(self, elem: Element) -> Element:
+        return elem
```
HenryL27, line 542: I wonder if you can encode enough logic in here and in `should_merge` to not need to override `merge_elements`? prob not I guess since you might need to merge non-adjacent elements?

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: Yea, cause the text_representation would change when I add table header addition

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: Yes, cause of non-adjacent

lib/sycamore/sycamore/transforms/llm_query.py
dhruvkaliraman7: @dtecuci ?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -483,7 +483,9 @@ def merge_elements(self, document: Document) -> Document:
         return document
 
     def should_merge(self, element1: Element, element2: Element) -> bool:
-        return "true" in element2["properties"]["table_continuation"].lower()
+        if "table_continuation" in element2["properties"]:
+            return "true" in element2["properties"]["table_continuation"].lower()
```
HenryL27, line 487: can we just make `elt['properties']['table_continuation']` a bool?

HenryL27: oh nvm it's coming from llm

lib/sycamore/sycamore/transforms/llm_query.py
dhruvkaliraman7: Decided to not have subclass

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: Tabled for later if workload increases

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: done

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: Appending to 'Title'

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: done

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: done

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -429,12 +430,13 @@ class TableMerger(ElementMerger):
...
-            1. Column headers: Must be near identical in terms of text(the ordering/text may contain minor errors because of OCR quality)\
-               in both tables. If the headers are almost the same check the number of columns, they should be roughly the same.\
-            2. Missing headers: If the header/columns in the second table are missing, then the first row in the second table should logically\
-               be in continutaion of the last row in the first table.\
+            prompt = "Analyze two CSV tables that may be parts of a single table split across pages. Determine\
+            if the second table is a continuation of the first with 100% certainty. Check either of the following:\
+            1. Column headers: Must be near identical in terms of text(the ordering/text may contain minor errors \
+            because of OCR quality) in both tables. If the headers are almost the same check the number of columns,\
+                 they should be roughly the same.\
+            2. Missing headers: If the header/columns in the second table are missing, then the first row in the
```
HenryL27, line 438: Is the spacing gonna do the right thing here? I think you'll get "...the same.2. Missing..." but you want "...the same. 2. Missing headers" or something. tbh I'm not sure how python escapes whitespace tho so I might be wrong

HenryL27 (approved):

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: done

PR #882, opened 2024-10-07
Postprocessing to modify plans returned by llm planner; minor issues with query-ui

amolvdeshpande:
1. Added a post-processing step where the plan returned by the LLM Planner is modified if it uses a "QueryVectorDatabase" in the beginning, followed by a "count" or "extract_entity". Instead, we use a "QueryDatabase" followed by an LlmFilter. 

2. Fixed a couple of minor issues in query-ui so that "Execute Code" works and prints results for queries with a single answer.


lib/sycamore/sycamore/query/planner.py
```diff
@@ -129,6 +135,105 @@ def process_json_plan(
...
+            "role": "system",
+            "content": """You are a helpful agent that assists in small transformations
+                    of input text as per the instructions. You should make minimal changes
+                    to the provided input and keep your response short""",
+        },
+        {"role": "user", "content": user_message},
+    ]
+
+    prompt_kwargs = {"messages": messages}
+    chat_completion = OpenAI(OpenAIModels.GPT_4O.value).generate(
```
baitsguy, line 150: we'll want to use the llm object from the class here so we can support non-openai cases

lib/sycamore/sycamore/query/planner.py
```diff
@@ -129,6 +135,105 @@ def process_json_plan(
...
+                    {op['description']}"""
+        )
+
+        if "opensearch_filter" in op:
+            new_op = {
+                "operatorName": "QueryDatabase",
+                "description": modified_description,
+                "index": op["index"],
+                "node_id": 0,
+                "query": op["opensearch_filter"],
```
baitsguy, line 187: might be better to create the new_op dict first and only set "query" based on the condition, saves us from mistakes if adding/removing fields

lib/sycamore/sycamore/query/planner.py
```diff
@@ -129,6 +135,105 @@ def process_json_plan(
...
+                "node_id": 0,
+                "query": {"match_all": {}},
+            }
+
+        postprocessed_plan[0] = new_op
+
+        # Add an LLM Filter as the second operator
+        llm_op_description = postprocess_llm_helper(
+            f"""
+                Generate a one-line description for a python function whose goal is to filter the input r
```
baitsguy, line 203: does the break in the word "records" matter here

lib/sycamore/sycamore/query/planner.py
```diff
@@ -129,6 +135,105 @@ def process_json_plan(
...
+                ecords based on whether they contain {op['query_phrase']}.
+                The output should be of the format: Filter to records involving wildfires.
```
baitsguy, line 205: should we say this is an example instead of saying its the format

lib/sycamore/sycamore/tests/unit/query/test_plan.py
```diff
@@ -159,7 +159,7 @@ def vector_search_filter_plan():
 
 
 def get_logical_plan(plan):
```
baitsguy, line 161: can we add a test for post_process as well

lib/sycamore/sycamore/query/planner.py
```diff
@@ -77,7 +78,7 @@
 
 
 def process_json_plan(
-    plan: typing.Union[str, Any], operators: Optional[List[Type[LogicalOperator]]] = None
+    plan: typing.Union[str, Any], operators: Optional[List[Type[LogicalOperator]]] = None, postProcess: bool = False
```
baitsguy, line 81: nit: we're using `post_process` style naming, also can you default to True? I think we always want this now

baitsguy (commented):
Looks good, a few small questions/nits and requested a unit test

lib/sycamore/sycamore/query/planner.py
amolvdeshpande: Fixed to use llm_client from the class.

lib/sycamore/sycamore/query/planner.py
amolvdeshpande: Done.

lib/sycamore/sycamore/query/planner.py
amolvdeshpande: Fixed.

lib/sycamore/sycamore/query/planner.py
amolvdeshpande: Fixed.

lib/sycamore/sycamore/tests/unit/query/test_plan.py
amolvdeshpande: Added unit tests.

lib/sycamore/sycamore/query/planner.py
amolvdeshpande: Remove postprocess as a flag to process_json_plan (which was added to handle unit tests for that function).

baitsguy (approved):
We can work on writing a separate optimizer as we build a techniques but this should get us started

apps/query-ui/queryui/Sycamore_Query.py
```diff
@@ -32,28 +37,43 @@ def show_schema(_client: SycamoreQueryClient, index: str):
...
+                # exec(code, global_context)
             except Exception as e:
                 st.exception(e)
             if code_locals and "result" in code_locals:
                 st.subheader("Result", divider="rainbow")
                 st.success(code_locals["result"])
-            if st.session_state.trace_dir:
-                st.subheader("Traces", divider="blue")
-                util.show_query_traces(st.session_state.trace_dir, st.session_state.query_id)
+            # Commenting this out for now -- we don't have the right query_id for this to work TODO
```
mdwelsh, line 72: This is only in the case where we did code generation, right?

lib/sycamore/sycamore/query/planner.py
```diff
@@ -129,6 +130,105 @@ def process_json_plan(
...
+
+    if postprocessed_plan[0]["operatorName"] == "QueryVectorDatabase" and (
+        len(postprocessed_plan) == 1 or postprocessed_plan[1]["operatorName"] in ["Count", "LlmExtractEntity"]
+    ):
+        # If the first operator has an "opensearch_filter", we will convert it to a QueryDatabase with
+        #        that opensearch_filter as "query"
+        # If not, we do a QueryDatabase with "match_all" instead
+        op = postprocessed_plan[0]
+
+        modified_description = postprocess_llm_helper(
```
mdwelsh, line 169: I'm very surprised you need an LLM call for this. Can you explain?

lib/sycamore/sycamore/query/planner.py
```diff
@@ -129,6 +130,105 @@ def process_json_plan(
...
+        new_op["description"] = modified_description
+
+        if "opensearch_filter" in new_op:
+            new_op["query"] = new_op.pop("opensearch_filter")
+        else:
+            new_op["query"] = {"match_all": {}}
+
+        postprocessed_plan[0] = new_op
+
+        # Add an LLM Filter as the second operator
```
mdwelsh, line 189: Again, surprised we need an LLM call to do this. Can't we just add the LLM filter op directly?

lib/sycamore/sycamore/query/planner.py
```diff
@@ -129,6 +130,105 @@ def process_json_plan(
...
+            "description": llm_op_description,
+            "input": [0],
+            "field": "text_representation",
+            "question": llm_op_question,
+        }
+        postprocessed_plan.insert(1, llm_op)
+
+        # shift the node ids for later operators, and modify the "inputs"
+        for step in postprocessed_plan[2:]:
+            step["node_id"] += 1
```
mdwelsh, line 224: Since node IDs don't matter and don't need to be in sequence, why not just generate a new node ID that is the max of the existing node IDs plus 1, and use that. That way we don't need to tweak the others.

lib/sycamore/sycamore/tests/unit/query/test_postprocess_plan.py
```diff
@@ -0,0 +1,101 @@
...
+            "description": "Generate an English response to the question",
+            "question": "How many Piper aircrafts were involved in accidents?",
+            "input": [1],
+            "node_id": 2,
+        },
+    ]
+
+
+def test_postprocess_plan():
+    # llm_client = OpenAI(OpenAIModels.GPT_4O.value)
```
mdwelsh, line 66: Remove this line.

lib/sycamore/sycamore/query/planner.py
baitsguy: it's just changing the description which can say this is a vector search, not the actual node object

apps/query-ui/queryui/Sycamore_Query.py
amolvdeshpande: Correct. If I understand the flow here correctly, query_id wouldn't be generated in this case.

lib/sycamore/sycamore/query/planner.py
amolvdeshpande: As Vinayak said, we need to convert the original description like "find all incidents that were caused by environmental conditions" to "find all incidents" -- assuming we want the description to be correct, an LLM would have to be used to manipulate the text properly..

lib/sycamore/sycamore/query/planner.py
amolvdeshpande: Same reason as above. The query phrase is something like "wildfires" -- needs to be converted to a boolean question like "Does this incident involve wildfires" for the LLM to answer.

lib/sycamore/sycamore/query/planner.py
amolvdeshpande: I feel like them being in sequence might be useful later (e.g., when we are drawing the plans or something like that) and it's not tricky to do it. But I can change it.

lib/sycamore/sycamore/tests/unit/query/test_postprocess_plan.py
amolvdeshpande: Done.

baitsguy (approved):

lib/sycamore/sycamore/query/planner.py
```diff
@@ -81,6 +80,104 @@ def process_json_plan(json_plan: str) -> LogicalPlan:
...
+                    to remove any functionality that specifically has to do with "{op.query_phrase}", thereby 
+                    generalizing the description to be more flexible.
+                    Return only the modified description. Do not make assumptions
+                    about the intent of the question that are not explicitly specified.
+                    {op.description}
+                    """,
+            llm_client,
+        )
+
+        new_op = QueryDatabase.model_validate(
```
baitsguy, line 126: curious why you're using model_validate and a dict as opposed to just create a regular object directly?

amolvdeshpande: Primarily to stay consistent with how they are being created in the other code right now and not duplicate functionality. Also, I didn't see an appropriate __init__ function, but perhaps it is being created implicitly? I can change it to direct object creation if that's the more appropriate way to do it.

baitsguy: You should just be able to initialize them like so: https://github.com/aryn-ai/sycamore/blob/0e2a9f5e586474c15bbfd1f85cf8b3bc3a1e4543/lib/sycamore/sycamore/tests/unit/query/test_planner.py#L68

I had suggested since we're creating the object ourselves as opposed to trying to create one from another data source (like the llm giving us a json).

Don't block on it though, it's your call if trivial to do

PR #884, opened 2024-10-08
Enable document model in sycamore.query + query-ui improvements

baitsguy:
A few changes all over to support doc model and usability for sycamore query.

1. Use document model in sycamore.query by enabling `reconstruct_document` in db scans and `use_elements` in llm map ops
2. Allow custom properties in `extract_entity` when `use_elements` is True
3. Write LogicalPlan in SycamoreExecutor when `trace_dir` provided
4. Query UI updates to support local mode, view query plan in TraceViewer. Local mode will only work for query plans that are runnable on `local_mode`

**note**: The queries have only been manually tested so it's possible we'll have to make more changes. Might need UI changes if we want to show/highlight elements within a doc that passed a filter/opensearch query


apps/query-ui/queryui/Sycamore_Query.py
```diff
@@ -141,11 +150,13 @@ def main():
...
         with st.form("query_form"):
             st.text_input("Query", key="query")
-            col1, col2 = st.columns(2)
+            col1, col2, col3 = st.columns(3)
             with col1:
                 submitted = st.form_submit_button("Run query")
             with col2:
                 st.toggle("Plan only", key="plan_only", value=False)
+            with col3:
+                st.toggle("Local mode", key="local_mode", value=False)
```
mdwelsh, line 159: I think "local mode" is a misnomer here since technically everything is local, so maybe change this to "Use Ray" and invert the meaning of the flag.

apps/query-ui/queryui/main.py
```diff
@@ -83,8 +84,11 @@ def main():
             trace_dir = args.trace_dir
         cmdline_args.extend(["--trace-dir", trace_dir])
 
-    ray_init()
-
+    if args.exec_mode == "ray":
```
mdwelsh, line 87: Maybe just `--exec_mode [ray|local]` for this flag, since I don't think we have a need for running without `--external-ray` anymore.

apps/query-ui/queryui/test_configuration.py
```diff
@@ -7,4 +7,4 @@ def test_configuration_sha():
         sha = sha256(bytes).hexdigest()
         # If the change was intentional, update the hash
         # Think about whether you have to make the change since everyone with a custom config will need to update it
-        assert sha == "0d5a24a1edfb9e523814dc847fa34cdcde2d2e78aff03e3f4489755e12be2c54", f"hash mismatch got {sha}"
+        assert sha == "cf89894116604d4b002f2c5b6c9acf25982bf764310a9a50827608dcdc6b1b2c", f"hash mismatch got {sha}"
```
mdwelsh, line 10: Can we get rid of this hack now? This seems like it was a useful short-term thing for Pfizer but I am not sure it is needed anymore.

lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -257,12 +258,24 @@ def get_code_string(self):
 """
         return result
 
+    def _write_metadata_to_trace_dir(self, plan: LogicalPlan, query_id: str):
+        assert self.trace_dir is not None, "Writing metadata requires trace_dir to be set"
+        path = os.path.join(self.trace_dir, f"{query_id}/metadata/")
+        os.makedirs(path, exist_ok=True)
+        with open(os.path.join(path, "query_plan.pickle"), "wb") as f:
```
mdwelsh, line 265: Could we store this as JSON, please? Pickle files are so unfriendly, and are brittle because they depend on a certain version of the Python classes being available. Given that we are already using PyDantic for query plans we should leverage it here instead of using pickle.

apps/query-ui/queryui/util.py
```diff
@@ -242,30 +243,73 @@ def readdata(self):
...
+class QueryMetadataTrace:
+    """Helper class to read and display metadata about a query."""
+
+    def __init__(self, metadata_dir: str):
+        self.metadata_dir = metadata_dir
+        self.query_plan = None
+        self.readdata()
+
+    def readdata(self):
+        f = os.path.join(self.metadata_dir, "query_plan.pickle")
```
mdwelsh, line 271: See below, let's use Pydantic + JSON and not pickle.

apps/query-ui/queryui/util.py
```diff
@@ -242,30 +243,73 @@ def readdata(self):
...
     """Helper class used to read and display query traces."""
 
     def __init__(self, trace_dir: str):
         self.trace_dir = trace_dir
-        self.node_traces = [QueryNodeTrace(trace_dir, node_id) for node_id in sorted(os.listdir(self.trace_dir))]
+        self.node_traces = []
+        for dir in sorted(os.listdir(self.trace_dir)):
+            if "metadata" not in dir:
+                self.node_traces += [QueryNodeTrace(trace_dir, dir)]
+            self.metadata = QueryMetadataTrace(self.trace_dir + "/" + dir)
```
mdwelsh, line 301: I don't think we need this extra class and indirection here. It's not a 'trace', and 'metadata' as a term is so heavily overused in our entire codebase that I would prefer to avoid it if possible :-)

Could you just have this function look for and read the file `query_plan.json` and set `self.query_plan` to the resulting object? We also don't need the `node_descriptions` map, I think, since it's easy enough to pluck the node we care about out of the `self.query_plan.nodes` map.

apps/query-ui/queryui/util.py
```diff
@@ -242,30 +243,73 @@ def readdata(self):
...
+        self.query_plan = None
+        self.readdata()
+
+    def readdata(self):
+        f = os.path.join(self.metadata_dir, "query_plan.pickle")
+        if os.path.isfile(f):
+            with open(f, "rb") as file:
+                self.query_plan = pickle.load(file)
+
+    def get_node_to_description(self) -> dict[str, str]:
```
mdwelsh, line 276: I don't think we need this; one can just access `query_plan.nodes` directly.

lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -257,12 +258,24 @@ def get_code_string(self):
...
+            pickle.dump(plan, f)
+
     def execute(self, plan: LogicalPlan, query_id: Optional[str] = None) -> Any:
         try:
             """Execute a logical plan using Sycamore."""
             if not query_id:
                 query_id = str(uuid.uuid4())
             bind_contextvars(query_id=query_id)
+
+            log.info("Writing query metadata")
```
mdwelsh, line 275: Say `metadata` one more time ...

Let's call this just `query_plan.json`. I don't think we need to keep using the term `metadata`.

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -245,7 +251,7 @@ def execute(self) -> Any:
                 context=self.context, val_key="llm", param_names=[OperationTypes.BINARY_CLASSIFIER.value]
             ),
             LLM,
-        ), "LLMFilter requies an 'llm' configured on the Context"
+        ), "LLMFilter requires an 'llm' configured on the Context"
```
mdwelsh, line 254:
```suggestion
        ), "SyamoreLlmFilter requires an 'llm' configured on the Context"
```

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -223,7 +228,8 @@ def script(self, input_var: Optional[str] = None, output_var: Optional[str] = No
 class SycamoreLlmFilter(SycamoreOperator):
     """
     Use an LLM to filter records on a Docset.
-    Args:
+    If trying to filter on a documents text contents, i.e. field == text_representation, we run the llm_filter
```
mdwelsh, line 231:
```suggestion
    If field == text_representation, the filter is run
```

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -538,6 +546,8 @@ def script(self, input_var: Optional[str] = None, output_var: Optional[str] = No
 
 class SycamoreTopK(SycamoreOperator):
     """
+    Note: top_k clustering only operators on document level fields. If you try to cluster on the text contents it
```
mdwelsh, line 549:
```suggestion
    Note: top_k clustering only operates on document level fields. If you try to cluster on the text contents it
```

mdwelsh, line 549: This comment is confusing. Should we just disallow `text_representation` to be passed as a the field to cluster on here?

apps/query-ui/queryui/Sycamore_Query.py
baitsguy: the sycamore exec mode is local so it makes sense I think? happy to change if you feel strongly

apps/query-ui/queryui/main.py
baitsguy: Ack, I'll get rid of the 'non-external' ray version

lib/sycamore/sycamore/query/execution/sycamore_executor.py
baitsguy: I had some serialization issues with json references in the past but I think that was before the pydantic refactoring so maybe it works now, let me give it a shot

lib/sycamore/sycamore/query/execution/sycamore_executor.py
baitsguy:
```
{
  "metadata": {
    "comment":
      {"response": "ack"}
    }
}
```

lib/sycamore/sycamore/query/execution/sycamore_operator.py
baitsguy: I can add a validation sure

baitsguy: Yeah let's disable text_representation here, we need to split this operator anyway

lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -257,12 +257,24 @@ def get_code_string(self):
 """
         return result
 
+    def _write_query_plan_to_trace_dir(self, plan: LogicalPlan, query_id: str):
+        assert self.trace_dir is not None, "Writing query_plan requires trace_dir to be set"
+        path = os.path.join(self.trace_dir, f"{query_id}/metadata/")
+        os.makedirs(path, exist_ok=True)
+        with open(os.path.join(path, "query_plan.json"), "w") as f:
+            f.write(plan.json())
```
mdwelsh, line 265: I think you want `model_dump_json()` here, right? `json()` was for Pydantic v1 (it still exists but I think it is deprecated).

apps/query-ui/queryui/Sycamore_Query.py
baitsguy: Cool I'll update the UI visuals to say "Use Ray" but let the internal/api continue to use sycamore terminology

apps/query-ui/queryui/util.py
baitsguy: Done

apps/query-ui/queryui/util.py
```diff
@@ -261,11 +268,31 @@ class QueryTrace:
...
-        self.node_traces = [QueryNodeTrace(trace_dir, node_id) for node_id in sorted(os.listdir(self.trace_dir))]
+        self.node_traces = []
+        self.query_plan = self._get_query_plan(self.trace_dir)
+        for dir in sorted(os.listdir(self.trace_dir)):
+            if "metadata" not in dir:
+                self.node_traces += [QueryNodeTrace(trace_dir, dir)]
+
+    def _get_query_plan(self, trace_dir: str):
+        metadata_dir = os.path.join(trace_dir, "metadata")
+        if os.path.isdir(metadata_dir):
```
mdwelsh, line 279: Nit: I think you can just do this in one step, i.e.,

`if os.path.isfile(os.path.join(trace_dir, "metadata", "query_plan.json")):`

lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -257,12 +257,24 @@ def get_code_string(self):
 """
         return result
 
+    def _write_query_plan_to_trace_dir(self, plan: LogicalPlan, query_id: str):
+        assert self.trace_dir is not None, "Writing query_plan requires trace_dir to be set"
+        path = os.path.join(self.trace_dir, f"{query_id}/metadata/")
```
mdwelsh, line 262: Since we're using `os.path.join` already please do

`path = os.path.join(self.trace_dir, query_id, "metadata")`

instead.

lib/sycamore/sycamore/tests/unit/query/test_plan.py
```diff
@@ -212,5 +212,5 @@ def test_compare_plans_structure_changed(llm_filter_plan):
     assert diff[0].diff_type == LogicalNodeDiffType.PLAN_STRUCTURE
     assert isinstance(diff[0].node_a, LlmFilter)
     assert isinstance(diff[0].node_b, LlmFilter)
-    assert len(diff[0].node_a.downstream_nodes) == 1
-    assert len(diff[0].node_b.downstream_nodes) == 0
+    assert len(diff[0].node_a.get_downstream_nodes()) == 1
```
mdwelsh, line 215: Not sure if this is the right place, but do we want to test that the values of `get_downstream_nodes()`/`get_dependencies()` are what we expect, and also that the `.dependencies` and `.downstream_nodes` properties return what we expect?

mdwelsh (approved):
LGTM - thanks!

lib/sycamore/sycamore/tests/unit/query/test_plan.py
baitsguy: Let's discuss offline, I hadn't done this initially because I didn't want to make assumptions on ids being stable. But we sort of already do that in other places so maybe it doesn't matter

baitsguy: I added a node_id assertion

PR #885, opened 2024-10-08
Add source_mode parameter to AutoMaterialize.

bsowell:
By passing in source_mode=MaterializeSourceMode.USE_STORED, this makes it possible to avoid re-execution of intermediate plans when using AutoMaterialize. The primary change is to avoid cleaning up the materialized data at the beginning of every execution when the source_mode is set to USE_STORED.

Also includes a utility for printing plans, which I found helpful when debugging.

### DIFF
--- a/lib/sycamore/sycamore/materialize.py
+++ b/lib/sycamore/sycamore/materialize.py
@@ -292,7 +292,7 @@ class AutoMaterialize(NodeTraverse):
 
        # The source_mode can be passed to AutoMaterialize, which will in turn be passed to all
        # created materialize nodes. To use each materialized node as a source:
-       ctx.rewrite_rules.append(AutoMaterialize(source_mode=MaterializeSourceMode.USE_STORED)
+       ctx.rewrite_rules.append(AutoMaterialize(source_mode=sycamore.MATERIALIZED_USE_STORED)
 
     Nodes in the plan will automatically be named. You can specify a name by defining it for the node:
        ctx = sycamore.init()
--- a/lib/sycamore/sycamore/plan_nodes.py
+++ b/lib/sycamore/sycamore/plan_nodes.py
@@ -220,16 +220,14 @@ def print_plan(node: Node, stream=sys.stdout) -> None:
             super().__init__()
             self.indent = 0
             self.stream = stream
-            self.before_fn = self._print_node
-            self.after_fn = self._after_node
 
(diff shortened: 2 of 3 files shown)


lib/sycamore/sycamore/materialize.py
```diff
@@ -286,14 +290,18 @@ class AutoMaterialize(NodeTraverse):
        a = AutoMaterialize({"root":Path|str, fs=pyarrow.fs, name=fn, clean=bool, tobin=fn})
        ctx.rewrite_rules.append(a)
 
+       # The source_mode can be passed to AutoMaterialize, which will in turn be passed to all
+       # created materialize nodes. To use each materialized node as a source:
+       ctx.rewrite_rules.append(AutoMaterialize(source_mode=MaterializeSourceMode.USE_STORED)
```
eric-anderson, line 295: `source_mode=sycamore.MATERIALIZE_USE_STORED`

both work, but that form involves less importing/typing
it can't be used for internal stuff because of circular dependencies.

lib/sycamore/sycamore/plan_nodes.py
```diff
@@ -199,3 +200,38 @@ def __init__(self, child: Node, **resource_args):
...
+      InternalNode2
+        LeafNode2
+    """
+
+    class PrintTraverse(NodeTraverse):
+        def __init__(self):
+            super().__init__()
+            self.indent = 0
+            self.stream = stream
+            self.before_fn = self._print_node
```
eric-anderson, line 223: this works, but you can remove the self.before_fn/after_fn setting and write the functions as:
```
def before(self, n: Node) -> Node:
   ... code for print_node ...

def after(self, n: Node) -> Node:
  ... code in after_node ...
```

lib/sycamore/sycamore/tests/unit/test_materialize.py
```diff
@@ -13,7 +13,7 @@
 import sycamore
 from sycamore.context import ExecMode
 from sycamore.data import Document, MetadataDocument
-from sycamore.materialize import AutoMaterialize, Materialize
+from sycamore.materialize import AutoMaterialize, Materialize, MaterializeSourceMode
```
eric-anderson, line 16: Here you can use sycamore.MATERIALIZE_USE_STORED. Less typing and no import.

eric-anderson (approved):

PR #886, opened 2024-10-08
add optimization from training development

HenryL27:
does the same as https://github.com/aryn-ai/managed-service/pull/560 but in sycamore

### DIFF
--- a/lib/sycamore/sycamore/transforms/table_structure/table_transformers.py
+++ b/lib/sycamore/sycamore/transforms/table_structure/table_transformers.py
@@ -144,7 +144,7 @@ def refine_rows(rows, tokens, score_threshold):
     """
 
     if len(tokens) > 0:
-        rows = nms_by_containment(rows, tokens, overlap_threshold=0.5)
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/transforms/table_structure/table_transformers.py
```diff
@@ -202,11 +205,21 @@ def nms_by_containment(container_objects, package_objects, overlap_threshold=0.5
 
 
 def slot_into_containers(
-    container_objects, package_objects, overlap_threshold=0.5, unique_assignment=True, forced_assignment=False
+    container_objects,
+    package_objects,
+    overlap_threshold=0.5,
+    unique_assignment=True,
+    forced_assignment=False,
+    _early_exit_vertical=False,  # yes. see docstring.
```
bsowell, line 213: why yes?

bsowell (approved):
LGTM.

lib/sycamore/sycamore/transforms/table_structure/table_transformers.py
HenryL27: bc when you read this you go "why?" to which the correct answer is "yes"

PR #887, opened 2024-10-09
Remove unused code (`_batchify`)

MarkLindblad:


alexaryn (approved):
Seems safe.  Ironically, we may be bringing back a batching iterator/generator in the future for the paddle stuff.

PR #889, opened 2024-10-09
Fix parallelism bug.

eric-anderson:
The aryn partitioner was the sole user of CompositeTransform which didn't respect parallelism settings on sub-operations.  Fix by explicitly applying the transform in CompositeTransform.

Add test that verifies different parallelism settings on sub-pieces of a composite transform work correctly.

Also verified that notebooks/sycamore-tutorial-intermediate-etl.ipynb worked correctly until the write to pinecone step.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ cache
 notebooks/default-prep-data
 .python-version
 apps/timetrace/ttviz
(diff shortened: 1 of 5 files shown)


bsowell (approved):
LGTM

PR #890, opened 2024-10-09
Fix PDFMiner Output Parsing

karanataryn:


MarkLindblad (approved):
This code only executes when partitioned using APS with Paddle OCR. It does not execute when partitioning locally since in that case `page_data` is a `PDFPage`.

PR #895, opened 2024-10-09
Allow passing custom ocr object to arynpartitioner

baitsguy:
Let's you pass in a configured object, the current implementation would allow you to configure the model

Tested with test

### DIFF
--- a/lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
+++ b/lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
@@ -1,3 +1,5 @@
+from unittest.mock import Mock
+
 from sycamore.data import Element
 from sycamore.transforms.detr_partitioner import ArynPDFPartitioner, DeformableDetr
 from sycamore.data import BoundingBox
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -542,16 +545,19 @@ def _run_text_extractor_document(
         use_cache: bool,
         use_ocr: bool,
         ocr_images: bool,
-        text_extractor_model: str,
+        text_extractor_model: Union[str, OcrModel],
```
karanataryn, line 548: Can we make this an instance of `TextExtractor`? I could imagine wanting to configure PDFMiner similarly as well.

lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
```diff
@@ -61,6 +64,17 @@ def test_partition(self):
         d = check_partition(s, TEST_DIR / "resources/data/pdfs/basic_table.pdf", use_ocr=True, use_cache=False)
         assert len(d) == 1
 
+    def test_partition_w_ocr_instance(self):
```
karanataryn, line 67: nit: `test_partition_with_ocr_instance`

karanataryn (approved):
LGTM. Have a few suggestions

lib/sycamore/sycamore/transforms/detr_partitioner.py
baitsguy: Sorry for the quick merge. Let's update this as the need arises, it shouldn't cause any issues

PR #898, opened 2024-10-10
Fix anonymous reading in materialize and add rate limited logging.

eric-anderson:
* In materialize, try reading using the credentials, but if it doesn't work, fall back to reading anonymously if that seems to be working.
* Add rate limited logging to reading via materialize in local mode.
* Check for no root before checking if a source since that makes more sense.
* switch ntsb_loader_materialized.py over to read in local mode, it was working (with the anonymous fix), but was very slow hence the logging.

### DIFF
--- a/lib/sycamore/sycamore/materialize.py
+++ b/lib/sycamore/sycamore/materialize.py
@@ -7,6 +7,7 @@ from sycamore.data import Document, MetadataDocument
 from sycamore.materialize_config import MaterializeSourceMode
 from sycamore.plan_nodes import Node, UnaryNode, NodeTraverse
 from sycamore.transforms.base import rename
+from sycamore.utils.sycamore_logger import LoggerFilter
 
 if TYPE_CHECKING:
     from ray import Dataset
@@ -14,6 +15,8 @@ if TYPE_CHECKING:
 
 
 logger = logging.getLogger(__name__)
+limited_logger = logging.getLogger(__name__ + ".limited")
+limited_logger.addFilter(LoggerFilter())
 
 
 class _PyArrowFsHelper:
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/materialize.py
```diff
@@ -91,17 +94,45 @@ def __init__(
...
+
     def prepare(self):
         """
         Clean up the materialize location if necessary.
         Validate that cleaning worked, but only once all materializes have finished cleaning.
         This protects against multiple materializes pointing to the same location.
         """
 
-        if self._will_be_source():
+        if self._root is None:
```
bsowell, line 132: Is there any meaning to this ordering change, or is it just to alphabetize?

lib/sycamore/sycamore/materialize.py
```diff
@@ -203,13 +234,21 @@ def local_source(self) -> list[Document]:
         logger.info(f"Using {self._orig_path} as cached source of data")
         if not self._fshelper.file_exists(self._success_path()):
             logging.warning(f"materialize.success not found in {self._orig_path}. Returning partial data")
+        from sycamore.utils.sycamore_logger import LoggerFilter
+
+        limited_logger = logging.getLogger(__name__ + ".limited_local_source")
+        limited_logger.addFilter(LoggerFilter())
```
bsowell, line 240: Is there a reason you define `limited_logger` at the top of the file and then redefine it here? Doesn't seem like you use it anywhere other than this method.

bsowell (approved):
Couple of minor questions, but otherwise LGTM.

lib/sycamore/sycamore/materialize.py
eric-anderson: Check for no root before checking if a source since that makes more sense.

lib/sycamore/sycamore/materialize.py
eric-anderson: Oops, it was only supposed to be down below. You need to have a separate limited logger for each thing you want to be separately rate limited.

PR #905, opened 2024-10-10
Refactor logical plan serialization.

mdwelsh:
This PR changes how we serialize and deserialize Luna LogicalPlans so everything uses the same set of Pydantic BaseModel subclasses.

* Removed LogicalOperator and folded its fields into Node.
* The LLM is instructed to generate serialized LogicalPlans directly, so there is no need for an additional JSON format.
* The `inputs` field of `Node` replaces `dependencies` and is a list of integers, always.
* There is a hidden `_input_nodes` field in Node that resolves `inputs` to the actual `Node` instances, with `input_nodes()` returning them.
* Use Pydantic model validator on `LogicalPlan` to ensure that `_input_nodes` are populated on nodes.
* Override `LogicalPlan` constructor to ensure that the correct duck-typed Node subclass is used, based on the `node_type` field.

Sorry for the length of this PR, but I think the result is more understandable and flexible.


baitsguy (approved):
mypy failing

Can you also validate the query-ui works and doesn't need updates

PR #908, opened 2024-10-11
Bugfix: Handle opensearch reader doc resconstruction when no parent doc in results

baitsguy:
^title

tested with test


lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -45,7 +45,7 @@ def read_records(self, query_params: BaseDBReader.QueryParams) -> "OpenSearchRea
         assert "index" not in query_params.kwargs and "body" not in query_params.kwargs
         if "scroll" not in query_params.kwargs:
             query_params.kwargs["scroll"] = "1m"
-        if "size" not in query_params.kwargs:
+        if "size" not in query_params.query:
```
mdwelsh, line 48: Scratching my head here. Why are you checking it in `.query` but setting it in `.kwargs` ?

mdwelsh (approved):

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
baitsguy: The first is trying to see if the user provided the size param in the query itself, the latter is trying to pass in a default value without modifying the query. I don't love it but I'm scared to touch it right now

PR #909, opened 2024-10-11
Add text reprentation for empty tables

dhruvkaliraman7:


HenryL27 (approved):
lgtm. do we ever get zero table cells if we have non zero structures?

dhruvkaliraman7:
> lgtm. do we ever get zero table cells if we have non zero structures?

yup, structure is gonna say here's a table which has 0 rows and n columns, and then post processing will say that's nuts, and chuck out the entire table

PR #910, opened 2024-10-11
Enable copying of the hash context.

alexaryn:


lib/sycamore/sycamore/utils/cache.py
```diff
@@ -19,13 +19,19 @@ class HashContext:
     This is a wrapper class for the hash context as Python/mypy/IDE does not like accessing _Hash from hashlib
     """
 
-    def __init__(self, algorithm="sha256"):
-        self.hash_obj = hashlib.new(algorithm)
+    def __init__(self, /, algorithm="sha256", source=None) -> None:
```
karanataryn, line 22: Can you add type annotations?

alexaryn: Not really.  The algorithm can be any number of things, not just string.  The source is a hashlib object, but they don't have a unifying base class.  Basically a lot of duck typing in hashlib.

karanataryn: Alright, that's unfortunate.

karanataryn (approved):

PR #911, opened 2024-10-11
Fix bug in entity extraction.

eric-anderson:
Notebooks like default-prep-script.ipynb would fail because the wrong way of generating the prompt would be used.

Also add a test for entity extraction, it's incomplete, but it fails if the bug is not fixed.


lib/sycamore/sycamore/tests/unit/transforms/test_extract_entity.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def generate(self, prompt_kwargs):
+        assert isinstance(prompt_kwargs["prompt"], EntityExtractorFewShotGuidancePrompt)
+        assert prompt_kwargs["entity"] == "title"
+        assert prompt_kwargs["query"] == "ELEMENT 1: ExampleTitle\nELEMENT 2: Footer\n"
+        assert prompt_kwargs["examples"] == "An Example Template"
+        return "ExampleTitle"
+
+
+def test_extract_with_template():
```
baitsguy, line 18: can you consolidate these in https://github.com/aryn-ai/sycamore/blob/main/lib/sycamore/sycamore/tests/unit/transforms/test_entity_extraction.py

eric-anderson: Done, also renamed the file since the test file name should match the file being tested.

baitsguy (approved):

PR #914, opened 2024-10-11
Opensearch kwargs fix

baitsguy:
When using `@context_params`, all kwargs specificied in the global Context would get passed into the reader.opensearch method. Those were being forwarded into the opensearchpy client. Changed the parameter to explicitly accept query_kwargs. I don't quiet know why we have kwargs anyway but don't want to mess with that

### DIFF
--- a/lib/sycamore/sycamore/reader.py
+++ b/lib/sycamore/sycamore/reader.py
@@ -234,6 +234,7 @@ class DocSetReader:
                 i.e. by collecting all elements belong to a single parent document (parent_id). This requires OpenSearch
                 to be an index of docset.explode() type. Default to false.
             query_kwargs: (Optional) Parameters to configure the underlying OpenSearch search query.
+            **kwargs: (Optional) kwargs to pass undefined parameters around.
 
         Example:
             The following shows how to write to data into a OpenSearch Index, and read it back into a DocSet.
--- a/lib/sycamore/sycamore/reader.py
+++ b/lib/sycamore/sycamore/reader.py
@@ -1,4 +1,4 @@
-from typing import Optional, Union, Callable, Dict
+from typing import Optional, Union, Callable, Dict, Any
 from pathlib import Path
 
 from pandas import DataFrame
@@ -217,6 +217,7 @@ class DocSetReader:
         index_name: str,
         query: Optional[Dict] = None,
         reconstruct_document: bool = False,
+        query_kwargs: dict[str, Any] = {},
         **kwargs,
     ) -> DocSet:
         """
@@ -232,6 +233,7 @@ class DocSetReader:
             reconstruct_document: Used to decide whether the returned DocSet comprises reconstructed documents,
                 i.e. by collecting all elements belong to a single parent document (parent_id). This requires OpenSearch
                 to be an index of docset.explode() type. Default to false.
+            query_kwargs: (Optional) Parameters to configure the underlying OpenSearch search query.
 
         Example:
             The following shows how to write to data into a OpenSearch Index, and read it back into a DocSet.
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/tests/unit/connectors/opensearch/test_opensearch.py
```diff
@@ -420,7 +420,9 @@ def test_get_knn_query(self):
                     },
                     "index_name": "test_index",
                 },
-                "default": {"text_embedder": embedder},
+                "default": {
+                    "text_embedder": embedder,
```
karanataryn, line 424: Think this diff came up because of the trailing comma after embedder.

karanataryn, line 424: Think this came up because of the trailing comma after embedder. Could you remove it to minimize irrelevant diffs?

lib/sycamore/sycamore/tests/integration/query/test_query_opensearch.py
```diff
@@ -74,6 +75,31 @@ def test_single_query(self, setup_index):
...
+                        "ssl_show_warn": False,
+                        "timeout": 120,
+                    },
+                    "index_name": self.INDEX,
+                },
+                "default": {"llm": "some unrelated value"},
+            },
+        )
+        result = context.read.opensearch(
+            query={"query": {"match_all": {}}, "size": 1}, reconstruct_document=True
```
karanataryn, line 99: Where are you testing `query_kwargs`?

lib/sycamore/sycamore/reader.py
```diff
@@ -280,11 +282,11 @@ def opensearch(
         client_params = OpenSearchReaderClientParams(os_client_args=os_client_args)
         query_params = (
             OpenSearchReaderQueryParams(
-                index_name=index_name, query=query, reconstruct_document=reconstruct_document, kwargs=kwargs
+                index_name=index_name, query=query, reconstruct_document=reconstruct_document, kwargs=query_kwargs
```
karanataryn, line 285: Can you add a comment here as to why we've switched this? It is a bit inscrutable.

lib/sycamore/sycamore/reader.py
```diff
@@ -217,6 +217,7 @@ def opensearch(
         index_name: str,
         query: Optional[Dict] = None,
         reconstruct_document: bool = False,
+        query_kwargs: dict[str, Any] = {},
```
karanataryn, line 220: I'm a bit unsure why we would keep `kwargs` at all here since it isn't being used. It could be quite confusing to customers. Is there no way to filter `kwargs` to only pass valid values in? That might be an easier fix.

karanataryn (commented):
Few comments and suggestions, I'm not entirely sure this is the correct fix.

lib/sycamore/sycamore/tests/unit/connectors/opensearch/test_opensearch.py
baitsguy: black/ruff did this, not me

lib/sycamore/sycamore/reader.py
baitsguy: Discussed offline, I've added a docstring about query_kwargs above already

lib/sycamore/sycamore/tests/integration/query/test_query_opensearch.py
baitsguy: this isn't testing query_kwargs, I don't see any tests related to opensearch params

lib/sycamore/sycamore/tests/unit/connectors/opensearch/test_opensearch.py
karanataryn: Yes, I mean it did it because of the trailing comma.

baitsguy: Our convention seems to be have trailing commas after each item? At least from what I see - there's probably a mix.

baitsguy: I tried to remove the command and whitespace, it reverts to this in auto fix 🤷

karanataryn (approved):

lib/sycamore/sycamore/tests/unit/connectors/opensearch/test_opensearch.py
karanataryn: Unfortunate :/

PR #915, opened 2024-10-12
Make metadata-extraction.ipynb work

eric-anderson:
* Use materialize to save the work of re-partitioning which is >50% of notebook runtime.
* Use materialize to avoid re-computation from show/take/write
* Fix readers to use the override_num_blocks parameter instead of parallelism; that's the recommended parameter from ray, and using parallelism created confusion with the parallelism setting for ray execution
* Remove parallelism setting from the notebook. It was causing hangs.

### DIFF
--- a/lib/sycamore/sycamore/connectors/file/file_scan.py
+++ b/lib/sycamore/sycamore/connectors/file/file_scan.py
@@ -76,13 +76,15 @@ class FileScan(Scan):
         paths: Union[str, list[str]],
         *,
         filesystem: Optional[FileSystem] = None,
-        parallelism: Optional[int] = None,
+        parallelism: Optional[str] = None,
+        override_num_blocks: Optional[int] = None,
         **resource_args,
     ):
         super().__init__(**resource_args)
         self._paths = paths
         self._filesystem = filesystem
-        self.parallelism = parallelism
+        assert parallelism is None, "Use override_num_blocks; remove parameter after 2025-03-01"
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/connectors/file/file_scan.py
```diff
@@ -76,13 +76,15 @@ def __init__(
         paths: Union[str, list[str]],
         *,
         filesystem: Optional[FileSystem] = None,
-        parallelism: Optional[int] = None,
+        parallelism: Optional[str] = None,
+        override_num_blocks: Optional[int] = None,
```
bsowell, line 80: I almost wonder if we want to expose this at all. I get that it makes sense as an alternative to parallelism, but it is kind of ray specific and probably hard for people to use. Should we just leave it in `resources_args`?

bsowell (approved):
Code looks fine. One design question that I will leave to you.

lib/sycamore/sycamore/connectors/file/file_scan.py
jonfritz: +1

jonfritz (approved):

lib/sycamore/sycamore/connectors/file/file_scan.py
eric-anderson: We probably don't want to expose it, but at this point we have. I added a comment to consider removing it and updated the documentation to be more clear people probably don't want this.

PR #916, opened 2024-10-12
Update DuckDB Reader to Package Change

karanataryn:
Updates the DuckDB reader to work after a breaking change in the output.


bsowell (commented):
Can we have a test that fails without this update and succeeds with? Do the existing integ tests catch it?

bsowell (approved):

PR #919, opened 2024-10-13
Luna client local schema (take 2)

dtecuci:
Removed requirement for schema file to be json format as the schema is not json. Added more exception handling.


lib/sycamore/sycamore/query/client.py
```diff
@@ -346,22 +347,50 @@ def main():
...
+        parser.error("Either index or schema-file is required")
 
+    # query is required
     if not args.query:
         parser.error("Query is required")
 
-    schema = client.get_opensearch_schema(args.index)
+    client = SycamoreQueryClient(s3_cache_path=args.s3_cache_path, trace_dir=args.trace_dir, cache_dir=args.cache_dir)
+
+    # get schema (schema_file overrides index)
```
mdwelsh, line 360: I think we should just use YAML for this (like I am doing with the query-eval tool), or JSON, not Python literals. Is there a good reason not to?

dtecuci: changed file format to yaml in latest commit [9e7ac10](https://github.com/aryn-ai/sycamore/pull/919/commits/9e7ac105124c5ac1dba8b290cd8db24cf69e3770)

mdwelsh (approved):

PR #922, opened 2024-10-14
Make OpenSearchSchema a proper Pydantic model.

mdwelsh:
Not yet done but this is what I'm proposing to make the OpenSearchSchema more flexible.


baitsguy (approved):
Nice

PR #925, opened 2024-10-15
Add local mode support for json scan and json document scan

bohou-aryn:


lib/sycamore/sycamore/connectors/file/file_scan.py
```diff
@@ -321,5 +363,42 @@ def execute(self, **kwargs) -> "Dataset":
...
 
+    def local_source(self, **kwargs) -> list[Document]:
+        if isinstance(self._paths, str):
+            paths = [self._paths]
+        else:
+            paths = self._paths
+
+        documents = []
+
+        def process_file(info):
```
bsowell, line 374: I know the actual document parsing is somewhat different between the two json methods, but can we factor out the filesystem stuff? Seems common with `BinaryScan.local_source` as well, no?

lib/sycamore/sycamore/tests/unit/scans/test_file_scan.py
```diff
@@ -58,6 +67,17 @@ def test_json_scan_all_props(self):
         assert doc.properties["props"] == "propValue"
         assert isinstance(doc.properties["web-app"], dict)
 
+    def test_json_scan_all_props_local(self):
```
bsowell, line 70: Do we not have tests for JsonDocumentScan?

lib/sycamore/sycamore/connectors/file/file_scan.py
bohou-aryn: yes, make sense, would try to factor out as much as possible.

lib/sycamore/sycamore/tests/unit/scans/test_file_scan.py
bohou-aryn: no, we don't have. Test against the notebook having jsondocumentscan

lib/sycamore/sycamore/connectors/file/file_scan.py
```diff
@@ -164,51 +192,27 @@ def execute(self, **kwargs) -> "Dataset":
...
-        return documents
+        document = Document()
+        document.doc_id = str(uuid.uuid1())
+        document.type = self._binary_format
+        document.binary_representation = binary_data
+        document.properties["path"] = info.path
+        if "filetype" not in document.properties and self._binary_format is not None:
+            document.properties["filetype"] = self._file_mime_type()
+        if self._is_s3_scheme():
+            document.properties["path"] = "s3://" + info.path
```
bsowell, line 212: Is this s3 stuff needed needed in the other implementations of process_file?

bsowell (approved):

lib/sycamore/sycamore/connectors/file/file_scan.py
bohou-aryn: I keep it the same, json_scan is handled inside the _to_document; for json_document_scan, I assume it's whatever inside the json file.

PR #928, opened 2024-10-15
Bugfixes: DocumentSource enum serialization and missing element_id in old data

baitsguy:
1. **Allow missing element_id in similarity score source for backwards compatibility**

Data loaded in opensearch prior to https://github.com/aryn-ai/sycamore/commit/0a3ad992a471d68658d0bebc9d50b73b599d092f fails in similarity scoring because of missing element_id. 

I'm choosing to explicitly handle the missing value here rather than defaulting when loading from the index to prevent unintended behavior.

2. **Un-enum DocumentSource because it causes json serialization problems**
To serialize a Document object we'd need to use a custom json encoder because of this which is overkill for what is essential a class of constant strings

### DIFF
--- a/lib/sycamore/sycamore/data/document.py
+++ b/lib/sycamore/sycamore/data/document.py
@@ -1,6 +1,5 @@
 from collections import UserDict
 import json
-from enum import Enum
 from typing import Any, Optional
 import uuid
 
@@ -8,7 +7,7 @@ from sycamore.data import BoundingBox, Element
 from sycamore.data.element import create_element
 
 
-class DocumentSource(Enum):
+class DocumentSource:
     UNKNOWN = "UNKNOWN"
     DB_QUERY = "DB_QUERY"
     DOCUMENT_RECONSTRUCTION_RETRIEVAL = "DOCUMENT_RECONSTRUCTION_RETRIEVAL"
--- a/lib/sycamore/sycamore/transforms/similarity.py
+++ b/lib/sycamore/sycamore/transforms/similarity.py
@@ -63,12 +63,11 @@ class SimilarityScorer(ABC):
         doc_score = document.properties.get(score_property_name, float("-inf"))
         if score > doc_score:
             document.properties[score_property_name] = score
-            source = element.element_index
-            if source is None:
+            if element.element_index is None:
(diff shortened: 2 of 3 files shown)


lib/sycamore/sycamore/data/document.py
```diff
@@ -1,14 +1,13 @@
...
-from enum import Enum
 from typing import Any, Optional
 import uuid
 
 from sycamore.data import BoundingBox, Element
 from sycamore.data.element import create_element
 
 
-class DocumentSource(Enum):
+class DocumentSource:
```
mdwelsh, line 10: Can't you use this?

https://docs.pydantic.dev/2.0/usage/types/enums/

lib/sycamore/sycamore/transforms/similarity.py
```diff
@@ -30,9 +30,7 @@ class SimilarityScorer(ABC):
         ignore_doc_structure: Ignore Document model (Document->Elements) in a DocSet
     """
 
-    def __init__(
-        self, ignore_element_sources: Optional[list[DocumentSource]] = None, ignore_doc_structure: bool = False
-    ):
+    def __init__(self, ignore_element_sources: Optional[list[str]] = None, ignore_doc_structure: bool = False):
```
mdwelsh, line 33: I think you should still be able to use an enum, but if not, can you use `typing.Literal` here

from typing import Literal
ignore_element_sources: Optional[List[Literal[DocumentSource.UNKNOWN, ...]]]

mdwelsh (approved):

lib/sycamore/sycamore/data/document.py
baitsguy: Seems like it wants me to convert Document to a BaseModel too then?

lib/sycamore/sycamore/transforms/similarity.py
baitsguy: Does this mean listing every literal supported? It's really just a string filter so I don't think its that useful here

lib/sycamore/sycamore/data/document.py
baitsguy: I'm going to skip this for now until we have an actual need for an enum

PR #930, opened 2024-10-15
Fix bugs in Unflattening Data

karanataryn:
Fixes bugs in unflattening data and comparing documents and the arising broken tests and bugs in various connectors.


lib/sycamore/sycamore/connectors/common.py
```diff
@@ -88,49 +112,86 @@ def flatten_data(
...
                 else:
-                    current += "\\" + char
-                escape = False
-            elif char == "\\":
-                escape = True
-            elif char == separator:
+                    # Trailing backslash, treat it as literal backslash
+                    current += "\\"
+                    i += 1
+            elif i + len(separator) < len(key) and key[i : i + len(separator)] == separator:
```
HenryL27, line 138: pretty sure python just truncates if you ask for a slice that's too big. ie
[1, 2, 3, 4, 5][3:7] == [4, 5]

lib/sycamore/sycamore/connectors/common.py
```diff
@@ -88,49 +112,86 @@ def flatten_data(
...
-    result: dict[Any, Any] = {}
+def unflatten_data(data: dict[Any, Any], separator: str = ".") -> dict[Any, Any]:
+    """
+    Unflattens a dictionary with keys that contain separators into a nested dictionary. The separator can be escaped,
+    and if there are integer keys in the path, the result will be a list instead of a dictionary.
+    """
 
-    def parse_key(key: str) -> list:
-        # Handle escaped separator
+    def split_key(key: str, separator: str = ".") -> list[str]:
```
HenryL27, line 121: why not just use str.split(sep=...)?

lib/sycamore/sycamore/connectors/common.py
```diff
@@ -88,49 +112,86 @@ def flatten_data(
...
+        parts = split_key(flat_key, separator)
         current = result
         for i, part in enumerate(parts):
-            part_key: Union[str, int] = int(part) if part.isdigit() else part
+            # Determine whether the key part is an integer (for list indices)
+            key: Union[str, int]
+            try:
+                key = int(part)
+            except ValueError:
+                key = part
```
HenryL27, line 159: could prob do `if key.isnumeric()`

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
```diff
@@ -88,7 +89,8 @@ def bulk_action_generator():
                         "_index": target_params.index_name,
                         "_id": r.doc_id,
                         "properties": r.properties,
-                        "embeddings": r.embeddings,
+                        "embeddings": r.embedding,
```
HenryL27, line 92: hey can we use embedding (singular) in the db as well?

lib/sycamore/sycamore/connectors/pinecone/pinecone_writer.py
```diff
@@ -119,10 +119,10 @@ def from_doc(cls, document: Document, target_params: "BaseDBWriter.TargetParams"
         else:
             id = f"{document.parent_id}#{document.doc_id}"
         values = document.embedding
-        metadata = {
+        metadata: dict[str, Any] = {
             "type": document.type,
             "text_representation": document.text_representation,
-            "bbox": document.bbox.to_dict() if document.bbox else None,
+            "bbox": document.bbox.to_list() if document.bbox else None,
```
HenryL27, line 125: why list? this will make it show up in pc as
```
bbox.0: 325,
bbox.1: 14364,
bbox.2: 264,
bbox.3: 436245,
``` 
rather than 
```
bbox.x1: 136y4,
bbox.y1: 364,
bbox.x2: 2634,
bbox.y2: 2643
```

lib/sycamore/sycamore/connectors/weaviate/weaviate_reader.py
```diff
@@ -92,6 +92,9 @@ def to_docs(self, query_params: "BaseDBReader.QueryParams") -> list[Document]:
                 | unflatten_data(dict(object.properties), "__")
                 | {"doc_id": str(object.uuid)}
             )  # type: ignore
+            doc.properties["parent_id"] = str(doc.properties.get("parent_id", "")).replace("-", "")
+            doc.properties["element_id"] = str(doc.properties.get("element_id", "")).replace("-", "")
+            doc.parent_id = str(doc.get("parent_id", ""))
```
HenryL27, line 97: maybe replace with underscore for reconstructability? Also should the doc.parent_id have a replace on it?

HenryL27: oh wait this is in the reader? Now I'm def confused how these things got here

lib/sycamore/sycamore/connectors/common.py
```diff
@@ -29,25 +32,46 @@ def generate_random_string(length=8):
...
         if k not in dict2:
             return False
-        if dict1[k] != dict2[k]:
+        if dict1[k] != dict2[k] and (dict1[k] or dict2[k]):
             return False
     return True
 
 
-def compare_docs(doc1, doc2):
+def compare_docs(doc1: Document, doc2: Document):
```
HenryL27, line 50: can I get a unit test or two for this fn?

lib/sycamore/sycamore/connectors/common.py
karanataryn: This allows us to ignore escaped characters. It's behavior that was present in the previous function and we have tests that check for this so didn't want to change it.

lib/sycamore/sycamore/connectors/common.py
```diff
@@ -29,25 +32,46 @@ def generate_random_string(length=8):
...
     filtered_doc1 = filter_doc(doc1, DEFAULT_RECORD_PROPERTIES.keys())
     filtered_doc2 = filter_doc(doc2, DEFAULT_RECORD_PROPERTIES.keys())
-    return filtered_doc1 == filtered_doc2
+    for key in filtered_doc1:
+        if (
+            isinstance(filtered_doc1[key], list)
+            or isinstance(filtered_doc1[key], np.ndarray)
```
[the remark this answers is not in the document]
karanataryn: Changed.

lib/sycamore/sycamore/connectors/common.py
karanataryn: Part will also be a `str` though? So I would have to cast it to make it work I think.

lib/sycamore/sycamore/connectors/common.py
karanataryn: Changed.

lib/sycamore/sycamore/connectors/common.py
karanataryn: Done.

lib/sycamore/sycamore/connectors/elasticsearch/elasticsearch_writer.py
karanataryn: Done.

lib/sycamore/sycamore/connectors/weaviate/weaviate_reader.py
karanataryn: Discussed offline.

lib/sycamore/sycamore/connectors/pinecone/pinecone_writer.py
karanataryn: Made the fix in the reader.

HenryL27 (approved):
lgtm

lib/sycamore/sycamore/connectors/common.py
HenryL27: isnumeric is a string method to tell you if all the characters are digits. so you could do
```python
if part.isnumeric():
    part = int(part)
```

karanataryn: Ah that makes sense. I can do that.

HenryL27: wait isdigit does the same thing. Why did we change this from the prev implementation?

karanataryn: This is because both `isdigit` and `isnumeric` don't handle decimals or negative numbers. I'll stick with the current implementation for now.

PR #931, opened 2024-10-15
Basic Bedrock LLM client.

mdwelsh:
This is a crude but serviceable LLM client that uses Bedrock. I have currently tested it only with Sonnet 3.5 and different models may have different requirements and parameters, but we can add those over time as needed.

Since there are so many different models with somewhat different formats, it's a little hard to test them all.

I lifted the cache logic into the `LLM` base class so it is reusable across subclasses, although the OpenAI variant needs to account for the response type in the cache key.

### DIFF
--- a/lib/sycamore/sycamore/llms/bedrock.py
+++ b/lib/sycamore/sycamore/llms/bedrock.py
@@ -1,0 +1,70 @@
+import boto3
+import json
+from sycamore.utils.cache import Cache
+from typing import Dict, Optional
+
+
+from sycamore.llms.llms import LLM
+
+DEFAULT_BEDROCK_MODEL = "anthropic.claude-3-5-sonnet-20240620-v1:0"
+DEFAULT_MAX_TOKENS = 1000
+DEFAULT_ANTHROPIC_VERSION = "bedrock-2023-05-31"
+
+
+class Bedrock(LLM):
+    """This is an LLM implementation that uses the AWS Bedrock API to generate text.
+
+    Args:
+        model_name: The name of the Bedrock model to use.
+        cache: A cache object to use for caching results.
+    """
+
+    def __init__(
+        self,
+        model_name: str = DEFAULT_BEDROCK_MODEL,
+        cache: Optional[Cache] = None,
+    ):
+        self._client = boto3.client(service_name="bedrock-runtime")
+        super().__init__(model_name, cache)
+
+    def is_chat_mode(self) -> bool:
+        """Returns True if the LLM is in chat mode, False otherwise."""
+        return True
+
+    def _get_generate_kwargs(self, prompt_kwargs: Dict, llm_kwargs: Optional[Dict] = None) -> Dict:
+        kwargs = {
+            **(llm_kwargs or {}),
+        }
+        if self._model_name.startswith("anthropic."):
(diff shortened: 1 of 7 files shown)


lib/sycamore/sycamore/llms/bedrock.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def is_chat_mode(self) -> bool:
+        """Returns True if the LLM is in chat mode, False otherwise."""
+        return True
+
+    def _get_generate_kwargs(self, prompt_kwargs: Dict, llm_kwargs: Optional[Dict] = None) -> Dict:
+        kwargs = {
+            **(llm_kwargs or {}),
+        }
+        if self._model_name.startswith("anthropic."):
```
baitsguy, line 38: where would one set the temp? We set openai to default 0

baitsguy (commented):
Can you add/extend an integ test that uses this for some level of testing

lib/sycamore/sycamore/llms/bedrock.py
```diff
@@ -0,0 +1,70 @@
...
+    """This is an LLM implementation that uses the AWS Bedrock API to generate text.
+
+    Args:
+        model_name: The name of the Bedrock model to use.
+        cache: A cache object to use for caching results.
+    """
+
+    def __init__(
+        self,
+        model_name: str = DEFAULT_BEDROCK_MODEL,
```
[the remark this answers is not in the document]
mdwelsh: Sure, done.

lib/sycamore/sycamore/llms/bedrock.py
mdwelsh: Updated to be consistent with the OpenAI behavior, though I don't know how I feel about defaulting to 0.

mdwelsh:
Thanks! Added integration test and addressed other comments. PTAL.

baitsguy:
Changes look good, open_ai integ test is failing tho

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -350,41 +365,15 @@ def is_chat_mode(self):
...
-        combined = {"prompt_kwargs": prompt_kwargs, "llm_kwargs": llm_kwargs, "model_name": self.model.name}
-        data = pickle.dumps(combined)
-        return self._cache.get_hash_context(data).hexdigest()
-
-    def _cache_get(self, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None):
-        if (llm_kwargs or {}).get("temperature", 0) != 0 or not self._cache:
-            # Never cache when temperature setting is nonzero.
-            return (None, None)
-
+        llm_kwargs = llm_kwargs.copy() if llm_kwargs else {}
```
baitsguy, line 370: Bug/bug reveal: The previous code was working because we were modifying llm_kwargs and serializing the `response_type` value in _get_cache_key. Now that we're copying, the llm_kwargs still has the unserialized value and that fails to pickle. If you drop this line it should work but you probably want a better solution

mdwelsh: Fixed.

baitsguy (approved):

PR #933, opened 2024-10-16
Fixup notebook tests

eric-anderson:
* Add script to run tests that will
  * optionally run in both local and ray mode
  * split between fast and slow tests
  * verify all notebooks are accounted
* Fix file scans so they can run with invalid AWS credentials in environment, same trick as for materialize, try being anonymous if default doesn't work.
* Fix ndd_example notebook to not specify parallelism.
* Fix tutorial notebook to use materialize to avoid some re-execution


lib/sycamore/sycamore/connectors/file/file_scan.py
```diff
@@ -92,6 +99,28 @@ def _is_s3_scheme(self) -> bool:
         else:
             return all(path.startswith("s3:") for path in self._paths)
 
+    def _try_infer_fs(self):
```
bsowell, line 102: This seems related/overlapping with https://github.com/aryn-ai/sycamore/pull/925. Are they going to conflict?

bsowell (approved):
Fine with this as long as you coordinate with Bohou on the file scan stuff.

PR #935, opened 2024-10-16
Support LLM selection in Sycamore Query Client.

mdwelsh:
This PR adds support for selecting an LLM by name in Sycamore Query Client, and adding it as an `--llm` argument to `client.py` and `apps/queryeval`.

To make this flexible I added a function `get_llm` which takes in a string model identifier and returns a function that can be called to create the model instance.

I also renamed `s3_cache_path` to `llm_cache_dir` to be consistent with the query result cache.


lib/sycamore/sycamore/query/client.py
```diff
@@ -245,18 +251,33 @@ def default_text_embedder():
         return SentenceTransformerEmbedder(batch_size=100, model_name="sentence-transformers/all-MiniLM-L6-v2")
 
     @staticmethod
-    def _get_default_context(s3_cache_path, os_client_args, sycamore_exec_mode) -> Context:
+    def _get_default_context(
+        llm_cache_dir: Optional[str],
+        os_client_args: Optional[dict],
+        sycamore_exec_mode: ExecMode,
+        llm: Optional[Union[str, LLM]],
```
baitsguy, line 258: this is fine for now, but using a bigboy llm for the record level operators will probably break

baitsguy (approved):

lib/sycamore/sycamore/query/client.py
mdwelsh: Break or just be slow and expensive?

PR #936, opened 2024-10-16
Evaluate query plans in query-eval

baitsguy:
Tested with good and bad plan


apps/query-eval/queryeval/driver.py
```diff
@@ -255,12 +255,30 @@ def do_eval(self, _query: QueryEvalQuery, result: QueryEvalResult) -> QueryEvalR
...
 
+        # Evalute query plans
+        if not _query.expected_plan:
+            console.print("[yellow]:construction: No expected query plan found, skipping.. ")
+        elif not result.plan:
+            console.print("[yellow]:construction: No computed query plan found, skipping.. ")
+        else:
+            plan_diff = _query.expected_plan.compare(result.plan)
+            if len(plan_diff) == 0:
+                console.print("[green]\U00002714 Plan correct")
```
mdwelsh, line 266: You wild man

(Python supports unicode directly in string literals so you can just use that :-)

apps/query-eval/queryeval/driver.py
```diff
@@ -255,12 +255,30 @@ def do_eval(self, _query: QueryEvalQuery, result: QueryEvalResult) -> QueryEvalR
...
+            else:
+                console.print("[red]:x: Plan incorrect")
```
mdwelsh, line 268: Maybe "Plan mismatch" (and "Plan match") since "correct" is of course dependent on many things...

mdwelsh (approved):

apps/query-eval/queryeval/driver.py
baitsguy: Fine, what's the fun in that though

apps/query-eval/queryeval/driver.py
baitsguy: Done

PR #940, opened 2024-10-17
Add plan correctness metrics summary + K in TopK optional

baitsguy:
1. Tested on local results files
2. Update TopK operator to be consistent with docset.top_k

### DIFF
--- a/apps/query-eval/queryeval/driver.py
+++ b/apps/query-eval/queryeval/driver.py
@@ -334,8 +334,8 @@ class QueryEvalDriver:
         )
         console.print(f"Avg. plan correctness: {average_plan_correctness}")
         console.print(
-            f"Avg. plan diff count: {sum(result.metrics.plan_diff_count for result in self.results_map.values()) 
-                                     / len(self.results_map)}"
+            "Avg. plan diff count: "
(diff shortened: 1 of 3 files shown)


apps/query-eval/queryeval/driver.py
```diff
@@ -279,11 +281,16 @@ def do_eval(self, _query: QueryEvalQuery, result: QueryEvalResult) -> QueryEvalR
                     console.print(f"Expected node: {diff.node_a!r}")
                     console.print(f"Actual node: [red]{diff.node_b!r}")
                     console.print()
+                metrics.plan_similarity = (len(_query.expected_plan.nodes) - len(plan_diff)) / len(
```
mdwelsh, line 284: Usually I think it's best to capture raw counts rather than aggregate metrics like a ratio, since the ratio can be hard to interpret in some cases (i.e., if a query plan only has one node). Could you capture the number of diffs instead (maybe of of each type)?

mdwelsh (approved):
LGTM but I have a suggestion

apps/query-eval/queryeval/driver.py
baitsguy: Let me add both yeah

PR #941, opened 2024-10-17
Support SummarizeImages with non-OpenAI LLMs.

bsowell:
This commit adds a generic LLMImageSummarizer that takes an LLM instance. In order to account for differences in image formats accepted by the LLMs, I also added a `format_image` method to the LLM interface. Finally, I fixed a serialization issue in the Bedrock class by not serializing the boto client.

### DIFF
--- a/lib/sycamore/sycamore/llms/bedrock.py
+++ b/lib/sycamore/sycamore/llms/bedrock.py
@@ -2,11 +2,13 @@ from dataclasses import dataclass
 from enum import Enum
 import boto3
 import json
-from typing import Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
+from PIL import Image
 
 from sycamore.llms.llms import LLM
 from sycamore.utils.cache import Cache
+from sycamore.utils.image_utils import base64_data
 
 DEFAULT_MAX_TOKENS = 1000
 DEFAULT_ANTHROPIC_VERSION = "bedrock-2023-05-31"
@@ -48,6 +50,8 @@ class Bedrock(LLM):
         model_name: Union[BedrockModels, str],
         cache: Optional[Cache] = None,
     ):
+        self.model_name = model_name
+
         if isinstance(model_name, BedrockModels):
             self.model = model_name.value
         elif isinstance(model_name, str):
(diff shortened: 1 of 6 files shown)


lib/sycamore/sycamore/llms/bedrock.py
```diff
@@ -56,10 +60,20 @@ def __init__(
...
+            return Bedrock(**kwargs)
+
+        kwargs = {"model_name": self.model_name, "cache": self._cache}
+        return deserializer, (kwargs,)
+
     def is_chat_mode(self) -> bool:
         """Returns True if the LLM is in chat mode, False otherwise."""
         return True
 
+    def format_image(self, image: Image.Image) -> dict[str, Any]:
```
mdwelsh, line 74: Is this true for all Bedrock models or only Anthropic? If you can only support anthropic models here just throw an error if the model_name does not indicate an anthropic model.

lib/sycamore/sycamore/tests/integration/transforms/test_summarize_images.py
```diff
@@ -17,5 +18,28 @@ def test_summarize_images():
...
+    image_docs = (
+        context.read.binary(paths=[str(path)], binary_format="pdf")
+        .partition(ArynPartitioner(extract_images=True, use_partitioning_service=False, use_cache=False))
+        .transform(SummarizeImages, summarizer=LLMImageSummarizer(llm=llm))
+        .explode()
+        .filter(lambda d: d.type == "Image")
+        .take_all()
+    )
+
+    print(f"result: {image_docs[0].properties['summary']}")
```
mdwelsh, line 42: Remove this line

lib/sycamore/sycamore/tests/integration/transforms/test_summarize_images.py
```diff
@@ -1,7 +1,8 @@
 import sycamore
 from sycamore.transforms.partition import ArynPartitioner
-from sycamore.transforms.summarize_images import SummarizeImages
+from sycamore.transforms.summarize_images import LLMImageSummarizer, SummarizeImages
 from sycamore.tests.config import TEST_DIR
+from sycamore.llms.bedrock import BedrockModels, Bedrock
 
 
 def test_summarize_images():
```
mdwelsh, line 8: Mind making this explicitly `test_summarize_images_openai` and suggest explicitly passing `llm=OpenAI...`

lib/sycamore/sycamore/tests/integration/transforms/test_summarize_images.py
```diff
@@ -17,5 +18,28 @@ def test_summarize_images():
         .take_all()
     )
 
+    print(f"result: {image_docs[0].properties['summary']}")
```
mdwelsh, line 21: Remove this line

mdwelsh (approved):
Nice!

lib/sycamore/sycamore/llms/bedrock.py
bsowell: Good call. Will do.

PR #944, opened 2024-10-17
Update standardizer.py

jonfritz:
Adding a new class called ignore_errors to enable the standardizers to gracefully continue when encountering null key:value pairs

### DIFF
--- a/lib/sycamore/sycamore/transforms/standardizer.py
+++ b/lib/sycamore/sycamore/transforms/standardizer.py
@@ -298,7 +298,7 @@ def ignore_errors(doc, standardizer, key_path: list[str]):
     Example:
         .. code-block:: python
 
-            docset.map(lambda doc: IgnoreErrors.ignore_errors(doc, DateTimeStandardizer, ["properties", "entity", "dateAndTime"])
+            docset.map(lambda doc: ignore_errors(doc, DateTimeStandardizer, ["properties", "entity", "dateAndTime"])
     """
     
     try:
--- a/lib/sycamore/sycamore/transforms/standardizer.py
+++ b/lib/sycamore/sycamore/transforms/standardizer.py
@@ -8,6 +8,9 @@ from sycamore.plan_nodes import Node
 from sycamore.data import Document
 from sycamore.transforms.map import Map
 
+import logging
+logger = logging.getLogger(__name__)
+
 
 class Standardizer(ABC):
     """
@@ -284,18 +287,27 @@ class StandardizeProperty(Map):
     ):
         super().__init__(child, f=standardizer.standardize, args=path, kwargs=kwargs)
 
-class IgnoreErrors(standardizer)
-
-   """
+def ignore_errors(doc, standardizer, key_path: list[str]):
+   
+    """
     A class for applying the behavior of a standardizer to log errors and continue when encountering null values.
 
     This class allows for the execution of standardization logic not to fail when encountering null key:value pairs. It will
     instead log a warning stating what key:value pairs in what documents were missing.
+
+    Example:
+        .. code-block:: python
+
+            docset.map(lambda doc: IgnoreErrors.ignore_errors(doc, DateTimeStandardizer, ["properties", "entity", "dateAndTime"])
     """
     
-    def ignore_errors(doc, standardizer, key_path: list[str]):
-        try:
-            doc = standardizer.standardize(doc, key_path=key_path)
-        except:
-            print(f"Key {key_path} not found in document: {doc}")
-        return doc
+    try:
+        doc = standardizer.standardize(doc, key_path=key_path)
+    except KeyError:
+        logger.warn(f"Key {key_path} not found in document: {doc}")
+    except Exception as e:
+        logger.error(e)
+    return doc
+
+
+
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -283,3 +283,19 @@ def __init__(
         **kwargs,
     ):
         super().__init__(child, f=standardizer.standardize, args=path, kwargs=kwargs)
+
+class IgnoreErrors(standardizer)
```
HenryL27, line 287: Technically this is not a standardizer, so no need to subclass it. Also colon at the end

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -283,3 +283,19 @@ def __init__(
...
+
+   """
```
HenryL27, line 289: indentation

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -283,3 +283,19 @@ def __init__(
...
+    A class for applying the behavior of a standardizer to log errors and continue when encountering null values.
+
+    This class allows for the execution of standardization logic not to fail when encountering null key:value pairs. It will
+    instead log a warning stating what key:value pairs in what documents were missing.
```
HenryL27, line 293: Add an example probably. 
```
"""
    Example:
        .. code-block:: python

            docset.map(lambda doc: IgnoreErrors.ignore_errors(doc, DateTimeStandardizer, ["properties", "entity", "dateAndTime"])
"""
```

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -283,3 +283,19 @@ def __init__(
...
+    """
+    
+    def ignore_errors(doc, standardizer, key_path: list[str]):
+        try:
+            doc = standardizer.standardize(doc, key_path=key_path)
+        except:
```
HenryL27, line 299: prob wanna `except KeyError` if you're only trying to catch those. If you're ok with any exception, then you prob want something like 
```python
except Exception as e:
    print(e)
```

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -283,3 +283,19 @@ def __init__(
...
+            print(f"Key {key_path} not found in document: {doc}")
```
HenryL27, line 300: prob want to use logging rather than print. I don't remember how to hook it correctly to ray tho

lib/sycamore/sycamore/transforms/standardizer.py
mdwelsh, line 287: This isn't quite the right syntax, so I don't think this will work.

@bsowell My guess is that we would like to have this logic be more generic, perhaps a subclass of `Map` that swallows errors from the function being applied?

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -283,3 +283,19 @@ def __init__(
...
+class IgnoreErrors(standardizer)
+
+   """
+    A class for applying the behavior of a standardizer to log errors and continue when encountering null values.
+
+    This class allows for the execution of standardization logic not to fail when encountering null key:value pairs. It will
+    instead log a warning stating what key:value pairs in what documents were missing.
+    """
+    
+    def ignore_errors(doc, standardizer, key_path: list[str]):
```
HenryL27, line 296: add `@staticmethod` one line above this

HenryL27 (commented):
It might be worth just making this a function rather than a full class - the main benefit you get from having classes with single static methods is that you can do interfaces and stuff. I don't imagine there will be any other things that follow the same contract as `IgnoreErrors` so it should probably just be a function (useable like `docset.map(lambda d: ignore_errors(d, standardizer, key_path))`). 

If you want to be really fancy (and actually I think this would be a good interface) you can make this a higher-order function - i.e. have it return a function that calls the standardizer and swallows errors. Then the usage can become `docset.map(ignore_errors(standardizer, key_path))`.

lib/sycamore/sycamore/transforms/standardizer.py
jonfritz: Ack

lib/sycamore/sycamore/transforms/standardizer.py
jonfritz: Ack

lib/sycamore/sycamore/transforms/standardizer.py
jonfritz: Ack

lib/sycamore/sycamore/transforms/standardizer.py
jonfritz: Ack

lib/sycamore/sycamore/transforms/standardizer.py
jonfritz: Ack

jonfritz: ack

lib/sycamore/sycamore/transforms/standardizer.py
jonfritz: Ack

lib/sycamore/sycamore/transforms/standardizer.py
jonfritz: Ack

HenryL27 (approved):
lgtm, just get the linting to work

PR #945, opened 2024-10-17
Create LLMSchemaExtractor and LLMPropertyExtractor classes.

bsowell:
We previously had OpenAI-specific versions of these, even though they weren't particularly tied to the LLM. This commit generalizes those classes and tweaks the Bedrock llm implementation to handle the basic guidance prompts so that it works with the existing implementation.

### DIFF
--- a/lib/sycamore/sycamore/transforms/extract_schema.py
+++ b/lib/sycamore/sycamore/transforms/extract_schema.py
@@ -42,7 +42,7 @@ class PropertyExtractor(ABC):
         pass
 
 
-class OpenAISchemaExtractor(SchemaExtractor):
+class LLMSchemaExtractor(SchemaExtractor):
     """
     OpenAISchema uses one of OpenAI's language model (LLM) for schema extraction,
     given a suggested entity type to be extracted.
@@ -110,7 +110,11 @@ class OpenAISchemaExtractor(SchemaExtractor):
         return entities
 
 
-class OpenAIPropertyExtractor(PropertyExtractor):
+class OpenAISchemaExtractor(LLMSchemaExtractor):
+    pass
+
+
+class LLMPropertyExtractor(PropertyExtractor):
     """
     OpenAISchema uses one of OpenAI's language model (LLM) to extract actual property values once
     a schema has been detected or provided.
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/transforms/extract_schema.py
```diff
@@ -219,6 +223,10 @@ def __init__(self, child: Node, schema_extractor: SchemaExtractor, **resource_ar
         super().__init__(child, f=schema_extractor.extract_schema, **resource_args)
 
 
+class OpenAIPropertyExtractor(LLMPropertyExtractor):
```
mdwelsh, line 226: Please add docstring and indicate this is for backwards-compatibility. Consider using the `deprecated` directive for doc generation.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

lib/sycamore/sycamore/transforms/extract_schema.py
```diff
@@ -110,7 +110,11 @@ def _handle_zero_shot_prompting(self, document: Document) -> Any:
         return entities
 
 
-class OpenAIPropertyExtractor(PropertyExtractor):
+class OpenAISchemaExtractor(LLMSchemaExtractor):
```
mdwelsh, line 113: Please add docstring and indicate this is for backwards-compatibility. Consider using the deprecated directive for doc generation.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

lib/sycamore/sycamore/transforms/extract_schema.py
```diff
@@ -42,7 +42,7 @@ def extract_properties(self, document: Document) -> Document:
         pass
 
 
-class OpenAISchemaExtractor(SchemaExtractor):
+class LLMSchemaExtractor(SchemaExtractor):
```
mdwelsh, line 45: I would recommend updating the unit tests to use the new LLMSchemaExtractor and LLMPropertyExtractor classes instead of the OpenAI-specific ones.

mdwelsh (approved):

PR #946, opened 2024-10-18
Add HeaderAugmenterMerger

dhruvkaliraman7:


lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,192 @@ def postprocess_element(self, elem: Element) -> Element:
         return elem
 
 
+class HeaderAugmenterMerger(ElementMerger):
+    """
+    The ``HeaderAugmenterMerger`` groups together different elements in a Document and enhances the text
+    representation of the elements by adding the preceeding section-header/title.
+
+    - It merges elements ("Text", "List-item", "Caption", "Footnote", "Formula" , "Page-footer", "Page-header").
```
MarkLindblad, line 619: Maybe you want to remove the extra space in `..."Formula" , "Page-footer"...`

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,192 @@ def postprocess_element(self, elem: Element) -> Element:
...
+        Args:
+            document (Document): A document with elements to be merged.
+
+        Returns:
+            Document: The same document, with its elements merged
+        """
+        if len(document.elements) < 2:
+            return document
+
+        firstHeader = False
```
MarkLindblad, line 653: To keep the naming convention consistent, can you name this `first_header` instead?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,192 @@ def postprocess_element(self, elem: Element) -> Element:
...
+
+        # DO NOT MERGE across pages
+        if (
+            not self.merge_across_pages
+            and element1.properties[DocumentPropertyTypes.PAGE_NUMBER]
+            != element2.properties[DocumentPropertyTypes.PAGE_NUMBER]
+        ):
+            return False
+
+        if element1.data["token_count"] + 1 + element2.data["token_count"] > self.max_tokens:
```
MarkLindblad, line 691: What if the next element is a Section-header and this if statement is true? Won't we overwrite the `_header` field of the upcoming Section-header?

dhruvkaliraman7: good catch

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: done

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: done

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
         return elem
 
 
+class HeaderAugmenterMerger(ElementMerger):
+    """
+    The ``HeaderAugmenterMerger`` groups together different elements in a Document and enhances the text
+    representation of the elements by adding the preceeding section-header/title.
+
+    - It merges elements ("Text", "List-item", "Caption", "Footnote", "Formula", "Page-footer", "Page-header").
```
MarkLindblad, line 619: Add the word 'certain' as in "It merges certain elements ...."

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+
+        to_merge = [self.preprocess_element(e) for e in document.elements]
+        new_elements = [to_merge[0]]
+        for element in to_merge[1:]:
+            if self.should_merge(new_elements[-1], element):
+                new_elements[-1] = self.merge(new_elements[-1], element)
+            else:
+                new_elements.append(element)
+        document.elements = [self.postprocess_element(e) for e in new_elements]
+        document.elements = [e for e in new_elements if e.type not in ["Section-header", "Title"]]
```
MarkLindblad, line 670: You can combine these two lines.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+        # deal with empty elements
+        if (
+            DocumentPropertyTypes.PAGE_NUMBER not in element1.properties
+            or DocumentPropertyTypes.PAGE_NUMBER not in element2.properties
+            or element1.type is None
+            or element2.type is None
+        ):
+            return False
+
+        # DO NOT MERGE across pages
```
MarkLindblad, line 683: I know this is what the other mergers say, but can you change this to say "conditionally prevent merging across pages"

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+            not self.merge_across_pages
+            and element1.properties[DocumentPropertyTypes.PAGE_NUMBER]
+            != element2.properties[DocumentPropertyTypes.PAGE_NUMBER]
+        ):
+            return False
+
+        if element1.data["token_count"] + 1 + element2.data["token_count"] > self.max_tokens and element2.type not in [
+            "Section-header",
+            "Title",
+        ]:
```
MarkLindblad, line 694: This if statement is checking the wrong thing. It should be checking the number of tokens in `element1["header"]`, not the number of tokens in what occasionally is `element1.text_representation`.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+        if element1.type in ["Section-header", "Title"] and element2.type in ["Section-header", "Title"]:
+            return True
+
+        # MERGE adjacent 'text' elements
+        small = {"Text", "List-item", "Caption", "Footnote", "Formula", "Page-footer", "Page-header"}
+        if (element1 is not None) and (element1.type in small) and (element2 is not None) and (element2.type in small):
+            return True
+
+        # Add header to next element
+        if element2.type not in ["Section-header", "Title"]:
```
MarkLindblad, line 714: Add a comment saying this is for pictures and tables.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+            Document: The same document, with its elements merged
+        """
+        if len(document.elements) < 2:
+            return document
+
+        first_header = False
+        for element in document.elements:
+            if element.type in ["Section-header", "Title"]:
+                first_header = True
+                element.data["_header"] = element.text_representation
```
MarkLindblad, line 657: Do we want to add a warning here if the `_header` is larger than the `max_tokens`?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+            if element2.text_representation:
+                if element2.data["_header"]:
+                    element2.text_representation = element2.data["_header"] + "\n" + element2.text_representation
+            else:
+                element2.text_representation = element2.data["_header"]
+        return False
+
+    def merge(self, elt1: Element, elt2: Element) -> Element:
+        """Merge two elements; the new element's fields will be set as:
+            - type: "Section-header" , "MergedText"
```
MarkLindblad, line 725: Can you remove the extra space in `"Section-header" , "MergedText"`?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+            - binary_representation: elt1.binary_representation + elt2.binary_representation
+            - text_representation: elt1.text_representation + elt2.text_representation
+            - bbox: the minimal bbox that contains both elt1's and elt2's bboxes
+            - properties: elt1's properties + any of elt2's properties that are not in elt1
+            note: if elt1 and elt2 have different values for the same property, we take elt1's value
+            note: if any input field is None we take the other element's field without merge logic
+
+        Args:
+            element1 (Tuple[Element, int]): the first element (and number of tokens in it)
+            element2 (Tuple[Element, int]): the second element (and number of tokens in it)
```
MarkLindblad, line 735: I know that this is what the other mergers say, but can you change this to the below?
```
Args:
    element1 (Element): the first element (numbers of tokens in it is stored by `preprocess_element` as element1["token_count"])
    element2 (Element): the first element (numbers of tokens in it is stored by `preprocess_element` as element2["token_count"])
```

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+
+        Returns:
+            Tuple[Element, int]: a new merged element from the inputs (and number of tokens in it)
```
MarkLindblad, line 738: This is not the return type. The return type is `Element`. Fix as above.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+
+        # Merge binary representations by concatenation
+        if elt1.binary_representation is None or elt2.binary_representation is None:
+            new_elt.binary_representation = elt1.binary_representation or elt2.binary_representation
+        else:
+            new_elt.binary_representation = elt1.binary_representation + elt2.binary_representation
+
+        # Merge text representations by concatenation with a newline
+
+        new_elt_text_representation = "\n".join(filter(None, [elt1.text_representation, elt2.text_representation]))
```
MarkLindblad, line 758: Remove line 757 (the extra newline).

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+        new_elt.text_representation = new_elt_text_representation if new_elt_text_representation else None
+        # new_elt.text_representation = elt1.text_representation + "\n" + elt2.text_representation
```
MarkLindblad, line 760: Remove this comment.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+        new_elt.data["token_count"] = tok1 + 1 + tok2
```
MarkLindblad, line 761: This line is wrong when only one of the two old elements has a `text_representation`.

MarkLindblad: You will need to refactor this to not use filter and use an if instead I believe.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -611,6 +611,195 @@ def postprocess_element(self, elem: Element) -> Element:
...
+            pass
+        elif elt1.bbox is None or elt2.bbox is None:
+            new_elt.bbox = elt1.bbox or elt2.bbox
+        else:
+            new_elt.bbox = BoundingBox(
+                elt1.bbox.x1,
+                elt1.bbox.y1,
+                elt1.bbox.x2,
+                elt1.bbox.y2,
+            )
```
MarkLindblad, line 774: Don't we want to use the `min` and `max` logic here? Like below:
```python
new_elt.bbox = BoundingBox(
                min(elt1.bbox.x1, elt2.bbox.x1),
                min(elt1.bbox.y1, elt2.bbox.y1),
                max(elt1.bbox.x2, elt2.bbox.x2),
                max(elt1.bbox.y2, elt2.bbox.y2),
            )
```

dhruvkaliraman7: I am unsure, what if it is extending multiple pages? We never made a decision about how the UI would look like (do we even show the chunked document?

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: Not use filter? Just gotta remove 1 from the token_count right?

lib/sycamore/sycamore/transforms/split_elements.py
```diff
@@ -34,12 +34,17 @@ def __init__(self, child: Node, tokenizer: Tokenizer, maximum: int, **kwargs):
     def split_doc(parent: Document, tokenizer: Tokenizer, max: int) -> Document:
         result = []
         for elem in parent.elements:
+            # Ensure the _header does not take up more than a third of the tokens
+            # Also avoid max resursive depth error
+            if elem.get("_header") and len(tokenizer.tokenize(elem["_header"])) / max > 0.33:
+                del elem["_header"]
```
MarkLindblad, line 40: Do we want to add a warning log message if the header is dropped?

lib/sycamore/sycamore/transforms/split_elements.py
```diff
@@ -34,12 +34,17 @@ def __init__(self, child: Node, tokenizer: Tokenizer, maximum: int, **kwargs):
...
             result.extend(SplitElements.split_one(elem, tokenizer, max))
         parent.elements = result
         return parent
 
     @staticmethod
     def split_one(elem: Element, tokenizer: Tokenizer, max: int) -> list[Element]:
+
         txt = elem.text_representation
         if not txt:
             return [elem]
```
MarkLindblad, line 50: If the element has a `_header` but no `text_representation`, its `_header` won't be added in this code path. Is this a possible scenario?

lib/sycamore/sycamore/transforms/split_elements.py
dhruvkaliraman7: Do you want to replace _header with "Over token limit" in this case or some unique str? would it be useful printing a warning without knowing which element it's for?

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: We are in essence removing section-header elements, so we are okay even if _header is greater than max_tokens. Split elements should take care of splitting the _header which would be part of the text_representation of some element. Lmk if you think there is a better way of doing this

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -614,6 +614,197 @@ def postprocess_element(self, elem: Element) -> Element:
...
+                else:
+                    element2.text_representation = element1.data["_header"]
+            return False
+
+        # Merge consecutive section headers/titles and save as a section-header element
+        if element1.type in ["Section-header", "Title"] and element2.type in ["Section-header", "Title"]:
+            return True
+
+        # MERGE adjacent 'text' elements
+        small = {"Text", "List-item", "Caption", "Footnote", "Formula", "Page-footer", "Page-header"}
```
MarkLindblad, line 713: Can you use `text_like` instead of `small`? Why is it `small`?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -614,6 +614,197 @@ def postprocess_element(self, elem: Element) -> Element:
...
+
+        # Add header to next element (images, tables)
+        if element2.type not in ["Section-header", "Title"]:
+            element2.data["_header"] = element1.data["_header"]
+            if element2.text_representation:
+                if element2.data["_header"]:
+                    element2.text_representation = element2.data["_header"] + "\n" + element2.text_representation
+            else:
+                element2.text_representation = element2.data["_header"]
+        return False
```
MarkLindblad, line 725: Do we want to rewrite the end to error on unknown element types or do we want to keep it as-is so that people can chain mergers (`GreedySectionMerger` creates elements of type `"Section-header+table"`)?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -614,6 +614,197 @@ def postprocess_element(self, elem: Element) -> Element:
...
+
+    def merge(self, elt1: Element, elt2: Element) -> Element:
+        """Merge two elements; the new element's fields will be set as:
+            - type: "Section-header", "MergedText"
```
MarkLindblad, line 729: This line is wrong. We never set type to `"MergedText"`.

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: I think I will add "Section" in the list of elements allowed to be merged(from greedytextelement merger). For your comment, let us keep as is, since augment header is only going to make the chunking better?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -614,6 +614,197 @@ def postprocess_element(self, elem: Element) -> Element:
...
+            return document
+
+        first_header = False
+        for element in document.elements:
+            if element.type in ["Section-header", "Title"]:
+                first_header = True
+                element.data["_header"] = element.text_representation
+            else:
+                if not first_header:
+                    element.data["_header"] = None
```
MarkLindblad, line 663: Isn't this `else` case unnecessary? We already have the situation where there will be no `_header` to copy when we break across pages. Please remove this.

MarkLindblad, line 663: Isn't this `else` case unecessary? Our logic handles missing `_header`s and we already have the situation of a missing `_header` when we break across pages and the first element on the new page is not a `Section-header` or `Title`.  Please remove this.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -614,6 +614,197 @@ def postprocess_element(self, elem: Element) -> Element:
...
+            "Title",
+        ]:
+            # Add header to next element
+            element2["_header"] = element1["_header"]
+            if element1.data["_header"]:
+                if element2.text_representation:
+                    element2.text_representation = element1.data["_header"] + "\n" + element2.text_representation
+                else:
+                    element2.text_representation = element1.data["_header"]
+            return False
```
MarkLindblad, line 706: I think this is right. How do our embeddings work for images? Do we embed the image itself, its summary, and its text_representation?  Do we embed the text_representation separately from the summary? Does embedding the image itself count against the embedding token limit?

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -614,6 +614,197 @@ def postprocess_element(self, elem: Element) -> Element:
...
+            "Section-header",
+            "Title",
+        ]:
+            # Add header to next element
+            element2["_header"] = element1["_header"]
+            if element1.data["_header"]:
+                if element2.text_representation:
+                    element2.text_representation = element1.data["_header"] + "\n" + element2.text_representation
+                else:
+                    element2.text_representation = element1.data["_header"]
```
MarkLindblad, line 705: Do we want to be incrementing the `token_count` of the elements we add the `_header` to?

lib/sycamore/sycamore/transforms/split_elements.py
MarkLindblad: I'd rather have it log a warning. I don't want users to see "Over token limit" in their final product and not know what it is. The output is still usable even if it is missing the `_header`. You can add the `_header` text to the warning message.

lib/sycamore/sycamore/transforms/merge_elements.py
```diff
@@ -614,6 +614,197 @@ def postprocess_element(self, elem: Element) -> Element:
...
+        else:
+            new_elt.binary_representation = elt1.binary_representation + elt2.binary_representation
+
+        # Merge text representations by concatenation with a newline
+        new_elt_text_representation = "\n".join(filter(None, [elt1.text_representation, elt2.text_representation]))
+        new_elt.text_representation = new_elt_text_representation if new_elt_text_representation else None
+        if elt1.text_representation is None or elt2.text_representation is None:
+            new_elt.data["token_count"] = tok1 + tok2
+        else:
+            new_elt.data["token_count"] = tok1 + 1 + tok2
```
MarkLindblad, line 766: The logic for calculating the new `token_count` is wrong for elements the merged in an image that started with a summary. It looks like HeaderAugmenterMerger typically throws away the summary and keeps the text_representation which could have a greater or lesser number of tokens than the summary except when we don't merge an Image and leave it by itself. The splitter takes care of this issue so this can be a "won't fix" if you want.

One fix would be to change `preprocess_element`.

dhruvkaliraman7: Rather than having every merger do this, I think it would be better to make the decision of adding summary to text representation in the summarize and summarize_images transform? Right now, the greedy section merger adds summary only of images to text_representation.

lib/sycamore/sycamore/tests/unit/transforms/test_merge_elements.py
```diff
@@ -386,3 +392,129 @@ def test_merge_empty_text_works(self):
...
+        input_dataset = ray.data.from_items([{"doc": self.doc.serialize()}])
+        execute = mocker.patch.object(node, "execute")
+        execute.return_value = input_dataset
+        tokenizer = HuggingFaceTokenizer("sentence-transformers/all-MiniLM-L6-v2")
+        merger = HeaderAugmenterMerger(tokenizer, 120, merge_across_pages=True)
+        merge = Merge(node, merger)
+        output_dataset = merge.execute()
+        output_dataset.show()
+
+    def test_docset_greedy(self):
```
MarkLindblad, line 508: Should this be called `test_docset_augmented` instead?

lib/sycamore/sycamore/transforms/merge_elements.py
MarkLindblad: Update: Based on your slack conversation with @bsowell, it seems like he does not want mergers to use the summary. In that case, just change `preprocess_element`.

MarkLindblad: Are we okay with the fact we currently throw away summaries when we merge images with other elements?

MarkLindblad: > Rather than having every merger do this, I think it would be better to make the decision of adding summary to text representation in the summarize and summarize_images transform? Right now, the greedy section merger adds summary only of images to text_representation.

Ok.

lib/sycamore/sycamore/transforms/merge_elements.py
dhruvkaliraman7: Decided to do text consolidation at summarize, summarize_images transform level

lib/sycamore/sycamore/transforms/merge_elements.py
MarkLindblad: Ok.

MarkLindblad (approved):
Looks great!

PR #948, opened 2024-10-18
fix empty strings in section headers

HenryL27:
there was a code path that would cause new elements to have no "token_count", which would then crash the program


alexaryn (approved):
LGTM

PR #949, opened 2024-10-18
Update trace UI to reflect cache hits.

mdwelsh:
I will disable caching for now on the live demo instance, but at least this will make it clear why a given node apparently has no traces.


mdwelsh:
Closing in favor of #959

PR #950, opened 2024-10-18
Work around hang observed during vcrpy recording.

alexaryn:
In response to customer issue via email.

### DIFF
--- a/lib/aryn-sdk/aryn_sdk/partition/partition.py
+++ b/lib/aryn-sdk/aryn_sdk/partition/partition.py
@@ -107,7 +107,7 @@ def partition_file(
 
     _logger.debug(f"{options_str}")
 
-    # Work around hanging behavior when vcrpy is recording...
+    # Workaround for vcr.  See https://github.com/aryn-ai/sycamore/issues/958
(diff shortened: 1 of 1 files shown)


HenryL27:
say what now?

HenryL27 (approved):
alrighty then

lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -107,9 +107,19 @@ def partition_file(
 
     _logger.debug(f"{options_str}")
 
+    # Work around hanging behavior when vcrpy is recording...
```
HenryL27, line 110: actually can you add more verbosity to the comment to describe why this is necessary

HenryL27 (changes requested):

lib/aryn-sdk/aryn_sdk/partition/partition.py
HenryL27: maybe stick the details in an issue, link it and close it

HenryL27 (approved):

PR #952, opened 2024-10-18
Use pytest-xdist for unit tests.

mdwelsh:
This PR adds `pytest-xdist` to unit tests which allows them to be run in parallel. Using `pytest -n auto` will run tests in multiple concurrent processes, with the number of processes determined based on hardware parallelism. This speeds up test execution by ~2x on my laptop but maybe more in other settings.


HenryL27 (approved):
looks like it gets us from ~5 mins to ~4 mins in github. Nice! I wonder if there's any weirdnesses from the couple of unit tests that still run ray

PR #953, opened 2024-10-18
Fix source_directory path in conf.py

sravan1946:
The `edit this page` button in docs redirect to wrong path
image


alexaryn (approved):
LGTM.  Thanks!

PR #959, opened 2024-10-21
Allow for returning query traces on cached query executions.

mdwelsh:
This PR replaces #949 with a better implementation that allows the results of cached query executions to return traces. This is done by removing the concept of a separate 'trace_dir' and returning the location of the cache results for each node in the query plan. This way we can have our cache and traces too :-)

The PR also makes a few other small changes:

* Returning a query ID and a result from the query client was becoming unwieldy, so I added a new `SycamoreQueryResult` type which includes the query ID, query plan, query result, and the location of the query traces.
* I replaced a couple of places where we were manually reading in trace files with the use of `.materialize()` so we can use the same logic for reading the traces back everywhere.
* I fixed up the Sycamore query integration tests, relaxing the constraints on the exact answer, and fixing a bug whereby the code generated for `.rerank` was broken. This test now passes!


lib/sycamore/sycamore/query/result.py
```diff
@@ -0,0 +1,45 @@
+import io
+from typing import Any, Dict, Optional
+
+from pydantic import BaseModel
+
+from sycamore.query.logical_plan import LogicalPlan
+from sycamore import DocSet
+
+
+class SycamoreQueryResult(BaseModel):
```
baitsguy, line 10: 🙇

lib/sycamore/sycamore/query/result.py
```diff
@@ -0,0 +1,45 @@
...
+
+    result: Any
+    """The result of the query operation. Depending on the query, this could be a list of documents,
+    a single document, a string, an integer, etc.
+    """
+
+    code: Optional[str] = None
+    """The Python code corresponding to the query plan."""
+
+    trace_dirs: Optional[Dict[int, Optional[str]]] = None
```
baitsguy, line 27: This is essentially a copy of the logical_plan.nodes with runtime information added to it. It's just cache path here but I want to add metrics (at least counts and latency) soon too.

Does it make sense to have a ExecutionResult style object that contains all this? Overall I think we want query level information (basically this object), and then finer-access to individual nodes. 

We can also choose to dump everything in the trace dir (like the [query-plan file](https://github.com/aryn-ai/sycamore/blob/9e2afb8531a31b9f19b4cad762eac3d01fe60ce6/lib/sycamore/sycamore/query/execution/sycamore_executor.py#L272)) but we'd still need an objet that deserializes those files. It would be good for the API response to just give you that stuff

lib/sycamore/sycamore/tests/integration/query/execution/test_sycamore_query.py
```diff
@@ -19,9 +19,9 @@ def test_simple(self, query_integration_test_index: str, codegen_mode: bool):
             schema,
             natural_language_response=True,
         )
-        query_id, result = client.run_plan(plan, codegen_mode=codegen_mode)
-        assert isinstance(result, str)
-        assert len(result) > 0
+        result = client.run_plan(plan, codegen_mode=codegen_mode)
+        assert isinstance(result.result, str)
```
baitsguy, line 23: seems like this is failing

lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -81,32 +82,33 @@ def __init__(
...
-    def process_node(self, logical_node: Node, query_id: str, is_result_node: Optional[bool] = False) -> Any:
+    def process_node(
+        self, logical_node: Node, result: SycamoreQueryResult, is_result_node: Optional[bool] = False
+    ) -> Any:
         """Process the given node. Recursively processes dependencies first."""
 
-        bind_contextvars(logical_node=logical_node)
+        query_id = result.query_id
+
         # This is lifted up here to avoid serialization issues with Ray.
```
baitsguy, line 92: is this comment still relevant?

lib/sycamore/sycamore/query/result.py
baitsguy: If we do that we should probably also rename the variable to execution dir or something

lib/sycamore/sycamore/query/result.py
mdwelsh: I cannot tell if this is someone bowing in thanks for smacking their head on the table

lib/sycamore/sycamore/query/result.py
mdwelsh: Yeah, I thought we might want to just put cache_dir into the LogicalPlan, but them we're mixing logical plan and physical execution stuff. Ideally the plan is "frozen" once generated and we don't manipulate it afterwards, but you're right we need to capture things like metrics and stuff.

I don't have a strong feeling about this. We could make this type an "ExecutionNode" which we could extend with metrics etc, or we could fold it into LogicalPlan. WDYT?

lib/sycamore/sycamore/query/result.py
baitsguy: Bowing in thanks!

lib/sycamore/sycamore/query/result.py
baitsguy: Seems like we need a Query object? Where a GetQuery gives you:

```
class Query:
  LogicalPlan
  configs (cache_dir, codegen, llm)
  execution_info (log pointers, node->cached data, metrics)
```

or something along those lines?

lib/sycamore/sycamore/query/execution/sycamore_executor.py
mdwelsh: nope

lib/sycamore/sycamore/tests/integration/query/execution/test_sycamore_query.py
mdwelsh: It was passing for me but the test is still flaky

lib/sycamore/sycamore/query/result.py
mdwelsh: See what I've done with the NodeExecution type.

apps/query-ui/queryui/util.py
```diff
@@ -218,29 +220,36 @@ def format_value(value):
...
         for filename in os.listdir(directory):
             f = os.path.join(directory, filename)
             if os.path.isfile(f):
+                self.num_files += 1
                 newdoc = self.readfile(f)
                 if newdoc:
                     docs.append(newdoc)
 
         # Only keep parent docs if there are child docs in the list.
         parent_docs = [x for x in docs if x.get("parent_id") is None]
```
baitsguy, line 247: we might want to consider getting rid of this at some point now that we are using reconstructed documents in queries. Will be a way to enforce we're using the same model all through the stack

apps/query-ui/queryui/util.py
```diff
@@ -250,36 +259,30 @@ def readdata(self):
...
-        st.subheader(f"Node {self.node_id}")
+        st.subheader(f"Node {self.node_id} - {node.node_type if node else 'n/a'}")
         st.markdown(f"*Description: {node.description if node else 'n/a'}*")
         if self.df is None or not len(self.df):
-            st.write(":red[0] documents")
-            st.write("No data.")
+            if self.num_files > 0:
+                st.write(":green[Cache hit - node execution skipped!]")
+            else:
+                st.write(f":green[Node type {node.node_type} does not produce document traces]")
```
baitsguy, line 268: nice

lib/sycamore/sycamore/query/execution/sycamore_executor.py
```diff
@@ -81,32 +82,33 @@ def __init__(
...
 
-        if self.trace_dir and not self.dry_run:
-            trace_dir = os.path.join(self.trace_dir, query_id, str(logical_node.node_id))
-            os.makedirs(trace_dir, exist_ok=True)
-        else:
-            trace_dir = None
-
         if self.cache_dir and not self.dry_run:
             cache_dir = os.path.join(self.cache_dir, logical_node.cache_key())
+            if result.execution is None:
```
baitsguy, line 100: nit: can probably simplify this to not have if statements directly, e.g.
result.execution = result.execution or {}
result.execution[logical_node.node_id] = result.execution.get(logical_node.node_id, NodeExecution(...))

stylistic pref I suppose

lib/sycamore/sycamore/query/result.py
```diff
@@ -0,0 +1,54 @@
+import io
+from typing import Any, Dict, Optional
+
+from pydantic import BaseModel
+
+from sycamore.query.logical_plan import LogicalPlan
+from sycamore import DocSet
+
+
+class NodeExecution(BaseModel):
```
baitsguy, line 10: 🙇

baitsguy (approved):

PR #963, opened 2024-10-21
Couple of minor fixes and tweaks to the table merger.

bsowell:
1. I observed the LLM sometimes having issues distinguishing where one table ended and the next started, so I added headers - either ELEMENT 1/ELEMENT 2 or DOCUMENT 1/DOCUMENT 2 in the llm_query transform.

2. Added a default implementation for preprocess_element and postprocess_element in the ElementMerger base class. This just reduces code in some simple sub-classes.

3. Added code to update the table_continuation to be True or False. This is because I ended up getting better results when asking the LLM to explain its reasoning in some cases, but I didn't want all of the output carried along in the element.

4. Minor fix to address an invalid escape character warning in a comment.

### DIFF
--- a/lib/sycamore/sycamore/transforms/llm_query.py
+++ b/lib/sycamore/sycamore/transforms/llm_query.py
@@ -104,8 +104,17 @@ class LLMTextQueryAgent:
                     .render(doc=object)
                 )
             else:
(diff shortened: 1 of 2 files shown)


dhruvkaliraman7 (approved):
lgtm. Although I didn't understand 3. How did adding boolean help ?

PR #965, opened 2024-10-22
Update Makefile.docker-base

AranavMahalpure:
The changes include switching to a slimmer base image (python:3.11-slim), combining RUN commands, and adding cleanup steps to reduce the final image size. Additionally, a multi-stage build was implemented to ensure only necessary runtime files are included, enhancing efficiency and maintainability.

### DIFF
--- a/apps/docker-base/Makefile.docker-base
+++ b/apps/docker-base/Makefile.docker-base
@@ -1,35 +1,40 @@
-# -*- makefile -*-
+# Makefile for Docker image setup
+
 help:
 	@echo "This should be run as part of the Dockerfile"
 	false
 
 user-setup:
+	# Create app group and user
 	groupadd --gid 1000 app
 	useradd -d /app --uid 1000 --gid app app
 	chown -R app:app /app
 
 apt-setup:
+	# Configure APT to keep downloaded packages
 	rm -f /etc/apt/apt.conf.d/docker-clean
 	echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
 
 apt-install:
-	DEBIAN_FRONTEND=noninteractive apt update
-	# gcc and python3-dev needed on arm for guidance
-	DEBIAN_FRONTEND=noninteractive apt -y install --no-install-recommends python3-poetry gcc python3-dev
+	# Install necessary packages without recommended packages
+	DEBIAN_FRONTEND=noninteractive apt update && \
+	DEBIAN_FRONTEND=noninteractive apt -y install --no-install-recommends python3-poetry gcc python3-dev && \
+	# Clean up to reduce image size
+	apt clean && rm -rf /var/lib/apt/lists/*
 
 non-root-files-check:
+	# Check for any files owned by root
 	find . -uid 0 -ls
 	test $$(find . -uid 0 -print | wc -w) = 0
 
 record-version:
+	# Ensure GIT_COMMIT is set and record the version
 	test "$(GIT_COMMIT)" != ""
 	test "$(GIT_COMMIT)" != "unknown"
 	touch .git.commit.$(GIT_COMMIT)
 
-# Allow images that depend on the docker base image to verify that the version for their
-# source code is consistent with the version in the base image. If the code is inconsistent,
-# the resulting image could behave unexpectedly.
 check-version-compatibility:
+	# Verify the version consistency
 	test "$(GIT_COMMIT)" != ""
 	test "$(GIT_COMMIT)" != "unknown"
 	ls .git.commit.*


apps/docker-base/Makefile.docker-base
```diff
@@ -1,35 +1,40 @@
...
 	false
 
 user-setup:
+	# Create app group and user
 	groupadd --gid 1000 app
 	useradd -d /app --uid 1000 --gid app app
 	chown -R app:app /app
 
 apt-setup:
+	# Configure APT to keep downloaded packages
```
eric-anderson, line 14: because we cache them as part of the build so want them to be retained.

apps/docker-base/Makefile.docker-base
```diff
@@ -1,35 +1,40 @@
...
 	rm -f /etc/apt/apt.conf.d/docker-clean
 	echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
 
 apt-install:
-	DEBIAN_FRONTEND=noninteractive apt update
-	# gcc and python3-dev needed on arm for guidance
-	DEBIAN_FRONTEND=noninteractive apt -y install --no-install-recommends python3-poetry gcc python3-dev
+	# Install necessary packages without recommended packages
+	DEBIAN_FRONTEND=noninteractive apt update && \
```
eric-anderson, line 20: You don't need to do the && thing here.
make will run both of the steps and will abort if the previous one fails.

apps/docker-base/Makefile.docker-base
```diff
@@ -1,35 +1,40 @@
...
 	chown -R app:app /app
 
 apt-setup:
+	# Configure APT to keep downloaded packages
 	rm -f /etc/apt/apt.conf.d/docker-clean
 	echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
 
 apt-install:
-	DEBIAN_FRONTEND=noninteractive apt update
-	# gcc and python3-dev needed on arm for guidance
```
eric-anderson, line 17: Please don't lose this comment. It's very important to understand why we're installing gcc & python3-dev (which otherwise we wouldn't expect to need).

apps/docker-base/Makefile.docker-base
```diff
@@ -1,35 +1,40 @@
...
-	DEBIAN_FRONTEND=noninteractive apt -y install --no-install-recommends python3-poetry gcc python3-dev
+	# Install necessary packages without recommended packages
+	DEBIAN_FRONTEND=noninteractive apt update && \
+	DEBIAN_FRONTEND=noninteractive apt -y install --no-install-recommends python3-poetry gcc python3-dev && \
+	# Clean up to reduce image size
```
eric-anderson, line 22: We don't want this because of the caching.   From docker-base/Dockerfile.buildx
```
RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
    --mount=type=cache,target=/var/lib/apt,sharing=locked \
    make -f Makefile.docker-base apt-install
```
The first two lines are caching the directories, so they don't show up in the image.

You can verify that on the existing containers:
```
% docker run -it -u root arynai/sycamore-base bash
root@9983ae28df15:/app# find /var/cache/apt
/var/cache/apt
/var/cache/apt/archives
/var/cache/apt/archives/lock
/var/cache/apt/archives/partial
root@9983ae28df15:/app# find /var/lib/apt
/var/lib/apt
/var/lib/apt/mirrors
/var/lib/apt/mirrors/partial
/var/lib/apt/periodic
/var/lib/apt/extended_states
/var/lib/apt/lists
```

apps/docker-base/Makefile.docker-base
```diff
@@ -1,35 +1,40 @@
...
 	find . -uid 0 -ls
 	test $$(find . -uid 0 -print | wc -w) = 0
 
 record-version:
+	# Ensure GIT_COMMIT is set and record the version
 	test "$(GIT_COMMIT)" != ""
 	test "$(GIT_COMMIT)" != "unknown"
 	touch .git.commit.$(GIT_COMMIT)
 
-# Allow images that depend on the docker base image to verify that the version for their
```
eric-anderson, line 29: Please don't get rid of the longer comment that explains why we are doing this.
Feel free to move it under the target for consistency, but readers of the code need to understand what version consistency is being checked and why.

apps/docker-base/Makefile.docker-base
```diff
@@ -1,35 +1,40 @@
-# -*- makefile -*-
```
eric-anderson, line 1: Please don't get rid of this line. It's magic that helps some editors.

eric-anderson (changes requested):
Thanks for the PR, adding the comments to each of the pieces should help people reading the file. 

Your PR description mentions changing the base image and a multi-stage build. I don't see any of those steps in here (I expected to see changes to docker-base/Dockerfile.buildx.  Did you intend to have a different PR description?

apps/docker-base/Makefile.docker-base
AranavMahalpure: Thanks for your reply.
  I will check all  this

eric-anderson:

PR #971, opened 2024-10-22
Add score to each document in OpenSearch query result.

bsowell:
This makes it possible to sort the results by score and do a limit.


baitsguy (approved):
Can you update the unit test to test this before pushing

PR #972, opened 2024-10-23
Refactor `mark_bbox_preset` to expose function outside `DocSet`

MarkLindblad:


lib/sycamore/sycamore/docset.py
```diff
@@ -723,30 +723,58 @@ def summarize(self, summarizer: Summarizer, **kwargs) -> "DocSet":
...
-            SortByPageBbox,
-            MarkDropTiny,
-            MarkDropHeaderFooter,
-            MarkBreakPage,
-            MarkBreakByColumn,
-            MarkBreakByTokens,
-        )
+        from sycamore.transforms.mark_misc import MarkBBoxPreset
+
+        preset = MarkBBoxPreset(self.plan, tokenizer, token_limit, **kwargs)
```
alexaryn, line 775: For clearer camel-case tokenization, can we call it `MarkBboxPreset`?

lib/sycamore/sycamore/transforms/mark_misc.py
```diff
@@ -116,3 +116,30 @@ def mark_break_by_tokens(parent: Document, tokenizer: Tokenizer, limit: int) ->
...
+
+class MarkBBoxPreset(SingleThreadUser, NonGPUUser, Map):
+    """
+    See DocSet.mark_bbox_preset for details.
+    """
+    def __init__(self, child: Node, tokenizer: Tokenizer, token_limit: int = 512, **resource_args):
+        super().__init__(child, f=MarkBBoxPreset.mark_bbox_preset, args=[tokenizer, token_limit], **resource_args)
+
+    @staticmethod
+    @timetrace("markBBoxPreset")
```
alexaryn, line 132: We don't need a timetrace here because all the functions called below are wrapped in more specific timetraces.

lib/sycamore/sycamore/transforms/mark_misc.py
```diff
@@ -116,3 +116,30 @@ def mark_break_by_tokens(parent: Document, tokenizer: Tokenizer, limit: int) ->
...
+    def mark_bbox_preset(parent: Document, tokenizer: Tokenizer, token_limit: int) -> Document:
```
alexaryn, line 133: The tokenizer could be large.  In the past we had problems if we tried to pass such things into an Actor in APS.  The solution in that case was to send None and default it on the callee-side.

lib/sycamore/sycamore/transforms/mark_misc.py
```diff
@@ -116,3 +116,30 @@ def mark_break_by_tokens(parent: Document, tokenizer: Tokenizer, limit: int) ->
...
+        from sycamore.transforms.bbox_merge import MarkDropHeaderFooter, SortByPageBbox
+        from sycamore.transforms.bbox_merge import MarkBreakByColumn
```
alexaryn, line 135: Should these be combined?

lib/sycamore/sycamore/docset.py
```diff
@@ -723,30 +723,58 @@ def summarize(self, summarizer: Summarizer, **kwargs) -> "DocSet":
...
+                    MarkBreakByTokens,
+                )
+
+                plan0 = SortByPageBbox(self.plan, **kwargs)
+                plan1 = MarkDropTiny(plan0, 2, **kwargs)
+                plan2 = MarkDropHeaderFooter(plan1, 0.05, 0.05, **kwargs)
+                plan3 = MarkBreakPage(plan2, **kwargs)
+                plan4 = MarkBreakByColumn(plan3, **kwargs)
+                plan5 = MarkBreakByTokens(plan4, tokenizer, token_limit, **kwargs)
+                return DocSet(self.context, plan5)
```
alexaryn, line 771: This might be too much and too nerdy.  If we want the old code, we can check source control.  If a user wanted to do a la carte marking, I'd expect them to do something more like:
```
.transform(SortByPageBbox)
.transform(MarkDropTiny, minimum=2)
...
```

alexaryn (commented):
I wonder if the word "preset" is appropriate anymore.  I guess we're stuck for backward-compatibility, but it seems like it's now more "default" or "standard".

lib/sycamore/sycamore/transforms/mark_misc.py
MarkLindblad: Done.

lib/sycamore/sycamore/transforms/mark_misc.py
MarkLindblad: Removed.

lib/sycamore/sycamore/docset.py
MarkLindblad: Done.

alexaryn (approved):

lib/sycamore/sycamore/transforms/mark_misc.py
MarkLindblad: Done.

PR #974, opened 2024-10-23
OpenSearch improvements

baitsguy:
1. Reduce batch size when loading elements to reconstruct docs, by default OpenSearch supports a max of 10k results so this keeps us within that limit without needing to play with the index and cluster
2. Allow get_knn_query to work with min_score knn queries

### DIFF
--- a/lib/sycamore/sycamore/connectors/opensearch/utils.py
+++ b/lib/sycamore/sycamore/connectors/opensearch/utils.py
@@ -6,20 +6,29 @@ from sycamore.transforms import Embedder
 
 
 @context_params("opensearch")
-def get_knn_query(text_embedder: Embedder, query_phrase: str, k: int = 500, context: Optional[Context] = None):
+def get_knn_query(
+    text_embedder: Embedder,
+    query_phrase: str,
+    k: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/connectors/opensearch/utils.py
```diff
@@ -6,20 +6,29 @@
 
 
 @context_params("opensearch")
-def get_knn_query(text_embedder: Embedder, query_phrase: str, k: int = 500, context: Optional[Context] = None):
+def get_knn_query(
```
mdwelsh, line 9: Nit: Why not throw in `max_distance` while you're at it so all three radial search options are available?

mdwelsh (approved):

lib/sycamore/sycamore/connectors/opensearch/utils.py
baitsguy: Because it only supports a min_score lol. There is a max_distance param that expects an actual physical distance value, skipping that because it's just going to cause confusion until we have a use case

https://opensearch.org/docs/latest/search-plugins/knn/radial-search-knn/#supported-cases

PR #975, opened 2024-10-23
Fix `mark_bbox_preset`'s `MarkDropHeaderFooter` parameter

MarkLindblad:
Fixes bug that where `MarkDropHeaderFooter` always adds `_drop` to all elements, causing all elements to be dropped. This is because the parameters for `MarkDropHeaderFooter.__init__` are different from `MarkDropHeaderFootermark_drop_header_footer`:

```python
  def __init__(self, child: Node, top: float = 0.05, bottom: Optional[float] = None, **resource_args):
      if bottom is None:
          bottom = top
      lo = top
      hi = 1.0 - bottom
      super().__init__(child, f=MarkDropHeaderFooter.mark_drop_header_and_footer, args=[lo, hi], **resource_args)
  
  @staticmethod
  @timetrace("markHeadFoot")
  def mark_drop_header_and_footer(parent: Document, lo: float, hi: float) -> Document:
      for elem in parent.elements:
          bbox = elem.data.get("bbox")
          if (bbox is not None) and ((bbox[1] > hi) or (bbox[3] < lo)):
              elem.data["_drop"] = True  # mark for removal
      return parent
```

### DIFF
--- a/lib/sycamore/sycamore/transforms/bbox_merge.py
+++ b/lib/sycamore/sycamore/transforms/bbox_merge.py
@@ -152,7 +152,7 @@ class MarkDropHeaderFooter(SingleThreadUser, NonGPUUser, Map):
     """
 
     def __init__(self, child: Node, top: float = 0.05, bottom: Optional[float] = None, **resource_args):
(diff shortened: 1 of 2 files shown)


alexaryn (approved):
Nice.

PR #982, opened 2024-10-25
Add query planning strategy objects

baitsguy:
Allows the planner to work with a custom set of operators and rewrite rules. Added a default strategy (current behavior) and a vector search only strategy.

We might consider refactoring the plan processors eventually but this is functional for the current use cases.

You can enable this in evaluation by modifying https://github.com/aryn-ai/sycamore/blob/c18c2eb2daec4f9c30261e42348b1f065c0e0420/apps/query-eval/queryeval/driver.py#L116 as

```python
self.client = SycamoreQueryClient(
            llm_cache_dir=self.config.config.llm_cache_path,
            cache_dir=self.config.config.query_cache_path,
            llm=self.config.config.llm,
            query_plan_strategy=VectorSearchOnlyStrategy()
        )
```


lib/sycamore/sycamore/query/strategy.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class RemoveVectorSearchForAnalytics(LogicalPlanProcessor):
+
+    def __init__(self, llm: LLM) -> None:
+        super().__init__()
+        self.llm = llm
+
+    def __call__(self, plan: LogicalPlan) -> LogicalPlan:
+        logging.info("Executing DefaultLogicalPlanProcessor")
```
mdwelsh, line 47: Logging statement not quite right

mdwelsh (approved):
YOU ARE A MACHINE

PR #983, opened 2024-10-26
Move tokenized data to device

baitsguy:
The tokenized data was living in the CPU (since tokenizer is only on the cpu), needed to move it to the gpu for scoring


HenryL27 (approved):
it doesn't need to be moved back at the end does it?

baitsguy:
> it doesn't need to be moved back at the end does it?

Nope our cpu code should be able to access the tensors. Only if you're trying to do an op that's combing 2 data sources in difference devices is when issues typically arise

PR #986, opened 2024-10-26
Updated chunker.py

Ranjana761:
## Summary 

The improved implementation addresses several key issues and adds important features:

- Generic Type Support: Added proper generic type support using TypeVar, making the chunker work with any sequence type, not just text.
- Better Error Handling:
- Empty Input Handling: Properly handles empty input sequences by returning an empty list
- Improved Documentation:

### DIFF
--- a/lib/sycamore/sycamore/functions/chunker.py
+++ b/lib/sycamore/sycamore/functions/chunker.py
@@ -1,41 +1,86 @@
 from abc import abstractmethod
-from typing import Any
+from typing import Any, List, TypeVar, Generic
 
+T = TypeVar('T')
 
-class Chunker:
+class Chunker(Generic[T]):
     @abstractmethod
-    def chunk(self, tokens: list[Any]) -> list[Any]:
+    def chunk(self, tokens: List[T]) -> List[List[T]]:
         pass
 
-
-class TextOverlapChunker(Chunker):
+class TextOverlapChunker(Chunker[T]):
     """
-    TextOverlapChunker is a class for chunking text into smaller segments while allowing for token overlap.
-
-    This class inherits from the Chunker class and is designed to divide long text tokens into chunks, each containing
-    a specified number of tokens. It allows for a controlled overlap of tokens between adjacent chunks.
-
+    TextOverlapChunker is a class for chunking sequences into smaller segments with controlled overlap.
+    
+    This class inherits from the Chunker class and divides sequences into chunks of a specified size,
+    with configurable overlap between adjacent chunks. The implementation ensures that:
+    1. All tokens are included in at least one chunk
+    2. No chunk exceeds the specified maximum size
+    3. Overlap is consistently maintained between chunks
+    4. The last chunk is handled correctly even if smaller than chunk_token_count
+    
     Args:
         chunk_token_count: The maximum number of tokens to include in each chunk.
-        chunk_overlap_token_count: The number of tokens that can overlap between adjacent chunks.
-            This value must be less than the `chunk_token_count` to ensure meaningful chunking.
-
+        chunk_overlap_token_count: The number of tokens that should overlap between adjacent chunks.
+            Must be less than chunk_token_count.
+            
+    Raises:
+        ValueError: If chunk_overlap_token_count >= chunk_token_count or if either parameter is negative.
+        
     Example:
-         .. code-block:: python
-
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/functions/chunker.py
```diff
@@ -1,41 +1,86 @@
...
+        
+        for start in range(0, len(tokens), stride):
+            # Calculate end index for current chunk
+            end = min(start + self._chunk_token_count, len(tokens))
+            chunk = tokens[start:end]
+            
+            # Add chunk if it's the first chunk, maintains minimum size, or is the last piece
+            if (start == 0 or 
+                len(chunk) >= self._chunk_overlap_token_count or 
+                end == len(tokens)):
```
HenryL27, line 79: I'm struggling to understand why we need all these conditions. Is there ever a case, given the validations added in `__init__` where this can evaluate to False?

if this is not the last chunk, then
len(chunk) = _chunk_token_count >= _chunk_overlap_token_count

if this is the last chunk then we add it anyway.

I think you only need the if start == 0 case for when there is only one rather small chunk, in which case it's also the last chunk.

lib/sycamore/sycamore/functions/chunker.py
```diff
@@ -1,41 +1,86 @@
...
+            return []
+            
+        chunks = []
+        stride = self._chunk_token_count - self._chunk_overlap_token_count
+        
+        for start in range(0, len(tokens), stride):
+            # Calculate end index for current chunk
+            end = min(start + self._chunk_token_count, len(tokens))
+            chunk = tokens[start:end]
+
```
HenryL27, line 75: python can understand out-of-bounds slices and truncates appropriately. e.g.
```
>>> "01234"[2:10]
"234"
```

lib/sycamore/sycamore/functions/chunker.py
```diff
@@ -1,41 +1,86 @@
...
+            
+            # Add chunk if it's the first chunk, maintains minimum size, or is the last piece
+            if (start == 0 or 
+                len(chunk) >= self._chunk_overlap_token_count or 
+                end == len(tokens)):
+                chunks.append(chunk)
+                
+            # If we've processed all tokens, break
+            if end == len(tokens):
+                break
```
HenryL27, line 84: we can just exit the for loop because we're at the end of the for loop in this case, right?

HenryL27, line 84: putting all these together, you can rewrite the loop more concisely as
```python
for start in range(0, len(tokens), stride):
    chunks.append(tokens[start: start + self._chunk_token_count])
```
which is equivalent to the original list comprehension

HenryL27 (commented):
Docstrings and validations are appreciated, thank you.

It appears that some unit tests are failing. Can you make sure your implementation passes the tests? If the current implementation fails in some edge case can you add that to the tests? Otherwise I don't understand why it needs to be changed.
(`poetry run pytest lib/sycamore/sycamore/tests/unit/functions/test_text_chunker.py`)

Also please add a few test cases for the negative parameters

lib/sycamore/sycamore/functions/chunker.py
HenryL27: edit: I think this is the cause of the test failures - if the natural chunk boundary lines up exactly with the end of the sequence then you'll break prematurely, right?
overlap = 2, len = 5
"ABCDEFGH" -> "ABCDE", "DEFGH" instead of "ABCDE", "DEFGH", "GH"

lib/sycamore/sycamore/functions/chunker.py
```diff
@@ -1,41 +1,86 @@
...
+from typing import Any, List, TypeVar, Generic
 
+T = TypeVar('T')
 
-class Chunker:
+class Chunker(Generic[T]):
     @abstractmethod
-    def chunk(self, tokens: list[Any]) -> list[Any]:
+    def chunk(self, tokens: List[T]) -> List[List[T]]:
         pass
```
MarkLindblad, line 9: Generics were added in python 3.12. We want to be able to support >=3.9,<3.13
https://github.com/aryn-ai/sycamore/blob/96b61e0741b2f763e9341f64d06b45cbdda96928/lib/sycamore/pyproject.toml#L16

eric-anderson:

PR #988, opened 2024-10-28
Sync Locks across all Directories

karanataryn:
The product of changes after running `lib/poetry-lock/poetry-lock-all.sh` across all directories and fixing arising issues.


lib/poetry-lock/poetry-lock-all.sh
```diff
@@ -35,6 +35,9 @@ for i in ${tomls}; do
         cd $(dirname "$i")
         poetry lock --no-update || exit 1
         poetry install 2>&1 | tee /tmp/poetry-install.out || exit 1
-        perl -ne 'print qq{$1 = "$2"\n} if /Downgrading (\S+) \((\S+) ->/o;' </tmp/poetry-install.out
+        perl -ne 'print qq{$1 = "$2"\n} if /Downgrading (\S+) \((\S+) ->/o;' </tmp/poetry-install.out >/tmp/downgraded
+        cat /tmp/downgraded
+        [[ $(wc -l </tmp/downgraded) -eq 0 ]] || exit 1
     ) || exit 1
 done
+echo "SUCCESS"
```
eric-anderson, line 43: nit: missing newline.

lib/poetry-lock/pyproject.toml
```diff
@@ -28,15 +28,15 @@ networkx = "3.2.1"
 ipython = "8.18.1"
 
 # Package versions from being yanked
-zope-interface = ">=6.4.post0"
+zope-interface = "7.1.0"
```
eric-anderson, line 31: I'd do >=7.1.0 here so we don't force people to have zope-interface 7.1.0
Also this should be in a different section now, I'm assuming just below 7.1.0 wasn't yanked.

It's possible you can just remove it.  If we need 7.1.0 for recency, the dependency should be in some other repo.

lib/poetry-lock/pyproject.toml
```diff
@@ -28,15 +28,15 @@ networkx = "3.2.1"
...
 matplotlib = "3.9.0"
 
 # Package versions specified because of github vulnerability reports
 # aiohttp = ">=3.9.5"
 #requests = ">=2.32.0"
 scrapy = ">=2.11.2"
 jinja2 = ">=3.1.4"
-idna = ">=3.7"
+idna = "3.10"
```
eric-anderson, line 39: Same comment on >= here.
Was this updated here because of a vulnerability issue?  If we need 3.10 for code reason it should be in that repo and removed here.

karanataryn: Removed.

lib/poetry-lock/pyproject.toml
karanataryn: Removed.

lib/poetry-lock/poetry-lock-all.sh
karanataryn: Done.

eric-anderson (approved):

PR #989, opened 2024-10-29
Add Closure of Client Connections for Connectors

karanataryn:
Client connections often do not close properly when implicitly handled, here we add the ability to explicitly close them and add it for DuckDB, where this problem was first discovered.

### DIFF
--- a/lib/sycamore/sycamore/connectors/duckdb/duckdb_writer.py
+++ b/lib/sycamore/sycamore/connectors/duckdb/duckdb_writer.py
@@ -14,13 +14,12 @@ from sycamore.connectors.common import convert_to_str_dict, _get_pyarrow_type
 
 @dataclass
 class DuckDBWriterClientParams(BaseDBWriter.ClientParams):
-    pass
+    db_url: str
 
 
 @dataclass
 class DuckDBWriterTargetParams(BaseDBWriter.TargetParams):
     dimensions: int
-    db_url: Optional[str] = "tmp.db"
     table_name: Optional[str] = "default_table"
     batch_size: int = 1000
     schema: dict[str, str] = field(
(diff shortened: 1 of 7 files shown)


lib/sycamore/sycamore/connectors/duckdb/duckdb_writer.py
```diff
@@ -59,7 +56,9 @@ def compatible_with(self, other: BaseDBWriter.TargetParams) -> bool:
 class DuckDBClient(BaseDBWriter.Client):
     @requires_modules("duckdb", extra="duckdb")
     def __init__(self, client_params: DuckDBWriterClientParams):
-        pass
+        if not client_params.db_url:
+            raise ValueError(f"Must provide valid disk location. Location Specified: {client_params.db_url}")
```
HenryL27, line 60: "Must provide valid disk location" feels like it could be misleading since this errors if it's "" or None or something

lib/sycamore/sycamore/connectors/duckdb/duckdb_writer.py
```diff
@@ -125,23 +120,22 @@ def create_target_idempotent(self, target_params: BaseDBWriter.TargetParams):
...
-                client.sql(f"CREATE TABLE {dict_params.get('table_name')} ({columns_str})")
+                self._client.sql(f"CREATE TABLE {dict_params.get('table_name')} ({columns_str})")
             else:
                 logging.warning(
                     f"""Error creating table {dict_params.get('table_name')}
                     in database {dict_params.get('db_url')}: no schema provided"""
                 )
         except Exception:
-            return
+            logging.info(f"Table {dict_params.get('table_name')} already exists")
```
HenryL27, line 130: are there any other errors that could come up here? Maybe change the log to something like "Table <table> could not be created. most likely it already exists". Might also make it a debug level or something since you'll get hundreds of these on big jobs

HenryL27 (approved):

lib/sycamore/sycamore/connectors/duckdb/duckdb_writer.py
karanataryn: Changed.

PR #992, opened 2024-10-30
Merge in Fixes from Luna Demo Deployment

karanataryn:
Merges in fixes from Luna Demo deployment branch.


apps/query-ui/queryui/loaddata.py
```diff
@@ -148,6 +148,8 @@ def main():
 
     partitioned_docset = (
         docset.partition(partitioner=ArynPartitioner(extract_table_structure=True, use_ocr=True, extract_images=True))
+# XXX MDW 3 Oct 2024 - Disable this for now as it seems to be failing on the demo instance.
```
mdwelsh, line 151: Just remove these lines.

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -143,9 +143,11 @@ def execute(self) -> Any:
         os_query = get_knn_query(query_phrase=self.logical_node.query_phrase, context=self.context)
         if self.logical_node.opensearch_filter:
             os_query["query"]["knn"]["embedding"]["filter"] = self.logical_node.opensearch_filter
-        result = self.context.read.opensearch(
-            index_name=self.logical_node.index, query=os_query, reconstruct_document=True
-        ).rerank(query=self.logical_node.query_phrase)
+# XXX XXX XXX MDW 21 Oct 2024 - Disable .rerank() for the Luna demo since it seems to hurt performance a lot.
```
mdwelsh: This is one we can't just upstream in this case. I don't know quite what to do here. It probably needs to be an option on the QueryVectorDatabase operator. For now let's just leave this hack on the branch.

lib/sycamore/sycamore/tests/integration/query/execution/test_sycamore_query.py
```diff
@@ -80,7 +80,7 @@ def test_vector_search(self, query_integration_test_index: str, codegen_mode: bo
             "give me some wind related incidents",
             query_integration_test_index,
             schema,
-            natural_language_response=False,
+            natural_language_response=True,
```
mdwelsh, line 83: Not sure about this one. I think this is a bug as I think it should actually be `False` for this test.

mdwelsh (approved):

PR #993, opened 2024-10-31
Add table-chunker

dhruvkaliraman7:
1. Add a "column_headers" field to the table element, to store the column headers.
2. While Splitting elements for tables, split at a new line character, and attach column headers and other metadata to table chunk.

### DIFF
--- a/lib/sycamore/sycamore/data/table.py
+++ b/lib/sycamore/sycamore/data/table.py
@@ -99,7 +99,9 @@ class Table:
     are lossy, since, for instance, CSV does not natively support spanning cells.
     """
 
-    def __init__(self, cells: list[TableCell], caption: Optional[str] = None):
+    def __init__(
+        self, cells: list[TableCell], caption: Optional[str] = None, column_headers: Optional[list[str]] = None
+    ):
         """Creates a new Table.
 
         Args:
@@ -111,6 +113,7 @@ class Table:
         self.caption = caption
         self.num_rows = max(max(c.rows) for c in self.cells) + 1
         self.num_cols = max(max(c.cols) for c in self.cells) + 1
+        self.column_headers = column_headers
 
     def __eq__(self, other):
         if type(other) is not type(self):
@@ -305,6 +308,7 @@ class Table:
             index=None,
             columns=flattened_header if max_header_prefix_row >= 0 else None,
         )
+        self.column_headers = flattened_header if max_header_prefix_row >= 0 else []
         return df
 
     def to_csv(self, **kwargs) -> str:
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/data/table.py
```diff
@@ -305,6 +308,7 @@ def to_pandas(self) -> DataFrame:
             index=None,
             columns=flattened_header if max_header_prefix_row >= 0 else None,
         )
+        self.column_headers = flattened_header if max_header_prefix_row >= 0 else []
```
HenryL27, line 311: I really don't like mutating the table object in this method. Can we avoid this or do it in like init or something?

lib/sycamore/sycamore/data/table.py
```diff
@@ -112,6 +114,31 @@ def __init__(self, cells: list[TableCell], caption: Optional[str] = None):
...
+                    if cell.is_header and cell.rows[0] <= max_header_prefix_row:
+                        for col in cell.cols:
+                            header_array[cell.rows[0], col] = cell.content
+                        for row in cell.rows[1:]:
+                            for col in cell.cols:
+                                header_array[row, col] = ""
+
+            self.column_headers = []
+            for npcol in header_array.transpose():
+                self.column_headers.append(" | ".join(OrderedDict.fromkeys((c for c in npcol if c not in [None, ""]))))
```
HenryL27, line 140: any chance we can factor this as a function in common with the pandafication function too?

HenryL27 (commented):
now we've ended up with a bunch of duplicated code...

lib/sycamore/sycamore/data/table.py
dhruvkaliraman7: done

lib/sycamore/sycamore/data/table.py
```diff
@@ -275,7 +255,7 @@ def from_html(cls, html_str: Optional[str] = None, html_tag: Optional[Tag] = Non
     # we speculate that duplication may create confusion, so we default to only displaying a cells
     # content for the first row/column for which it is applicable. The exception is for header rows,
     # where we duplicate values to each columnn to ensure that every column has a fully qualified header.
-    def to_pandas(self) -> DataFrame:
+    def to_pandas(self, column_header: Optional[bool] = False) -> Union[DataFrame, List[str]]:
```
HenryL27, line 258: this type can just be bool with a default. I don't think you need 3 states
also (obnoxious nit) might call the param `column_header_only`

HenryL27 (approved):
lgtm, one nit

PR #995, opened 2024-11-01
Return the list of documents referenced in a Luna query.

mdwelsh:
This PR adds logic to `SycamoreQueryResult` to return the list of Document paths referenced by the query.

The idea here is to provide a "reasonably selective" set of documents that a query is using in generating its final response. To this end we walk the query tree from the result node up and return the set of docs from the first node (or set of nodes) that return docs, rather than those that return non-doc values (like text chunks or singleton values).

This PR also adds the resulting set to the response returned in `query-server`.

I ended up changing how this worked a bit from @baitsguy's previous implementation (in a private repo).


dtecuci:
Looks good to me

dtecuci (approved):

PR #996, opened 2024-11-02
Docprep notebook testing

sohamkasar19:


.github/workflows/testing.yml
```diff
@@ -256,3 +258,42 @@ jobs:
...
+    services:
+      opensearch:
+        image: opensearchproject/opensearch:2.10.0
+        env:
+          discovery.type: "single-node"
+        ports:
+          - 9200:9200
+
+    steps:
+      - name: Checkout
```
mdwelsh, line 276: There are a lot of potentially slow steps here that I wonder if a little refactoring of the workflow would not allow them to be done once, rather than separately for each batch of tests. Setting up poetry, python, doing an apt-get update, etc. are all expensive operations. Is it possible for the workflow to do this once?

sohamkasar19: Setting up Python, Poetry, and running apt-get update does take time. Since we use different Python versions for various jobs, I assume the setup is done within each job for compatibility. But thanks for the idea, I'll take a look at how I can optimize the workflows.

sohamkasar19: I looked into this. Apart from the actual test, Sycamore installation takes up most of the time required to complete the workflow. GitHub runs each job independently. So we can't have the setup run once.

HenryL27 (approved):
lgtm. Let's be vigilant about keeping this up to date

PR #997, opened 2024-11-04
Add expected documents in query-eval tool

baitsguy:
Tested with test

This metric only considers recall and doesn't have a capped N

### DIFF
--- a/apps/query-eval/queryeval/driver.py
+++ b/apps/query-eval/queryeval/driver.py
@@ -342,6 +342,9 @@ class QueryEvalDriver:
                 console.print("[red]:x: Document retrieval mismatch")
                 console.print(f"Missing docs: {query.expected_docs - result.retrieved_docs})")
             metrics.doc_retrieval_recall = len(result.retrieved_docs & query.expected_docs) / len(query.expected_docs)
+            metrics.doc_retrieval_precision = len(result.retrieved_docs & query.expected_docs) / len(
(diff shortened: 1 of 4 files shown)


apps/query-eval/queryeval/driver.py
```diff
@@ -330,6 +330,19 @@ def do_eval(self, query: QueryEvalQuery, result: QueryEvalResult) -> QueryEvalRe
...
+            console.print("[yellow]:construction: No expected document list found, skipping.. ")
+        elif not result.retrieved_docs:
+            console.print("[yellow]:construction: No computed document list found, skipping.. ")
+        else:
+            if query.expected_docs.issubset(result.retrieved_docs):
+                console.print("[green]✔ Documents retrieved match")
+            else:
+                console.print("[red]:x: Document retrieval mismatch")
+                console.print(f"Missing docs: {query.expected_docs - result.retrieved_docs})")
+            metrics.doc_retrieval_recall = len(result.retrieved_docs & query.expected_docs) / len(query.expected_docs)
```
mdwelsh, line 344: Mind adding precision while at it?

mdwelsh (approved):
YOU DA MAN

apps/query-eval/queryeval/driver.py
baitsguy: Done

PR #998, opened 2024-11-04
Don't try to put footers in columns

HenryL27:
Don't put page-footers in columns regardless of whether theyre in a column


alexaryn (approved):
No worries.

PR #1007, opened 2024-11-07
Improve suggested install command

HenryL27:
people see, copy, and paste `pip install sycamore-ai[extra]` in the requires_module error message and then get zsh errors bc they need to quote it.


jonfritz (approved):
LGTM!

PR #1009, opened 2024-11-07
Update sycamore to 0.1.26

HenryL27:


lib/sycamore/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "sycamore-ai"
-version = "0.1.25"
```
eric-anderson, line 3: Can you run the poetry lock at least in the root and lib/sycamore directories (up to you if you want to do a lock-all)
Every time we do this I get warnings about inconsistent sycamore versions from poetry.

eric-anderson (approved):

PR #1010, opened 2024-11-07
Add support for using Aryn DocParse chunking from `aryn-sdk`

MarkLindblad:


lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -101,6 +118,8 @@ def partition_file(
     if aryn_config is None:
         aryn_config = ArynConfig()
 
+    docparse_url = aps_url or docparse_url  # For backwards compatibility
```
HenryL27, line 121: would prob change docparse_url to `Optional[str]=None` and write this as 
```python
if aps_url is not None:
    if docparse_url is not None:
        logging.warning("\"aps_url\" and \"docparse_url\" parameters were both set. \"aps_url\" is deprecated. Using \"docparse_url\".")
    else:
        logging.warning("\"aps_url\" parameter is deprecated. Use \"docparse_url\" instead")
        docparse_url = aps_url
if docparse_url is None:
    docparse_url = ARYN_DOCPARSE_URL
assert docparse_url is not None, "wtf"
```

lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -183,6 +203,7 @@ def _json_options(
     extract_images: bool = False,
     selected_pages: Optional[list[Union[list[int], int]]] = None,
     output_format: Optional[str] = None,
+    chunking_options: Optional[dict[str, Any]] = None,
```
HenryL27, line 206: we handle malformed chunking options well in the service, yes?

HenryL27 (approved):
k. Not sure about all the bwc stuff for endpoint logic - we're still in v0.x so I'd prefer to just break it unless we have something critical that depends on the param existing.
Along that vein a middle path is to "deprecate" it - I wrote some code to make that more obvious (although I'll bet there's a more idiomatic way)

lib/aryn-sdk/aryn_sdk/partition/partition.py
MarkLindblad: Yes.

lib/aryn-sdk/aryn_sdk/partition/partition.py
MarkLindblad: I've now implemented a variant of your suggested change.

PR #1013, opened 2024-11-09
Add 120k max chars (default) for summarize_data

baitsguy:
Default 120k let's you work with 32k context window models. If we know we're using 128k token models, we should configure this explicitly.

### DIFF
--- a/lib/sycamore/sycamore/query/execution/operations.py
+++ b/lib/sycamore/sycamore/query/execution/operations.py
@@ -1,6 +1,8 @@
 import math
 from typing import Any, List, Union, Optional
 
+import structlog
+
 from sycamore import DocSet
 from sycamore.context import context_params, Context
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -128,6 +144,10 @@ def _get_text_for_summarize_data(
                         doc_text_representation += (element.text_representation or "") + "\n"
                 doc_text += f"Text contents:\n{doc_text_representation}\n"
 
+                if tokenizer is not None and max_tokens is not None:  # for mypy
+                    total_token_count = len(tokenizer.tokenize(text + doc_text))
+                    if total_token_count > max_tokens:
+                        break
```
mdwelsh, line 150: Add a logging .warn when this happens

mdwelsh (approved):

PR #1016, opened 2024-11-11
Fix up NTSB queries for query-eval tool.

mdwelsh:
The query-eval configs for NTSB queries were not using the few-shot examples for the planner and were resulting in bad query plans. This PR tries to fix this up. It also eliminates some duplicate queries and fixes the wording of a few queries to make them less ambiguous. 

I still need to evaluate the answers and query plans for ntsb-full, but this at least runs.


apps/query-eval/queryeval/driver.py
```diff
@@ -239,7 +241,7 @@ def do_plan(self, query: QueryEvalQuery, result: QueryEvalResult) -> QueryEvalRe
                 self.config.config.index,
                 self.data_schema,
                 examples=self.examples or None,
-                natural_language_response=self.config.config.natural_language_response or False,
+                natural_language_response=self.config.config.natural_language_response or True,
```
baitsguy, line 244: This might've broken an integ test, can you validate? Can merge after fixing or confirming

baitsguy (approved):

PR #1018, opened 2024-11-12
enable JSONifying tables

HenryL27:
lets `docset.write.json()` work when table extraction is on.

forgive the mypy shenanigans

### DIFF
--- a/lib/sycamore/sycamore/data/table.py
+++ b/lib/sycamore/sycamore/data/table.py
@@ -1,6 +1,6 @@
 from collections import OrderedDict
 from dataclasses import dataclass, field
-from typing import Any, Optional, TypeVar, Union, List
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/data/table.py
```diff
@@ -153,6 +172,18 @@ def from_dict(cls, dict_obj: dict[str, Any]) -> "Table":
 
         return ret
 
+    def to_dict(self) -> dict[str, Any]:
```
bsowell, line 175: Maybe worth adding a test that asserts that `Table.from_dict(table.to_dict()) == table` (for some value of ==)?

bsowell (approved):
I think it would be useful to have a simple unit test.

PR #1022, opened 2024-11-13
Data Model for tree structured Document Representation

bohou-aryn:


lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,388 @@
+from abc import abstractmethod, ABC
```
bsowell, line 1: Is your intention to merge this PR into Sycamore, or just to use it as a vehicle to discuss the data model? 

(To be clear, I want to get this into Sycamore, just want to get a sense of whether we should discuss the nuts and bolts now defer that).

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,388 @@
...
+    Picture = 7
+    SectionHeader = 8
+    Table = 9
+    Text = 10
+    Title = 11
+    Section = 101
+    Group = 102
+
+
+class Node:
```
bsowell, line 21: minor: Again we can defer this, but we should probably call this something like DocNode since we already use Node for the query plan.

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,388 @@
...
+        self._metadata = metadata
+
+    def node_id(self):
+        return self._node_id
+
+    @abstractmethod
+    def category(self):
+        pass
+
+    def metadata(self):
```
bsowell, line 33: Is this intended to replace properties? What is the signature, `dict[str, Any]`? Also, how would you update it if you want to add properties from like entity extraction or something?

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,388 @@
...
+class Internal(Node):
+    def __init__(self, node_id,  children, metadata):
+        super().__init__(node_id, metadata)
+        self._children = children
+
+    def children(self) -> list[Node]:
+        return self._children
+
+
+class Caption(Leaf):
```
bsowell, line 63: Certainly not critical, but would it make sense to create a parent class for all of the text-like leaf nodes (e.g. list items, text, etc.) I could imagine a lot of operations where you want to do something special for tables and images, but something common for all the text.

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,388 @@
...
+
+class Text(Leaf):
+    def __init__(self, node_id, box, content, metadata, continued=None):
+        super().__init__(node_id, box, content, metadata)
+        self._continued = continued
+
+    def category(self):
+        return Category.Caption
+
+    def continued(self):
```
bsowell, line 170: What are the semantics here? Why does Text have a continued field, but not, say, tables?

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,388 @@
...
+        self._header = header
+
+    def category(self):
+        return Category.Section
+
+    def accept(self, visitor):
+        return visitor.visit_section(self)
+
+
+class Visitor(ABC):
```
bsowell, line 214: Just curious, what was your rationale for using a visitor like this vs a traverser like we do for plan nodes (https://github.com/aryn-ai/sycamore/blob/9ddcaef57088b4430e02b4ec2aab276d5780a6e5/lib/sycamore/sycamore/plan_nodes.py#L10). To be clear, I don't think I have a strong opinion either way, just curious if you had a reason to prefer one over the other.

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,388 @@
...
+    @abstractmethod
+    def visit_caption(self, caption: Caption):
```
bsowell, line 216: Can the visit methods mutate the tree or are they purely read only? Do they return anything?

lib/sycamore/sycamore/document.py
bohou-aryn: yes, it's dict[str, Any]. Maybe not get your point, a visitor could help here.

lib/sycamore/sycamore/document.py
bohou-aryn: yes, sure.

lib/sycamore/sycamore/document.py
bohou-aryn: This is for the split objects. Table does have, it's just nothing is support there like merge across pages.

lib/sycamore/sycamore/document.py
bohou-aryn: It could return anything considering the dynamic typing, right?

lib/sycamore/sycamore/document.py
bohou-aryn: I'm not aware of this nodetraverse before. I looked a little bit on the printtraverse, seems it still either depend on modifying each node for polymorphism or squeeze in a big function handling all various node types.

lib/sycamore/sycamore/document.py
bohou-aryn: sure, or differentiate by package name.

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,390 @@
...
+        pass
+
+    @abstractmethod
+    def visit_group(self, group: Group):
+        pass
+
+    @abstractmethod
+    def visit_section(self, section: Section):
+        pass
+
```
HenryL27, line 267: I feel like a visit_generic or something would be useful. Something like
```
@abstractmethod
def visit_generic(self, node: Node):
    pass
```
and then the `visit_some_type` methods can default to calling visit_generic. I'm just looking at these and thinking it will be an absolute pain to have to explicitly handle each type when I wanna implement this interface ya know? Could save a bunch of code duplication in the NaiveSemanticVisitor eg

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,390 @@
...
+class Caption(Leaf):
+    def __init__(self, node_id, box, content, properties):
+        super().__init__(node_id, box, content, properties)
+
+    def category(self):
+        return Category.Caption
+
+    def accept(self, visitor):
+        return visitor.visit_caption(self)
```
eric-anderson, line 71: I would recommend moving the switch logic into the visitor class. 
In particular, I would have:
```
class Caption(Leaf):
   def __init__(self, node_id, box, content, properties):
        super().__init__(node_id, box, content, properties, category=Category.Caption)
        
class Leaf(...):
   def category(self):
        return self.category

   def accept(self, visitor):
       return visitor.visit(self, self.category) # optional to just have self since the category is bound

class Visitor():
   def visit(self, node, category):
        if category == Category.Caption:
            self.visit_caption(node)
       elif ...
```

That way someone can override just the visit function if they want to handle the grouping differently rather than having to implement a whole bunch of functions.

You may also want to implement more flexible visiting. For plan nodes we needed pre-traversal, post-traversal, in-place mutation, and single call for the entire tree.
See plan_nodes.py:NodeTraverse.  My guess is that if we're going to do mutations of the structure via traversal, we'll end up needing the same thing.

lib/sycamore/sycamore/document.py
```diff
@@ -0,0 +1,390 @@
...
+    def visit_group(self, group: Group):
+        pass
+
+    @abstractmethod
+    def visit_section(self, section: Section):
+        pass
+
+
+class NaiveSemanticVisitor(Visitor):
+    def visit_caption(self, caption):
```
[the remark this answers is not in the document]
bohou-aryn: Hey, thanks for your feedback. I actually have different opinions on this.
1. this does not align with a pre-existing design pattern
2. Considering code scalability, I feel there is more requirement to extend new functions/transformations instead of a new node type.
3. the visitor considers more about the inheritance, for the above naivesemanticvisitor example, it's possible to have requirement for a newer semanticvisitor which only modifies the function for handling table and leave all the other unchanged.

PR #1023, opened 2024-11-14
adding one shot prompting along with multimodal request

Soeb-aryn:
Making following changes
1. Adding one shot prompting 
2. Adding table image to the llm request for better key-value identification 
3. Reducing the total number of LLM request to one for each table.

### DIFF
--- a/lib/sycamore/sycamore/llms/prompts/default_prompts.py
+++ b/lib/sycamore/sycamore/llms/prompts/default_prompts.py
@@ -116,6 +116,7 @@ class GraphRelationshipExtractorPrompt(SimplePrompt):
 
     """
 
+
 class ExtractTablePropertiesPrompt(SimplePrompt):
     user = """
         You are given a text string represented as a CSV (comma-separated values) and an image of a table.
@@ -158,6 +159,7 @@ class ExtractTablePropertiesPrompt(SimplePrompt):
 
             return False
             """
+
 
 class EntityExtractorMessagesPrompt(SimplePrompt):
     def __init__(self, question: str, field: str, format: Optional[str], discrete: bool = False):
--- a/lib/sycamore/sycamore/llms/prompts/default_prompts.py
+++ b/lib/sycamore/sycamore/llms/prompts/default_prompts.py
@@ -116,38 +116,48 @@ class GraphRelationshipExtractorPrompt(SimplePrompt):
 
     """
 
-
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -119,36 +119,48 @@ class GraphRelationshipExtractorPrompt(SimplePrompt):
...
-            6. Ensure that key names are enclosed in double quotes.
-            7. return only the json object between ```
+        You are given a text string represented as a CSV (comma-separated values) and an image of a table.
+
+        Instructions:
+            Check if the table contains key-value pairs. A key-value pair table is a table where data is structured as key-value pairs. Generally, the first column contains the key and the second column contains the value. However, key-value pairs can also appear in other formats.
+            If there is a one-to-one mapping between two cells, even if the relationship is not direct, they should be considered key-value pairs.
+            If the table is a key-value pair table, return its key-value pairs as a JSON object.
+            If the table is not a key-value pair table, return False.
+            Parse the table, check the image, and return a flattened JSON object representing the key-value pairs from the table. The extracted key-value pairs should be formatted as a JSON object.
```
dhruvkaliraman7, line 129: Is L127 redundant if you already have this?

Soeb-aryn: LLM hallucinates without that line, reiterating simply solve the problem.

lib/sycamore/sycamore/transforms/extract_table_properties.py
```diff
@@ -65,30 +66,33 @@ def extract_table_properties(
...
         and populate them as a property of that element.
         """
-        prompt_find_table = prompt_find_table or ExtractTablePropertiesTablePrompt().user
-        query_agent = LLMTextQueryAgent(
-            prompt=prompt_find_table, llm=llm, output_property="keyValueTable", element_type="table"
-        )
-        query_agent.execute_query(parent)
+        doc1 = split_and_convert_to_image(parent)
+        img_list = []
+        for img in doc1:
```
dhruvkaliraman7, line 71: we are converting all elements to images, how costly is this operation?

dhruvkaliraman7 (approved):
Lgtm! Imma trust you on the test :)

PR #1027, opened 2024-11-17
Add streaming support and tests for query-server.

mdwelsh:
For our upcoming customer deployment, we need a REST server that will return query plans, results, and the set of retrieved docs used in a query. This PR adds this capability to the existing `query-server` app.

I have also added streaming responses using SSE, in order to support streaming back large result sets (and to ensure the query plan and status are streamed back to the client while the query is running). Streaming is enabled if `stream=true` in the query request, otherwise, a non-streamed response is returned.

I've added a unit test for the app as well.


apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,158 @@ async def get_index(
...
+    """
+
+    logger.info(f"Running query: {query}")
+
+    if query.query is None and query.plan is None:
+        raise ValueError("query or plan is required")
+    if query.query is not None and query.plan is not None:
+        raise ValueError("query and plan cannot both be specified")
+
+    if query.stream:
```
eric-anderson, line 223: I don't understand why we wouldn't always stream results (once we fix it to actually stream something).

However, if we don't want to, I would have thought this would become:
```
ret = []
for i in run_query_stream(query):
    ret.append(i)
return ret.join("")
```

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,158 @@ async def get_index(
...
+                    "data": sqresult.result,
+                }
+                await asyncio.sleep(0.1)
+
+            for doc in sqresult.retrieved_docs():
+                yield {
+                    "event": "retrieved_doc",
+                    "data": doc,
+                }
+                await asyncio.sleep(0.1)
```
eric-anderson, line 193: why are there all these sleeps in here?

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,158 @@ async def get_index(
...
+                "event": "status",
+                "data": "Running plan",
+            }
+            await asyncio.sleep(0.1)
+            sqresult = sqclient.run_plan(plan)
+            t1 = time.time()
+            num_results = 0
+            if isinstance(sqresult.result, DocSet):
+                logger.info("Got DocSet result")
+                for doc in sqresult.result.take_all():
```
eric-anderson, line 170: This is unfortunately effectively not streaming. take_all doesn't return an iterator, it returns a list, so the entire execution happens, and then you iterate here.
```
        for doc in Execution(self.context).execute_iter(self.plan, **kwargs):
            if include_metadata or not isinstance(doc, MetadataDocument):
                  ... do something here ...
```
Is what  you need to do if you want to make it more incremental.
However, given how ray seems to operate, this is only slightly more incremental for what we are doing because of how the opensearch reader works.
```

File: apps/query-server/queryserver/main.py [hunk 4]
```diff
@@ -67,29 +97,158 @@ async def get_index(
...
+                        await asyncio.sleep(0.1)
+            else:
+                num_results += 1
+                yield {
+                    "event": "result",
+                    "data": sqresult.result,
+                }
+                await asyncio.sleep(0.1)
+
+            for doc in sqresult.retrieved_docs():
```
eric-anderson, line 188: This turns out to not work. If the node has a summarize, then the results can get truncated and retrieved_docs as written does the wrong thing by returning partial results since the code forces materialize to run.  A fixed version that looks for materialize.success also fails because it finds no directory with complete results.
To get it to work, I stripped the summarize step off the end of the plan in the code in the customer environment since right now we only care about the set of documents.
Also for the customer this doesn't happen to return what they want, path is the wrong property to return.

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,158 @@ async def get_index(
...
+        try:
+            logger.info(f"Generating plan for {query.index}: {query.query}")
+            yield {
+                "event": "status",
+                "data": "Generating plan",
+            }
+            await asyncio.sleep(0.1)
+            plan = sqclient.generate_plan(query.query, query.index, sqclient.get_opensearch_schema(query.index))
+            logger.info(f"Generated plan: {plan}")
+            # Don't want to return these through the API.
```
eric-anderson, line 151: I'd recommend allowlisting rather than denylisting -- the plan can be really large (I don't know if this fixes it). For the current code in the customer I explicitly only returned the nodes which was always small along with some other small debugging information.

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,158 @@ async def get_index(
...
+            }
+            await asyncio.sleep(0.1)
+            sqresult = sqclient.run_plan(plan)
+            t1 = time.time()
+            num_results = 0
+            if isinstance(sqresult.result, DocSet):
+                logger.info("Got DocSet result")
+                for doc in sqresult.result.take_all():
+                    rendered = doc_to_json(doc)
+                    logger.info(f"Doc: {rendered}")
```
eric-anderson, line 172: For the customer you may not want to do this, there are ~100(?) properties on the documents, it goes on for a long time.

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,158 @@ async def get_index(
...
+            else:
+                num_results += 1
+                yield {
+                    "event": "result",
+                    "data": sqresult.result,
+                }
+                await asyncio.sleep(0.1)
+
+            for doc in sqresult.retrieved_docs():
+                yield {
```
eric-anderson, line 189: If you do this will the resulting thing be a valid json object?  We had to do a bit of magic in the model server to make sure the entire result was a single json object.

apps/query-server/pyproject.toml
```diff
@@ -10,6 +10,7 @@ python = ">=3.9,<3.13"
 sycamore-ai = { path = "../../lib/sycamore", develop = true, extras = ["opensearch", "local-inference"] }
 fastapi = { version = "^0.115.0", extras = ["standard"]}
 pydantic = "^2.9.2"
+sse-starlette = "^2.1.3"
```
eric-anderson, line 13: Make sure to re-lock in the root as well.

apps/query-server/queryserver/main.py
mdwelsh: Because the output formats are different in a streaming and non-streaming case. I don't want to give them SSE format in a non-streaming instance, just JSON. This is going to be a lot easier for them to consume.

apps/query-server/queryserver/main.py
mdwelsh: Right, we can fix this later, for now I want to get the protocol nailed down. I think some deeper changes are needed to Sycamore to do streaming properly.

apps/query-server/queryserver/main.py
mdwelsh: This does work for me, but even so, your comment is irrelevant to this PR. If we need a fix to `retrieved_docs` we can do that separately.

apps/query-server/queryserver/main.py
mdwelsh: I can make this fix later if needed but I am being expedient for now.

apps/query-server/queryserver/main.py
mdwelsh: I'll make it a .debug log statement then.

apps/query-server/queryserver/main.py
mdwelsh: Seems like the intent here is unclear. For at least the platform demo we need this API to be streaming, so query execution is not blocking things like returning the generated plan, and status updates from the server while the query is running. I expect our customers will want this as well, but I made streaming optional in case they just want to get back a single JSON response. The formats of the streamed and non-streamed responses are fundamentally different, so one cannot just do as you say and concatenate them together.

For this I am using SSE which is a "standard" in the sense that browsers support it natively, and OpenAI (and many others) use it to stream partial responses on a REST API. Unfortunately the docs on the web on SSE are kind of crappy but this page explains the protocol reasonably well: https://hpbn.co/server-sent-events-sse/#event-stream-protocol

However if one is not streaming, just returning JSON makes the most sense.

apps/query-server/pyproject.toml
mdwelsh: Done!

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,157 @@ async def get_index(
...
+    schema = sqclient.get_opensearch_schema(index)
     return Index(index=index, index_schema=schema)
 
 
 @app.post("/v1/plan")
 async def generate_plan(query: Query) -> LogicalPlan:
     """Generate a query plan for the given query, but do not run it."""
 
-    plan = sqclient.generate_plan(query.query, query.index, util.get_schema(sqclient, query.index))
+    if query.query is None:
```
eric-anderson, line 108: Once the validator is in, I think you only have to check one side of this, but I'm assuming that you can't get called by something that fails validation.

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,157 @@ async def get_index(
...
-    """Run the provided query plan."""
+def doc_to_json(doc: Document) -> Optional[dict[str, Any]]:
+    """Render a Document as a JSON object. Returns None for MetadataDocuments."""
+    NUM_TEXT_CHARS_GENERATE = 1024
+
+    if isinstance(doc, MetadataDocument):
+        return None
+
+    props_dict = {}
+    props_dict.update(doc.properties)
```
eric-anderson, line 125: I recently learned about copy.  In this case, I think props_dict = copy.copy(doc.properties)
does what you want.

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,157 @@ async def get_index(
...
+
+    plan = sqclient.generate_plan(query.query, query.index, sqclient.get_opensearch_schema(query.index))
     return plan
 
 
-@app.post("/v1/plan/run")
-async def run_plan(plan: LogicalPlan) -> SycamoreQueryResult:
-    """Run the provided query plan."""
+def doc_to_json(doc: Document) -> Optional[dict[str, Any]]:
+    """Render a Document as a JSON object. Returns None for MetadataDocuments."""
```
eric-anderson, line 118: I think the description should be 
```
"""Return the JSON transformation of the document. In particular non-internal properties and truncated text_representation."""
```
It's dropping a bunch of other document properties (embedding, docid, ...).

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,157 @@ async def get_index(
...
+    NUM_TEXT_CHARS_GENERATE = 1024
+
+    if isinstance(doc, MetadataDocument):
+        return None
+
+    props_dict = {}
+    props_dict.update(doc.properties)
+    if "_schema" in props_dict:
```
eric-anderson, line 126: You can write this as props_dict.pop("_schema", None)
if you want to save a line everywhere.

apps/query-server/queryserver/main.py
eric-anderson: This was confusion about what you wanted to return. At the time I was reading this I thought you wanted to return a single json object, but you don't you want to return a stream of json objects.  I think this comment is irrelevant now.

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,157 @@ async def get_index(
...
 
 @app.post("/v1/plan")
 async def generate_plan(query: Query) -> LogicalPlan:
     """Generate a query plan for the given query, but do not run it."""
 
-    plan = sqclient.generate_plan(query.query, query.index, util.get_schema(sqclient, query.index))
+    if query.query is None:
+        raise ValueError("query is required")
+    if query.plan is not None:
+        raise ValueError("plan must not be specified")
```
austintlee, line 111: Why is this necessary?  You don't do this check on query.plan in `run_query_stream`.

austintlee (approved):

apps/query-server/queryserver/main.py
eric-anderson: Got it. If I had to guess the problem is that it's running single threaded and the actual work is happening rather than output processing.  That likely means that the server can only really handle 1 request at a time (which is fine for the demo?) At some point you'll probably want to move execution to a separate thread and pass the events back with a queue. Take a look at the model server or the cloud server for an example of this approach.

apps/query-server/queryserver/main.py
eric-anderson: Ok. I'm not a fan of having two separate implementations of query execution, you have to fix both of them when you make changes.   For expediency it's worth getting in this way, but I'd recommend writing something that can combine the events into the json response. Something like:
```
ret = {}
for e in run_query_streamed():
   if e["event"] == "status":
       continue # drop, we're already done
   ret[e["event"]] = e["data"]
```
would mean  you have one implementation of run_query and one transform of events into a json return value. You could also choose at this point to stream back the single json event with a status block like the model server does, or do smarter combining of events, e.g. turn something into an array.


FWIW, I found the missing assertion that query was not none in the async version by comparing with the sync one.  There were also a bunch of differences in logging.

apps/query-server/queryserver/main.py
mdwelsh: generate_plan does not allow the plan to be specified up front. But run_query and run_query_stream do allow a pre-defined plan. Make sense?

apps/query-server/queryserver/main.py
```diff
@@ -67,29 +97,157 @@ async def get_index(
...
+        doc.text_representation[:NUM_TEXT_CHARS_GENERATE] if doc.text_representation is not None else None
+    )
+    return props_dict
+
+
+async def run_query_stream(query: Query) -> EventSourceResponse:
+    """Streaming version of run_query. Returns a stream of results as they are generated."""
+
+    async def query_runner():
+        try:
```
eric-anderson, line 142: Do you need the same 
if query.plan is None: 
   .. make plan ..
else:
  plan = query.plan
```
logic from the non-streaming version? 
I'd do either that or assert query.plan is not None.

**Review by eric-anderson — APPROVED** [review]

File: apps/query-server/queryserver/main.py [hunk 15]
```diff
@@ -3,23 +3,28 @@
...
+from sse_starlette.sse import EventSourceResponse
+from sycamore import DocSet
+from sycamore.data import Document, MetadataDocument
 from sycamore.query.client import SycamoreQueryClient
 from sycamore.query.logical_plan import LogicalPlan
-from sycamore.query.result import SycamoreQueryResult
 from sycamore.query.schema import OpenSearchSchema
 
-import queryserver.util as util
+logger = logging.getLogger("uvicorn.error")
```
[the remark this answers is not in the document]
mdwelsh: Thanks. TBH I don't know why this is the way it is -- I think I was following a tutorial that explained that this was the way to do FastAPI logging (?).

apps/query-server/queryserver/main.py
```diff
@@ -30,33 +35,58 @@ class Index(BaseModel):
...
     description: Optional[str] = None
+    """Description of the index."""
+
     index_schema: OpenSearchSchema
+    """The schema for this index."""
 
 
 class Query(BaseModel):
-    """Query an index with a given natural language query string."""
+    """Query an index with a given natural language query string. One of 'query' or 'plan' must be provided."""
```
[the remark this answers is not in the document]
mdwelsh: Yeah, I will do that. The docstrings are important because they get turned into the auto-generated web documentation for the API by FastAPI.

apps/query-server/queryserver/main.py
mdwelsh: Done

apps/query-server/queryserver/main.py
mdwelsh: Good idea

apps/query-server/queryserver/main.py
mdwelsh: Good catch, done.

apps/query-server/queryserver/main.py
mdwelsh: For now I'd like the plain JSON result to be very close to SycamoreQueryResult. I can refactor this later.

PR #1030, opened 2024-11-18
Fix SummarizeData so that downstream .materialize operations will work.

mdwelsh:
This should fix the issue where SummarizeData leaves us with an incomplete materialize dir when the total doc size exceeds the LLM token limit.

### DIFF
--- a/lib/sycamore/sycamore/query/execution/operations.py
+++ b/lib/sycamore/sycamore/query/execution/operations.py
@@ -162,6 +162,7 @@ def _get_text_for_summarize_data(
                             f" Sending text from {di + 1} docs."
                         )
                         done = True
+                        continue
                 text += doc_text + "\n"
         else:
             text += str(result_data) + "\n"
--- a/lib/sycamore/sycamore/query/execution/operations.py
+++ b/lib/sycamore/sycamore/query/execution/operations.py
@@ -27,7 +27,6 @@ BASE_PROPS = [
     "entity",
 ]
 
-NUM_DOCS_GENERATE = 60
 NUM_TEXT_CHARS_GENERATE = 2500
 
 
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -127,12 +127,18 @@ def _get_text_for_summarize_data(
 
         # consolidates relevant properties to give to LLM
         if isinstance(result, DocSet):
-            for i, doc in enumerate(result.take(NUM_DOCS_GENERATE, **kwargs)):
+            done = False
+            # For query result caching in the executor, we need to consume the documents
+            # so that the materialized data is complete, even if they are not all included
+            # in the input prompt to the LLM.
+            for di, doc in enumerate(result.take_all()):
                 if isinstance(doc, MetadataDocument):
```
eric-anderson, line 135: You shouldn't need MetadataDocument check. take_all removes those.

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -127,12 +127,18 @@ def _get_text_for_summarize_data(
 
         # consolidates relevant properties to give to LLM
         if isinstance(result, DocSet):
-            for i, doc in enumerate(result.take(NUM_DOCS_GENERATE, **kwargs)):
+            done = False
+            # For query result caching in the executor, we need to consume the documents
+            # so that the materialized data is complete, even if they are not all included
+            # in the input prompt to the LLM.
+            for di, doc in enumerate(result.take_all()):
```
eric-anderson, line 134: This works at small scale, but will blow up memory at large scale.
Approving as this is NTSB only, but I suggest a TODO or something.

lib/sycamore/sycamore/tests/unit/query/execution/test_sycamore_executor.py
```diff
@@ -102,3 +104,55 @@ def test_run_plan_with_caching(test_count_docs_query_plan, mock_sycamore_docsetr
...
+                            "ssl_assert_hostname": False,
+                            "ssl_show_warn": False,
+                            "timeout": 120,
+                        }
+                    },
+                }
+            )
+
+            # First run should populate cache.
+            executor = SycamoreExecutor(context, cache_dir=temp_dir)
```
eric-anderson, line 144: Is this test fast?  If it is, then fine to leave in unit tests, but if it's more than 5-10s, I'd like to get it moved into integration tests.  I initially thought this would use ray (which basically guarantees it's slow), but I'm no longer sure.

eric-anderson (approved):

PR #1032, opened 2024-11-21
Convert docids from UUID to NanoID.

alexaryn:


lib/sycamore/sycamore/connectors/file/file_writer.py
```diff
@@ -48,7 +48,7 @@ def default_filename(doc: Document, extension: Optional[str] = None) -> str:
         extension: An optional extension that will be appended to the name following a '.'.
     """
     if doc.doc_id is None:
-        base_name = str(uuid.uuid4())
+        base_name = "aryn:d-" + nanoid.generate()
```
bsowell, line 51: Should this call mkdocid? just in case we want to change again?

bsowell (approved):

PR #1033, opened 2024-11-21
heuristics to get title from section headers

Soeb-aryn:
If the first page of a document lacks a title, this heuristic identifies section headers or caption with the largest font size and promotes the most largest fontsize one to the title.

With the use_ocr flag enabled, the font size is determined based on the height of the bounding box. For Paddle OCR, each bounding box represents only a single line.
With the use_ocr flag enabled, the font size is obtained using PDFMiner.


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -124,16 +124,22 @@ def _supplement_text(inferred: list[Element], text: list[Element], threshold: fl
             if matched:
                 matches = []
                 full_text = []
+                font_sizes = []
                 for m in matched:
                     matches.append(m)
                     if m.text_representation:
                         full_text.append(m.text_representation)
-
+                        if m.properties.get("font_size") is not None:
```
karanataryn, line 132: Replace with `if (font_size := m.properties.get("font_size"))` to reduce hash lookups

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -124,16 +124,22 @@ def _supplement_text(inferred: list[Element], text: list[Element], threshold: fl
...
+                            font_sizes.append(m.properties.get("font_size"))
                 if isinstance(i, TableElement):
                     i.tokens = [{"text": elem.text_representation, "bbox": elem.bbox} for elem in matches]
 
                 i.data["text_representation"] = " ".join(full_text)
-
+                i.properties["font_size"] = 0
+                if len(font_sizes) > 0:
+                    font_sizes_filtered = [size for size in font_sizes if size is not None]
```
karanataryn, line 140: Why do we need to filter if we only add if it is not None? We shouldn't need to do the conditional checks twice.

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -124,16 +124,22 @@ def _supplement_text(inferred: list[Element], text: list[Element], threshold: fl
...
                         full_text.append(m.text_representation)
-
+                        if m.properties.get("font_size") is not None:
+                            font_sizes.append(m.properties.get("font_size"))
                 if isinstance(i, TableElement):
                     i.tokens = [{"text": elem.text_representation, "bbox": elem.bbox} for elem in matches]
 
                 i.data["text_representation"] = " ".join(full_text)
-
+                i.properties["font_size"] = 0
```
karanataryn, line 138: Wouldn't `None` be better?

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -157,6 +163,7 @@ def partition_pdf(
         output_format: Optional[str] = None,
         text_extraction_options: dict[str, Any] = {},
         source: str = "",
+        promote_title: bool = False,
```
karanataryn, line 166: Can you check this from the `text_extraction_options`? You'll also need to add documentation in the top level partition call to say that `promote_title` is an accepted argument in `text_extraction_options` and explain its functionality.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -454,6 +455,7 @@ def __init__(
         output_format: Optional[str] = None,
         text_extraction_options: dict[str, Any] = {},
         source: str = "",
+        promote_title: bool = False,
```
karanataryn, line 458: Move to `text_extraction_options`

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -494,6 +496,7 @@ def __init__(
         self._pages_per_call = pages_per_call
         self._text_extraction_options = text_extraction_options
         self._source = source
+        self._promote_title = promote_title
```
karanataryn, line 499: As above.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -523,6 +526,7 @@ def partition(self, document: Document) -> Document:
                 output_format=self._output_format,
                 text_extraction_options=self._text_extraction_options,
                 source=self._source,
+                promote_title=self._promote_title,
```
karanataryn, line 529: As above.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -140,8 +142,8 @@ def __init__(self):
         self.tesseract = Tesseract()
         self.easy_ocr = EasyOcr()
 
-    def get_text(self, image: Image.Image) -> str:
-        return self.tesseract.get_text(image)
+    def get_text(self, image: Image.Image) -> tuple[str, float]:
+        return self.tesseract.get_text(image), 0.0
```
karanataryn, line 146: `self.tesseract.get_text(image)` returns a tuple.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -109,9 +111,9 @@ def __init__(self):
 
         self.pytesseract = pytesseract
 
-    def get_text(self, image: Image.Image) -> str:
+    def get_text(self, image: Image.Image) -> tuple[str, float]:
         val = self.pytesseract.image_to_string(image)
-        return val
+        return val, 0.0
```
karanataryn, line 116: Can you add a comment that is not implemented? If it's not possible to calculate it from the output of tesseract, I would rather we return `None` to indicate this is not a valid font size

lib/sycamore/sycamore/transforms/text_extraction/pdf_miner.py
```diff
@@ -88,6 +88,20 @@ def extract_document(self, filename: str, hash_key: str, use_cache=False, **kwar
                 pdf_miner_cache.set(hash_key, pages)
             return pages
 
+    @staticmethod
+    def _parse_obj(objs):
```
karanataryn, line 92: Can you rename this to `get_font_size` or something like it? `parse_obj` seems inaccurate for the functionality here.

lib/sycamore/sycamore/utils/pdf_utils.py
```diff
@@ -162,3 +162,23 @@ def display_page_and_table_properties(some_pages: list[Document]):
                 print("Element Type: ", e.type)
                 print("Element Properties: ", json.dumps(e.properties, indent=2, default=str))
                 display(HTML(e.text_representation))
+
+
+def promote_sectionheader_to_title(elements: list[Element]) -> list[Element]:
```
karanataryn, line 167: Why don't we modify this to promote any set of elements to a title and take it in as an argument? You can set the default to be `["Section-header", "Caption"]` and call the function `promote_elements`. This would make it more widely applicable.

lib/sycamore/sycamore/transforms/detr_partitioner.py
Soeb-aryn: done

Soeb-aryn: adding a 'output_label_options' option for driving all output label related functionalities.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
Soeb-aryn: done

lib/sycamore/sycamore/utils/pdf_utils.py
Soeb-aryn: good call, updating the function

lib/sycamore/sycamore/transforms/detr_partitioner.py
Soeb-aryn: makes sense, shouldn't be needed here

karanataryn (commented):
Looking better. We will need to add the documentation for this in this PR or a later one, I'll leave that to you.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -419,6 +419,10 @@ class ArynPartitioner(Partitioner):
              either pdfminer or OCR. Currently supports the 'object_type' property for pdfminer,
              which can be set to 'boxes' or 'lines' to control the granularity of output.
         source: The application that is using the partitioner. This is used for logging purposes.
+        output_label_options: A dictionary for configuring output label behavior. It supports two options: 
+        promote_title, a boolean that specifies whether to add a title to partitioned elements if one is missing, and
+            title_candidate_elements, a list of strings representing labels for potential titles.
+            default: {"promote_title": True ,  "title_candidate_elements":["Section-header", "Caption"]}
```
karanataryn, line 425: I'm not sure if it makes sense to have `title_candidate_elements` as a top-level attribute given that we only care about it if `promote_title` is true but this is similar to `use_ocr` and `ocr_images` so it's not worth blocking on this.

karanataryn (approved):
Looks good to me. Could you add a small unit test to ensure we behave as intended? It should be relatively simple.

lib/sycamore/sycamore/tests/unit/utils/test_pdf_utils.py
```diff
@@ -156,3 +157,50 @@ def test_select_pages():
...
+        Element(type="Caption", properties={"page_number": 2, "font_size": 16}),
+    ]
+    gt_elements = [
+        Element(type="Title", properties={"page_number": 1, "font_size": 12}),
+        Element(type="Section-header", properties={"page_number": 2, "font_size": 14}),
+        Element(type="Caption", properties={"page_number": 2, "font_size": 16}),
+    ]
+
+    result = promote_title(elements)
+    print(result)
```
karanataryn, line 187: Need to remove this

lib/sycamore/sycamore/transforms/partition.py
MarkLindblad, line 425: The default specified in this docstring is wrong. The default for promote_title is False, not True.
https://github.com/aryn-ai/sycamore/blob/1f053475507aa249d3aa839e93e3a3f67027c3c8/lib/sycamore/sycamore/transforms/detr_partitioner.py#L205

MarkLindblad, line 425: Let's make it more clear that this is only done on the first page.

https://github.com/aryn-ai/sycamore/blob/59f8f2c95eb304365d6857fd756babdc30e22e88/lib/sycamore/sycamore/utils/pdf_utils.py#L171-L172

MarkLindblad: This has been fixed in #1057

MarkLindblad: This has been fixed in #1057

PR #1036, opened 2024-11-22
Use NanoIDs in file_scan.

alexaryn:


lib/sycamore/sycamore/connectors/file/file_scan.py
```diff
@@ -287,7 +287,7 @@ def __init__(
     def _to_document(self, json_dict: dict[str, Any]) -> list[dict[str, Any]]:
         document = Document()
 
-        document.doc_id = str(uuid.uuid1())
+        document.doc_id = mkdocid()
```
bsowell, line 290: This one doesn't get a scheme?

bsowell (approved):

lib/sycamore/sycamore/connectors/file/file_scan.py
alexaryn: Defaults to "d".

PR #1037, opened 2024-11-22
extract table properties prompt & bug fix

Soeb-aryn:
Changes:
1. Updated the prompt to support key-value tables where all keys are set to null.
2. Fixed a bug causing the raw_answer variable to reset improperly.

### DIFF
--- a/lib/sycamore/sycamore/llms/prompts/default_prompts.py
+++ b/lib/sycamore/sycamore/llms/prompts/default_prompts.py
@@ -126,9 +126,9 @@ class ExtractTablePropertiesPrompt(SimplePrompt):
             If there is a one-to-one mapping between two cells, even if the relationship is not direct, they should be considered key-value pairs.
             If the table is a key-value pair table, return its key-value pairs as a JSON object.
             If the table is not a key-value pair table, return False.
-            Parse the table, check the image, and return a flattened JSON object representing the key-value pairs from the table. The extracted key-value pairs should be formatted as a JSON object.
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -126,9 +126,10 @@ class ExtractTablePropertiesPrompt(SimplePrompt):
             If there is a one-to-one mapping between two cells, even if the relationship is not direct, they should be considered key-value pairs.
             If the table is a key-value pair table, return its key-value pairs as a JSON object.
             If the table is not a key-value pair table, return False.
-            Parse the table, check the image, and return a flattened JSON object representing the key-value pairs from the table. The extracted key-value pairs should be formatted as a JSON object.
+            Use camelCase for the key names in the JSON object.
+            Parse the CSV table, check the image, and return a flattened JSON object representing the key-value pairs from the table. The extracted key-value pairs should be formatted as a JSON object.
             Do not return nested objects; keep the dictionary only one level deep. The only valid value types are numbers, strings, None, and lists.
-            Use camelCase for the key names.
+            A table can have multiple or all null values for a key. In such cases, return a JSON object with the specified key set to null for all rows in the table.
```
bohou-aryn, line 132: Do we need to handle empty cell case specially? in the prompt we sent to the llm, I would expect if our table representation could indicate the difference.

Soeb-aryn: I tried without explicitly mentioning this, but it didn’t work. Adding this line ensures it works consistently.

bohou-aryn (approved):

PR #1038, opened 2024-11-26
Convert DocIDs to UUIDs for Qdrant & Weaviate; unit tests.

alexaryn:


bsowell (approved):
TBH I haven't gone through the conversion in tremendous detail, but the tests look reasonable and everything is green so LGTM.

PR #1039, opened 2024-11-26
Added ragas to compute string metrics for evaluation.

akarshgupta7:


apps/query-eval/queryeval/driver.py
```diff
@@ -476,6 +551,13 @@ def eval_all(self):
 
     def run(self):
         """Run all stages."""
+        # Define scorers
+        bleu_scorer = BleuScore()
```
baitsguy, line 555: can you move this to the __init__ method? let's people configure these easily too then

apps/query-eval/queryeval/driver.py
```diff
@@ -371,11 +400,38 @@ def do_eval(self, query: QueryEvalQuery, result: QueryEvalResult) -> QueryEvalRe
...
                 console.print(f"Missing docs: {expected_doc_set - retrieved_doc_set})")
             metrics.doc_retrieval_recall = len(retrieved_doc_set & expected_doc_set) / len(expected_doc_set)
-            metrics.doc_retrieval_precision = len(retrieved_doc_set & expected_doc_set) / len(result.retrieved_docs)
+            metrics.doc_retrieval_precision = len(retrieved_doc_set & expected_doc_set) / len(retrieved_doc_set)
 
-        # Evaluate result
-        if not result.result:
-            console.print("[yellow] No query execution result available, skipping..", style="italic")
+        # Evaluate string metrics
+        if not query.expected:
```
baitsguy, line 406: this function is large enough now that it might make sense to separate the specific metric types into separate methods (doc_retrieval metrics, generated_answer metrics)

apps/query-eval/queryeval/driver.py
```diff
@@ -371,11 +400,38 @@ def do_eval(self, query: QueryEvalQuery, result: QueryEvalResult) -> QueryEvalRe
...
+                "[yellow] No query execution result available, skipping..",
+                style="italic",
+            )
+        else:
+            if isinstance(query.expected, str) and isinstance(result.result, str):
+                sample = SingleTurnSample(
+                    response=result.result,
+                    reference=query.expected,
+                )
+                scores = asyncio.run(
```
baitsguy, line 419: do we really need to make compute_string_metrics an async method? Can we just have it be sync and have it wait for processing within? 

This is adding a little complexity and I'm unsure why

baitsguy (commented):
Couple small refactoring requests

apps/query-eval/queryeval/driver.py
akarshgupta7: Yeah no idea how I missed that. Thanks!

apps/query-eval/queryeval/driver.py
akarshgupta7: I looked into this as well for a long time. Ragas method to get score is async because of which we have to do this. Do you have any suggestions?

apps/query-eval/queryeval/driver.py
akarshgupta7: I have made these changes.

apps/query-eval/queryeval/driver.py
baitsguy: You can do a variant off https://chatgpt.com/share/67466a08-1488-8007-b4fa-2d2538c74071

Basically do the waiting for async completion inside the compute_string_metrics method

akarshgupta7: There will be no waiting as such because do_eval runs one sample at a time. Ragas people have made it an async function to handle large workloads.

akarshgupta7: Regardless I have made the changes.

baitsguy: Right I just mean in terms of function APIs

apps/query-eval/queryeval/driver.py
```diff
@@ -371,11 +395,65 @@ def do_eval(self, query: QueryEvalQuery, result: QueryEvalResult) -> QueryEvalRe
                 console.print("[red]:x: Documents retrieved don't include all expected docs")
                 console.print(f"Missing docs: {expected_doc_set - retrieved_doc_set})")
             metrics.doc_retrieval_recall = len(retrieved_doc_set & expected_doc_set) / len(expected_doc_set)
-            metrics.doc_retrieval_precision = len(retrieved_doc_set & expected_doc_set) / len(result.retrieved_docs)
+            metrics.doc_retrieval_precision = len(retrieved_doc_set & expected_doc_set) / len(retrieved_doc_set)
+        return metrics
+
+    def get_string_metrics(
```
baitsguy, line 401: nit: generated_answer_metrics

baitsguy (approved):

PR #1040, opened 2024-11-26
Fix sort so that it works with an unspecified or None default_value.

eric-anderson:
The core of the problem is that the ray sort operator requires that each document be associated with a value of the same type. Ray sort doesn't let us specify a comparator, and we don't know the type. Documents can be missing the sort key,
metadata documents are mostly guaranteed to be missing the sort key.

* In the sycamore sort operator, drop documents that do not have the sort parameter if there is no default value.
* In the Luna sort operator accept a default None value for the default_value.
* Split the test into a local-mode (0.04s) unit test and ray (21.25s) integration test
* Fix the tests to check the new semantics, previously they checked the "crashes if no default value specified" semantics.

### DIFF
--- a/lib/sycamore/sycamore/tests/unit/query/execution/test_sycamore_operator.py
+++ b/lib/sycamore/sycamore/tests/unit/query/execution/test_sycamore_operator.py
@@ -331,20 +331,23 @@ def test_llm_extract_entity():
 
 
 def test_sort():
-    context = sycamore.init()
-    doc_set = Mock(spec=DocSet)
-    return_doc_set = Mock(spec=DocSet)
-    doc_set.sort.return_value = return_doc_set
-    logical_node = Sort(node_id=0, descending=True, field="properties.counter", default_value=0)
-    sycamore_operator = SycamoreSort(context, logical_node, query_id="test", inputs=[doc_set])
-    result = sycamore_operator.execute()
+    Sort(node_id=0, descending=True, field="no-default-value")
 
-    doc_set.sort.assert_called_once_with(
-        descending=logical_node.descending,
-        field=logical_node.field,
-        default_val=logical_node.default_value,
-    )
-    assert result == return_doc_set
+    for default_value in [None, 0]:
+        context = sycamore.init()
+        doc_set = Mock(spec=DocSet)
+        return_doc_set = Mock(spec=DocSet)
+        doc_set.sort.return_value = return_doc_set
+        logical_node = Sort(node_id=0, descending=True, field="properties.counter", default_value=default_value)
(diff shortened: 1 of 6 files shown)


lib/sycamore/sycamore/tests/unit/query/execution/test_sycamore_operator.py
```diff
@@ -331,20 +331,23 @@ def test_llm_extract_entity():
...
 
 def test_sort():
-    context = sycamore.init()
-    doc_set = Mock(spec=DocSet)
-    return_doc_set = Mock(spec=DocSet)
-    doc_set.sort.return_value = return_doc_set
-    logical_node = Sort(node_id=0, descending=True, field="properties.counter", default_value=0)
-    sycamore_operator = SycamoreSort(context, logical_node, query_id="test", inputs=[doc_set])
-    result = sycamore_operator.execute()
+    Sort(node_id=0, descending=True, field="no-default-value")
```
baitsguy, line 334: what's this testing?

lib/sycamore/sycamore/transforms/sort.py
```diff
@@ -6,6 +6,31 @@
 if TYPE_CHECKING:
     from ray.data import Dataset
 
+from sycamore.plan_nodes import UnaryNode
+
+
+class DropIfMissingField(UnaryNode):
```
baitsguy, line 12: Can this not just be a filter function?

lib/sycamore/sycamore/tests/unit/transforms/test_sort.py
```diff
@@ -61,25 +69,36 @@ def test_metadata_document(self):
...
-        for i in range(len(sorted_doc_list)):
-            if i == 0 or i == 1:
-                assert isinstance(sorted_doc_list[i], MetadataDocument)
-            elif i == 2:
-                assert sorted_doc_list[i].text_representation == "B"
-            elif i == 3:
-                assert sorted_doc_list[i].text_representation == "C"
-            elif i == 4:
-                assert sorted_doc_list[i].text_representation == "Z"
+        for i in range(3):
```
baitsguy, line 88: can you stick a comment explaining where the 3rd metadata document is coming from

baitsguy (commented):
Looks good, question about we need DropIfMissingField or can just have a filter function

lib/sycamore/sycamore/transforms/sort.py
eric-anderson: Nope, it needs to remove metadata documents also and there isn't an easy way to get at metadata as it flows through (by design you're not supposed to do what I'm doing)

baitsguy (approved):

baitsguy (approved):

PR #1051, opened 2024-12-03
Rename Embedding in ElasticSearch Notebook

karanataryn:
There's some extra diff here from how my notebook is structured, don't think it should affect execution.


AbhijitP-009 (approved):
just add a comment at the top with pip install info

PR #1053, opened 2024-12-04
Add deformable table extractor

HenryL27:
1. Abstract out the structure model initialization so I can easily swap a Deformable DETR in.
2. Add apply_thresholds param to extract to control whether to threshold the objects by score from the model
3. Change behavior when there are multiple table objects to take the one with the highest score rather than whatever's first.
4. Add subclass that uses Deformable DETR.


bsowell (approved):
LGTM

PR #1055, opened 2024-12-04
move deformable detr safe loading code

HenryL27:
something something lockfile something something deduplication

### DIFF
--- a/lib/sycamore/sycamore/utils/model_load.py
+++ b/lib/sycamore/sycamore/utils/model_load.py
@@ -1,0 +1,33 @@
+from sycamore.utils.import_utils import requires_modules
+from sycamore.utils.time_trace import LogTime
+import fasteners
+from pathlib import Path
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from transformers import DeformableDetrForObjectDetection
+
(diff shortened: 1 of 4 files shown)


alexaryn (commented):
I suspect the `to()` needs to be inside the lock.  Without the `to()` I don't think it knows what device to target.  Without that, I don't see how it could do JITC for the correct GPU instruction set.

lib/sycamore/sycamore/utils/model_load.py
```diff
@@ -0,0 +1,33 @@
...
+if TYPE_CHECKING:
+    from transformers import DeformableDetrForObjectDetection
+
+_DETR_LOCK_FILE = f"{Path.home()}/.cache/Aryn-Detr.lock"
+
+
+@requires_modules("transformers", "local_inference")
+def load_deformable_detr(model_name_or_path, device) -> "DeformableDetrForObjectDetection":
+    """Load deformable detr without getting concurrency issues in
+    jit-ing the deformable attention kernel.
```
alexaryn, line 17: It really should be `jitc`.

alexaryn (approved):
Seems good to me.  Hope it works.

lib/sycamore/sycamore/utils/model_load.py
HenryL27: thank you for the commit message setup

PR #1056, opened 2024-12-05
Handle military clock time (0800) in time standardizer.

alexaryn:


lib/sycamore/sycamore/tests/unit/transforms/test_standardizer.py
```diff
@@ -266,3 +267,41 @@ def test_ignore_errors_key_missing(self):
...
+        ("March 17, 2023, 14.25 Local", "2023-03-17 14:25:00"),
+        ("March 17, 2023, 14.25", "2023-03-17 14:25:00"),
+        ("March 17, 2023 14:25:00", "2023-03-17 14:25:00"),
+        ("March 17, 2023 2:25PM", "2023-03-17 14:25:00"),
+        ("March 17, 2023 2:25AM", "2023-03-17 02:25:00"),
+        ("17 March 2023 14:25", "2023-03-17 14:25:00"),
+        ("2023-07-15 10.30.00", "2023-07-15 10:30:00"),
+        ("15/07/2023 10.30.00", "2023-07-15 10:30:00"),
+        ("2023-07-15 10.30.00 Local", "2023-07-15 10:30:00"),
+        ("2023-07-15 10.30.00PDT", "2023-07-15 10:30:00-07:00"),
```
karanataryn, line 284: Why do we need to preserve the timezone? Why not convert to UTC time?

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -185,6 +185,9 @@ class DateTimeStandardizer(Standardizer):
     """
 
     DEFAULT_FORMAT = "%B %d, %Y %H:%M:%S%Z"
+    clock_re = re.compile(r"\d:[0-5]\d")
+    year_re = re.compile(r"([12]\d\d\d-)|(/[12]\d\d\d)|(\d/[0-3]?\d/\d)")
+    digitpair_re = re.compile(r"([0-2]\d)([0-5]\d)(\d\d)?")
```
karanataryn, line 190: Would appreciate comments for the regexes

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -205,10 +208,11 @@ def fixer(raw_dateTime: str) -> datetime:
         """
         assert raw_dateTime is not None, "raw_dateTime is None"
         try:
-            raw_dateTime = raw_dateTime.strip()
+            raw_dateTime = DateTimeStandardizer.preprocess(raw_dateTime)
             raw_dateTime = raw_dateTime.replace("Local", "")
             raw_dateTime = raw_dateTime.replace("local", "")
             raw_dateTime = raw_dateTime.replace(".", ":")
+            logging.error(f"FIXME {raw_dateTime}")
```
karanataryn, line 215: Why do we have this?

lib/sycamore/sycamore/transforms/standardizer.py
```diff
@@ -222,6 +226,35 @@ def fixer(raw_dateTime: str) -> datetime:
             # Handle any other exceptions
             raise RuntimeError(f"Unexpected error occurred while processing: {raw_dateTime}") from e
 
+    @staticmethod
+    def preprocess(raw: str) -> str:
```
karanataryn, line 230: This looks like it's only meant for military time, so I'd rather rename this to `miltiary_time_preprocess`, given that we're doing some processing in the main function as well.

karanataryn (commented):
Comments and suggestions. This helps handle a new case, though I suspect we will need to use LLMs to handle general cases soon.

lib/sycamore/sycamore/tests/unit/transforms/test_standardizer.py
alexaryn: It's not my decision to make.  I assume the reason is that sometimes the timezone isn't specified in either the document or the query and it's better to be in a position to do late binding.  The way datetime and dateparser work is that timezone can be specified if known and will be used if specified.  It's easy to convert a datetime to UTC if that's the goal.

lib/sycamore/sycamore/transforms/standardizer.py
alexaryn: Oops.

lib/sycamore/sycamore/transforms/standardizer.py
alexaryn: I had hoped the names were descriptive.  Do you want examples of what they match?  Not sure how to make regexes clearer without just explaining the grammar of regexes, which is already better done on various web pages.

lib/sycamore/sycamore/transforms/standardizer.py
alexaryn: Fine.  I was thinking other preprocessing might go here too.  Either way.

lib/sycamore/sycamore/transforms/standardizer.py
karanataryn: Examples would be good, or even a short explanation of their functionality. It is very dense right now, and not easy to understand without trying to make sense of the regex itself.

lib/sycamore/sycamore/transforms/standardizer.py
karanataryn: We can always rename it at that point.

lib/sycamore/sycamore/tests/unit/transforms/test_standardizer.py
karanataryn: But if we want to 'standardize' it, isn't standardizing to one timezone part of the process? It seems like that's likely to reduce errors as well in that case. Also, how does OpenSearch do with timezones?

PR #1060, opened 2024-12-05
don't die if box is None in try_draw_boxes

HenryL27:
we got a table with a cell with no bbox. looking into where that came from but for now let's not crash if we can't get a bbox.

```python
doc = "s3://aryn-public/ntsb/FTW95FA129.pdf"
ds = ctx.read.binary(paths = doc, binary_format = "pdf").partition(ArynPartitioner(extract_table_structure=True, extract_images=True))
show_pages(ds)
```


bsowell (approved):
Weird. Curious where it came from, but this seems like a reasonable defensive approach.

PR #1061, opened 2024-12-06
when there's no table structure, take the token bbox for the cell bbox

HenryL27:
Found the codepath where we don't attach a bbox to the table cells

### DIFF
--- a/lib/sycamore/sycamore/transforms/table_structure/extract.py
+++ b/lib/sycamore/sycamore/transforms/table_structure/extract.py
@@ -174,6 +174,8 @@ class TableTransformerStructureExtractor(TableStructureExtractor):
 
         # Convert the raw objects to our internal table representation. This involves multiple
         # phases of postprocessing.
+        element.properties["backup_tokens"] = tokens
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -174,6 +174,8 @@ def extract(
 
         # Convert the raw objects to our internal table representation. This involves multiple
         # phases of postprocessing.
+        element.properties["backup_tokens"] = tokens
+        element.properties["object"] = objects
```
bsowell, line 178: Are these net new properties that we are adding to table elements? Are they something we really want to expose to sycamore users?

bsowell: I know we remove tokens in DocParse. Would we need to remove these props too?

HenryL27: oh whoops, those are in from debugging. will drop

bsowell (approved):
lgtm

PR #1064, opened 2024-12-06
Fix `split_elements` for table elements with no `elem.table` attribute

MarkLindblad:
Previously, splitting documents that contained tables after partitioning them without using `extract_table_structure` could cause an exception to be raised when splitting a table element. This fix allows those documents to be split without error.


HenryL27 (approved):
lgtm

PR #1067, opened 2024-12-09
Fix Extract Schema Empty Return

karanataryn:
Extract schema would previously return a string with `None` in failure cases which is incompatible with the normal dictionary return type. This was causing issues when writing out to connectors, and is fixed here.


AbhijitP-009 (approved):
lgtm

PR #1069, opened 2024-12-09
Add logging of the full exception in base_writer.

eric-anderson:
Without this, tracking down what went wrong is much harder.


HenryL27 (approved):
lgtm

PR #1076, opened 2024-12-13
Merge-elements-LLM-filter

dhruvkaliraman7:
Modified LLM filter to combine elements up til token limit to make fewer LLM calls. (Previously 1 call per element)

### DIFF
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -3,6 +3,7 @@ import logging
 from pathlib import Path
 import pprint
 import sys
+import copy
 from typing import Callable, Optional, Any, Iterable, Type, Union, TYPE_CHECKING
 import re
 
@@ -32,6 +33,7 @@ from sycamore.transforms.merge_elements import ElementMerger
 from sycamore.utils.extract_json import extract_json
 from sycamore.transforms.query import QueryExecutor, Query
 from sycamore.materialize_config import MaterializeSourceMode
+from sycamore.functions.tokenizer import Tokenizer
 
 if TYPE_CHECKING:
     from sycamore.writer import DocSetWriter
@@ -984,6 +986,8 @@ class DocSet:
         use_elements: bool = False,
         similarity_query: Optional[str] = None,
         similarity_scorer: Optional[SimilarityScorer] = None,
+        max_tokens: Optional[int] = 512,
+        tokenizer: Optional[Tokenizer] = None,
         **resource_args,
     ) -> "DocSet":
         """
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/docset.py
```diff
@@ -1028,23 +1031,59 @@ def threshold_filter(doc: Document, threshold) -> bool:
...
-                score = int(re.findall(r"\d+", element.properties[new_field])[0])
-                # we're storing the element_index of the element that provides the highest match score for a document.
-                doc_source_field_name = f"{new_field}_source_element_index"
-                if score >= doc.get(doc_source_field_name, 0):
-                    doc.properties[f"{new_field}"] = score
-                    doc.properties[f"{new_field}_source_element_index"] = element.element_index
-                if score >= threshold:
-                    return True
-                evaluated_elements += 1
+            if tokenizer and max_tokens and field == "text_representation":
```
baitsguy, line 1035: I don't think you want custom logic for text_representation (a proxy of how data is stored in the document) here. I'd just use whichever property the function call requests. If every element is expected to have the same value for a given property, the caller is going to be setting use_elements to False and we won't get here

lib/sycamore/sycamore/docset.py
```diff
@@ -1028,23 +1031,59 @@ def threshold_filter(doc: Document, threshold) -> bool:
...
+                ind = 0
+                while ind < len(doc.elements):
+                    combined_text = ""
+                    window_indices = []
+                    current_tokens = 0
+                    for element in doc.elements[ind:]:
+                        txt = element.text_representation
```
baitsguy, line 1042: you'll want to update this to field_to_value

lib/sycamore/sycamore/docset.py
```diff
@@ -1028,23 +1031,59 @@ def threshold_filter(doc: Document, threshold) -> bool:
...
+                        if not txt:
+                            ind += 1
+                            continue
+                        element_tokens = len(tokenizer.tokenize(txt))
+                        if current_tokens + element_tokens > max_tokens and current_tokens != 0:
+                            break
+                        combined_text += element[field] + " "
```
baitsguy, line 1049: can you prefix this text with some element level properties (if they exist?). At least element type, index, and page_number. This is because elements are probably going to be out of order after score sorting

lib/sycamore/sycamore/docset.py
```diff
@@ -1028,23 +1031,59 @@ def threshold_filter(doc: Document, threshold) -> bool:
...
+                    e_doc = Document(dummy_element.data)
+                    e_doc = entity_extractor.extract_entity(e_doc)
+                    element.properties[new_field] = e_doc.properties[new_field]
+                    # todo: move data extraction and validation to entity extractor
+                    score = int(re.findall(r"\d+", element.properties[new_field])[0])
+                    # storing the element_index of the element(s) that provides the highest match score for a document.
+                    doc_source_field_name = f"{new_field}_source_element_index"
+                    if score >= doc.get(doc_source_field_name, 0):
+                        doc.properties[f"{new_field}"] = score
+                        doc.properties[f"{new_field}_source_element_index"] = window_indices
```
baitsguy, line 1064: Nice

lib/sycamore/sycamore/docset.py
```diff
@@ -1028,23 +1031,59 @@ def threshold_filter(doc: Document, threshold) -> bool:
...
+                        ind += 1
+                    dummy_element = copy.deepcopy(element)
+                    dummy_element[field] = combined_text
+                    e_doc = Document(dummy_element.data)
+                    e_doc = entity_extractor.extract_entity(e_doc)
+                    element.properties[new_field] = e_doc.properties[new_field]
+                    # todo: move data extraction and validation to entity extractor
+                    score = int(re.findall(r"\d+", element.properties[new_field])[0])
+                    # storing the element_index of the element(s) that provides the highest match score for a document.
+                    doc_source_field_name = f"{new_field}_source_element_index"
```
baitsguy, line 1061: can we also store the score in the source elements for the dummy_element? and also probably window_indices. This is so we can look at the score an element generated, and what elements it was sent as part of

baitsguy (commented):
Unit tests :)

baitsguy (approved):

PR #1077, opened 2024-12-16
Misc table transformers post-processing

HenryL27:
1. For thresholding, if we would eliminate all rows or columns instead set the threshold to the max-confidence one minus epsilon
2. For slot_into_containers, if a package is contained equally in two containers (usually this means that the package is fully contained in both), slot it into the smaller container.
3. Use the biggest "table" class object, rather than the most confident.

### DIFF
--- a/lib/sycamore/sycamore/transforms/table_structure/table_transformers.py
+++ b/lib/sycamore/sycamore/transforms/table_structure/table_transformers.py
@@ -53,6 +53,23 @@ def apply_class_thresholds(bboxes, labels, scores, class_names, class_thresholds
     return bboxes, scores, labels
 
 
+def apply_class_thresholds_or_take_best(bboxes, labels, scores, class_names, class_thresholds, epsilon=0.05):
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/transforms/table_structure/table_transformers.py
```diff
@@ -53,6 +53,23 @@ def apply_class_thresholds(bboxes, labels, scores, class_names, class_thresholds
     return bboxes, scores, labels
 
 
+def apply_class_thresholds_or_take_best(bboxes, labels, scores, class_names, class_thresholds, epsilon=0.05):
+    """
+    Filter out bounding boxes whose confidence is belor the confidence threshold for its
```
bsowell, line 58: typo: belor -> below

bsowell (approved):
Hard to eyeball this and say whether these changes will improve things, but I don't see any glaring issues, so lgtm.

PR #1079, opened 2024-12-18
Add map reduce style summarize to handle large texts for summarization.

austintlee:


bsowell:
I would love to see if we can integrate summarization into some of the distributed map-reduce stuff that @bohou-aryn has been working on. I get that this is slightly different, but that's what I think about when I think of Map Reduce.

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -31,6 +33,112 @@
...
+        i += additional
+        remaining = tokens_per_chunk - len(cur_summary)
+        additional = min(remaining, total - i)
+        if additional == 0:
+            break
+
+    return cur_summary
+
+
+class DocumentSummarizer(Summarizer):
```
baitsguy, line 94: would it make sense for this to live in docset?

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -31,6 +33,112 @@
...
+    def __init__(
+        self,
+        llm: LLM,
+        question: str,
+        chunk_size: int = 10 * 1000,
```
baitsguy, line 99: think you're going to want to pass in a tokenizer/max_tokens like we do for llm_filter etc

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -31,6 +33,112 @@
...
+            text, self.chunk_size, CharacterTokenizer(), QuestionAnsweringSummarizer(self.llm, self.question)
+        )
+        document.properties["summary"] = summary
+        return document
+
+    def get_text(self, doc: Document) -> str:
+        doc_text = ""
+        props_dict = doc.properties.get("entity", {})
+        props_dict.update({p: doc.properties[p] for p in set(doc.properties) - set(BASE_PROPS)})
+        # doc_text = f"Document {di}:\n"
```
baitsguy, line 123: leftover

baitsguy (commented):
Overall approach looks useful and an improvement. Can you move some of the logic into docset, and also incorporate token counting with a tokenizer instead of chars

lib/sycamore/sycamore/query/execution/operations.py
austintlee: I kept this because I wanted to keep the logic consistent with the current summarizer.  Do we not want these properties here?

lib/sycamore/sycamore/query/execution/operations.py
austintlee: Ok

lib/sycamore/sycamore/query/execution/operations.py
austintlee: I think so.  I will move it there.

baitsguy (approved):

PR #1080, opened 2024-12-18
Add kmeans clustering based on ray

bohou-aryn:
This includes generally three steps:
1. materialize a document's embedding
2. initialize centroids randomly
2. iterate the kmeans process until converge, this is based on ray dataset map group and aggregate operators.

The result centroids could be used for downstream work.

### DIFF
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -15,6 +15,7 @@ from sycamore.llms.prompts.default_prompts import (
 )
 from sycamore.plan_nodes import Node, Transform
 from sycamore.transforms.augment_text import TextAugmentor
+from sycamore.transforms.clustering import KMeans
 from sycamore.transforms.embed import Embedder
 from sycamore.transforms import DocumentStructure, Sort
 from sycamore.transforms.extract_entity import EntityExtractor, OpenAIEntityExtractor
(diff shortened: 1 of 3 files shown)


austintlee:
Don't we want the result of `.kmeans()` to be a DocSet?  I'm not sure what to do with the resulting array of vectors.

In the summarization case, we could use kmeans/clustering for topic discovery and select a few elements from each cluster and use those to summarize.  What I would need is membership info.

austintlee:
Technically, this is K means and I am talking about clustering.

bohou-aryn:
> Technically, this is K means and I am talking about clustering.

The github comment seems not work, your comment could not be replied in dialogue. Anyway, the PR already shows how the clustering method works. Basically, you do a similar trick to assign each row a cluster id using a map function by comparing with the centroids, and then do either sort or aggregate for your purpose. In your summarization case, you might chain an aggregate function to select N entries out for each group.

lib/sycamore/sycamore/docset.py
```diff
@@ -903,6 +906,28 @@ def map(self, f: Callable[[Document], Document], **resource_args) -> "DocSet":
         mapping = Map(self.plan, f=f, **resource_args)
         return DocSet(self.context, mapping)
 
+    def kmeans(self, K: int, iterations: int, init_mode: str = "random", epsilon: float = 1e-4):
```
bsowell, line 909: I get that it's different, but I do think we should have a method that returns a Docset with an extra field indicating that cluster that each row is assigned to. I know that it is just an extra map, but I expect it's how people would want to use it in practice and it makes things more chainable.

lib/sycamore/sycamore/docset.py
```diff
@@ -903,6 +906,28 @@ def map(self, f: Callable[[Document], Document], **resource_args) -> "DocSet":
...
+        """
+        Apply kmeans over embedding field
+
+        Args:
+            K: the count of centroids
+            iterations: the max iteration runs before converge
```
bsowell, line 915: Is there a reasonable default for this? I at least wouldn't know what a good value to pick would be.

bohou-aryn: spark uses 20, we could follow the same, but it should really be a tuning process.

lib/sycamore/sycamore/docset.py
bohou-aryn: yes, sure, would add one.

bohou-aryn:
> How hard would it be to add a local mode version of this? I guess it would require a local-mode aggregate. That would be convenient, though I confess I'm not sure how important it is.

would try to come up with one.

PR #1081, opened 2024-12-19
Add-tokenizer-and-reranking-to-LLM-ExtractEntity

dhruvkaliraman7:
1. Added reranking of elements and element merging/chunking for robustness in LLM Extract Entity. (Works only if LLM returns "None" if it doesn't find the entity in the chunk, which is the default in Luna operator prompt)
2. Refactored code to make use of element similarity sorter, mock tokenizer.
3. Minor bug fix: changed element_index to _element_index for adding to content for LLM Filter.

### DIFF
--- a/lib/sycamore/sycamore/transforms/extract_entity.py
+++ b/lib/sycamore/sycamore/transforms/extract_entity.py
@@ -14,6 +14,7 @@ from sycamore.utils.time_trace import timetrace
 from sycamore.functions.tokenizer import Tokenizer
 from sycamore.transforms.similarity import SimilarityScorer
 from sycamore.utils.similarity import make_element_sorter_fn
+from sycamore.utils.llm_utils import merge_elements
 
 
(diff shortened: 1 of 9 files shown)


lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -126,6 +144,42 @@ def _handle_element_prompting(self, document: Document) -> Any:
         else:
             return self._get_entities(content)
 
+    def _handle_element_chunking(self, document: Document) -> Any:
+        assert self._tokenizer is not None
+        ind = 0
+        while ind < len(document.elements):
```
baitsguy, line 150: can you combine this scrolling logic with what's in llm filter

dhruvkaliraman7: done

baitsguy (approved):

PR #1083, opened 2024-12-20
Schema object + entity extraction support

baitsguy:
(This is a sequence of PRs)

Introduces a Schema class in sycamore. Allows you to extract properties from that schema in the `extract_properties` transform. Test validates some basic cleanup and formatting in the extracted info

Next big step will be to refactor sycamore.query to work against this object.


bsowell (approved):
This looks reasonable to me. Going to go ahead and get it in so we can integrate with DocStore.

PR #1084, opened 2024-12-21
Add Embed at Element Level

dhruvkaliraman7:
1. Allow .embed() transform to operate on element level (without exploding). 
2. Refactor code for Embedder class to handle orchestration of batching.

### DIFF
--- a/lib/sycamore/sycamore/transforms/embed.py
+++ b/lib/sycamore/sycamore/transforms/embed.py
@@ -3,7 +3,6 @@ import logging
 from abc import ABC, abstractmethod
 from enum import Enum
 from typing import Any, Optional, Callable, Union, List
-import logging
 
 from openai import OpenAI as OpenAIClient
 from openai import AzureOpenAI as AzureOpenAIClient
--- a/lib/sycamore/sycamore/transforms/embed.py
+++ b/lib/sycamore/sycamore/transforms/embed.py
@@ -2,12 +2,13 @@ import json
 import logging
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import Any, Optional, Callable, Union
+from typing import Any, Optional, Callable, Union, List
+from collections import defaultdict
 
 from openai import OpenAI as OpenAIClient
 from openai import AzureOpenAI as AzureOpenAIClient
 
-from sycamore.data import Document
+from sycamore.data import Document, Element
 from sycamore.llms import OpenAIClientParameters
 from sycamore.utils import choose_device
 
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/transforms/embed.py
```diff
@@ -19,44 +20,117 @@
...
     def generate_embeddings(self, doc_batch: list[Document]) -> list[Document]:
-        pass
+        """Handle batching and document processing logic in parent class"""
+
+        # Collect documents and elements to embed
+        text_to_embed = []
+        element_indices = defaultdict(set)
+        doc_indices = set()
+
+        # First pass: collect all texts that need embedding
```
eric-anderson, line 90: I think the following works, and it means that you don't have to worry about the algorithm for adding text to embed matches the algorithm for assigning the embeddings, e.g. you could swap the two if branches in the first loop and it would work fine.
```
        obj_for_embedding = []
        for doc in doc_batch:
            if not _text_representation_is_empty(doc):
                text_to_embed.append(self.pre_process_document(doc))
                obj_for_embedding.append(doc)

            if isinstance(doc, Document) and doc.get("elements"):
                for element in doc.elements:
                    if not _text_representation_is_empty(element):
                        text_to_embed.append(self.pre_process_document(element))
                        obj_for_embedding.append(elemnt)

         # Return early if nothing to embed
        if not text_to_embed:
            return doc_batch

        # Generate embeddings
        all_embeddings = []
        for text_batch in batched(text_to_embed, self.model_batch_size):
            batch_embeddings = self.embed_texts(text_batch)
            all_embeddings.extend(batch_embeddings)

        for i, embedding in enumerate(all_embeddings):
            obj_for_embedding[i].embedding = embedding
```

lib/sycamore/sycamore/transforms/embed.py
```diff
@@ -252,49 +291,43 @@ def __init__(
...
-
-        return self._generate_embedding(client, text)
+    def _ensure_client(self):
+        if self._client is None:
+            import boto3
+
+            boto3.session.Session(*self.boto_session_args, **self.boto_session_kwargs)
+            self._client = boto3.client("bedrock-runtime")
+
+    def embed_texts(self, texts: List[str]) -> List[List[float]]:
```
eric-anderson, line 317: assert len(texts) == 1; then can remove the loop.

lib/sycamore/sycamore/transforms/embed.py
```diff
@@ -19,44 +20,117 @@
...
-        pass
+        """Handle batching and document processing logic in parent class"""
+
+        # Collect documents and elements to embed
+        text_to_embed = []
+        element_indices = defaultdict(set)
+        doc_indices = set()
+
+        # First pass: collect all texts that need embedding
+        for i, doc in enumerate(doc_batch):
```
eric-anderson, line 91: add tests for embedding just elements, and both text representation and elements.

lib/sycamore/sycamore/transforms/embed.py
```diff
@@ -19,44 +20,117 @@
...
+        client: Optional[OpenAIClient] = None,
     ):
         self.model_name = model_name
         self.batch_size = batch_size
-        self.model_batch_size = model_batch_size
         self.pre_process_document = pre_process_document if pre_process_document else _pre_process_document
-
         self.device = choose_device(device)
 
+        # Set default model batch sizes based on embedder type
```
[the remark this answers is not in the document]
eric-anderson: Also add tests for batch_size setting.

eric-anderson (approved):

PR #1087, opened 2024-12-27
Changed the default embedding model to openai.

akarshgupta7:
Latest version of ragas has a couple of abstract methods (aembed_query, aembed_documents) that are not defined when using the Huggingface embedding models which throws an error. So switched to the new models released by openai.


apps/query-eval/queryeval/driver.py
```diff
@@ -195,7 +198,8 @@ def __init__(
         self.rouge_scorer = RougeScore()
         self.semantic_similarity_scorer = SemanticSimilarity()
         self.semantic_similarity_scorer.embeddings = LangchainEmbeddingsWrapper(
-            HuggingfaceEmbeddings(model_name="sentence-transformers/all-MiniLM-L6-v2")
+            # HuggingfaceEmbeddings(model_name="sentence-transformers/all-MiniLM-L6-v2")
```
baitsguy, line 201: you can delete this line/import before merging

baitsguy (approved):

apps/query-eval/queryeval/driver.py
baitsguy: Also pls put some description/reference to the bug in the description :)

PR #1088, opened 2024-12-30
Get sycamore.query to work with Schema instead of only OpenSearchSchema

baitsguy:
Other places are currently passing OpenSearchSchema to the planner (e.g. query-eval, testing scripts) so I need to support both OpenSearchSchema and Schema in the planner right now.

Also had to add an explicit `OpenSearchSchema.to_schema()` method because both objects define a 'fields' attribute of different types :(

### DIFF
--- a/lib/sycamore/sycamore/query/planner.py
+++ b/lib/sycamore/sycamore/query/planner.py
@@ -1,7 +1,7 @@
 import logging
 import typing
 from dataclasses import dataclass
-from typing import Any, List, Optional, Tuple, Type
+from typing import Any, List, Optional, Tuple, Type, Union
 
 from sycamore.schema import Schema, SchemaField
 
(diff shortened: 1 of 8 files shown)


lib/sycamore/sycamore/query/planner.py
```diff
@@ -379,7 +388,7 @@ class LlmPlanner:
     def __init__(
         self,
         index: str,
-        data_schema: OpenSearchSchema,
+        data_schema: typing.Union[OpenSearchSchema, Schema],
```
bsowell, line 391: not a big deal at all, but why not just import Union -- looks like you already import a bunch of things from typing.

bsowell (approved):
I don't know the sycamore.query codebase well enough to guarantee this covers everything, but looks okay to me. Once this is in I will remove the last vestiges of OpenSearchSchema from DocStore.

lib/sycamore/sycamore/query/planner.py
baitsguy: oops IDE didn't want to. Will fix

PR #1089, opened 2024-12-31
Add hybrid table extractor

HenryL27:
Adds a 'hybrid' table extractor that wraps DeformableTableExtractor and TableTransformerExtractor by selecting one to use based on the size of the table element.

### DIFF
--- a/lib/sycamore/sycamore/transforms/table_structure/extract.py
+++ b/lib/sycamore/sycamore/transforms/table_structure/extract.py
@@ -1,5 +1,5 @@
 from abc import abstractmethod
-from typing import Any
+from typing import Any, Union
 
 from PIL import Image
 import pdf2image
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -231,6 +231,56 @@ def extract(
...
+            return self._tatr
+        width, height = doc_image.size
+        bb = element.bbox.to_absolute(width, height)
+        padding = 10
+        max_dim = max(bb.width, bb.height) + 2 * padding
+        if max_dim > 500:
+            return self._deformable
+        return self._tatr
+
+    def _init_structure_model(self):
```
dhruvkaliraman7, line 262: This is not going to be triggered correct?

HenryL27: The model server uses it to load models into the actors before any requests come through

dhruvkaliraman7 (approved):
lgtm!

PR #1094, opened 2025-01-03
Add materialize read reliability

dhruvkaliraman7:
### 1. MaterializeReadReliability: A new class that enables reliable batch processing of materialized files by:

- Tracking already processed files 

- Limiting batch sizes 

- Supporting incremental processing through batch resets

- Maintaining state between batch executions

### 2. Added utility functions:

- name_from_docid: Custom naming function using path-based SHA256 hashes

- docid_from_path: Generates document IDs(SHA 256) from path

- doc_only_to_binary: Serialization helper


lib/sycamore/sycamore/docset.py
```diff
@@ -1523,6 +1528,15 @@ def execute(self, **kwargs) -> None:
         """
 
         from sycamore.executor import Execution
+        from sycamore.materialize import Materialize
 
-        for doc in Execution(self.context).execute_iter(self.plan, **kwargs):
-            pass
+        # TO:DO assumes last node is materialize, need to add functionality if multiple materialize steps present
+        if isinstance(self.plan, Materialize) and self.plan._reliability is not None:
```
dhruvkaliraman7, line 1534: 1. Do we anticipate having multiple mrr objects (multiple read reliability materializes) in one pipeline?
2. I could also take in mrr in execute and not pass it through Materialize class.

eric-anderson: Per discussion, we will explicitly disallow multiple mrr objects.

lib/sycamore/sycamore/docset.py
```diff
@@ -1523,6 +1529,31 @@ def execute(self, **kwargs) -> None:
...
+                        pass
+                    if mrr.current_batch == 0:
+                        logger.info(f"\nProcessed {mrr.prev_seen} docs.")
+                        break
+                except Exception as e:
+                    logger.info(f"Retrying batch job because of {e}.\n Processed {len(mrr.seen)} docs at present.")
+                    cycle_error = traceback.format_exc()
+                    print(f"Detailed Trace:\n{cycle_error}")
+                mrr.reset_batch()
+                if mrr.retries_count > 20:
```
HenryL27, line 1552: does it make sense to make the 20 a field in the mrr object so we can change it?

lib/sycamore/sycamore/materialize.py
```diff
@@ -43,6 +43,99 @@ def safe_cleanup(self, path) -> None:
...
+        else:
+            self.retries_count = 0
+
+    @staticmethod
+    def _path_to_id(p: Path) -> Optional[str]:
+        if p.suffix != ".pickle":
+            return None
+        if not p.name.startswith("doc-path-sha256-"):
+            return None
+        return str(p.stem[4:])
```
HenryL27, line 82: Does this work with https://github.com/aryn-ai/sycamore/commit/66a9e89a5b0ef078f0faace0e004350d35603434 and all the other stuff in docid.py? Also might move this to docid.py

lib/sycamore/sycamore/materialize.py
```diff
@@ -43,6 +43,99 @@ def safe_cleanup(self, path) -> None:
...
+
+        (fs, path) = infer_fs(str(out_mat_path))
+        logger.info(f"Fetching files from {out_mat_path}")
+
+        self.fs = fs
+        self.path = path
+        self.max_batch = max_batch
+        self.current_batch = 0
+        self.retries_count = 0
+        self.prev_seen = 0
```
HenryL27, line 58: should this get set after _refresh_seen_files() to len(self.seen)?

lib/sycamore/sycamore/materialize.py
```diff
@@ -54,12 +147,16 @@ def __init__(
         context: Context,
         path: Optional[Union[Path, str, dict]] = None,
         source_mode: MaterializeSourceMode = MaterializeSourceMode.RECOMPUTE,
+        tolerate_input_errors=False,
```
HenryL27, line 150: type annotation?

lib/sycamore/sycamore/tests/unit/test_materialize.py
```diff
@@ -528,3 +537,206 @@ def test_s3_infer_filesystem():
...
+
+    def test_materialize_read_reliability_no_source_doc_ids(self):
+        ctx = sycamore.init(exec_mode=self.exec_mode)
+        # Value error, no pickle files present, since source docids were not written using docid_from_path
+
+        with tempfile.TemporaryDirectory() as tmpdir1:
+            docs = make_docs(3)
+            ds = ctx.read.document(docs).materialize(
+                path={"root": tmpdir1, "tobin": doc_only_to_binary}, source_mode=sycamore.MATERIALIZE_RECOMPUTE
+            )
```
HenryL27, line 621: does it make sense to have this do mrr too?

lib/sycamore/sycamore/tests/unit/test_materialize.py
```diff
@@ -528,3 +537,206 @@ def test_s3_infer_filesystem():
...
+        def mock_mrr_reset():
+            counter.x += 1
+            original_mrr_reset()
+
+        mrr.reset_batch = mock_mrr_reset
+        return mrr
+
+    def test_materialize_read_reliability(self):
+        ctx = sycamore.init(exec_mode=self.exec_mode)
+        with tempfile.TemporaryDirectory() as tmpdir1:
```
HenryL27, line 560: fyi you can do multiple withs like
```python
with TempDir() as td1, TempDir() as td2, TempDir() as td3:
    blah
```

lib/sycamore/sycamore/tests/unit/test_materialize.py
```diff
@@ -528,3 +537,206 @@ def test_s3_infer_filesystem():
...
+                failure_counter = NumCalls()
+
+                mrr = MaterializeReadReliability(tmpdir2, max_batch=3)
+
+                mrr = self.mock_mrr_reset_fn(mrr, retry_counter)
+
+                # Create a function that fails for specific documents
+                def failing_map(doc):
+                    failure_counter.x += 1
+                    if failure_counter.x >= 9:  # Fail batch with every 4th document
```
HenryL27, line 719: comment

lib/sycamore/sycamore/plan_nodes.py
```diff
@@ -138,6 +138,20 @@ def _traverse(self, obj: NodeTraverse) -> "Node":
...
+        if not self.children or all(c is None for c in self.children):
+            return isinstance(self, Materialize) and self._reliability is None
+        assert not isinstance(
+            self, Materialize
+        ), "For ensuring reliability, only first and last node must be materialize"
+
+        for child in self.children:
+            if child is not None:
+                return child.last_node_materialize()
+        return False
```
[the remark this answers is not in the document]
dhruvkaliraman7: If my pipeline begins at a materialize with no reliability, return True.
Just kept this for uniformity, separate assertion inside materialize for checking if filter function is passed

lib/sycamore/sycamore/docset.py
dhruvkaliraman7: ack, done

lib/sycamore/sycamore/tests/unit/test_materialize.py
dhruvkaliraman7: This is the read stage, would add separate tests for file read reliability, unless you mean something else?

lib/sycamore/sycamore/materialize.py
dhruvkaliraman7: hmm, do not have : in filename so shouldn't affect?

lib/sycamore/sycamore/plan_nodes.py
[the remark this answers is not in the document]
eric-anderson: I think this function should be in materialize since it's specific to that logic.
I also don't think it's checking the correct condition. At this point (we may improve in the future), we want the condition to be that:
  1) reliability can only be set on the last node in the plan, and 
  2) if reliability is enabled, then
      a) the plan is linear (each node has 0 or 1 children), 
      b) the first node is a materialize, 
      c) there is no materialize in between, and 
      d) neither materialize overrides the name,  tobin or filter options.
(2.c isn't completely required, we could allow a materialize that isn't a source, but I see little benefit and 2.c is a simpler rule).

I would explicitly check that.  Then we can simplify the usability. right now the tests are putting filter=mrr.filter in the first mat and setting stuff in the last one.
I think we can simplify it just down to:
(
   context.init()
   .read.materialize("in path/s3spec")
   .stuff()
   ...
    .materialize("out path/s3spec", reliability=MaterializeReadReliability([options like batch size]))
    .execute()
)
This design would require that when the execution happens that the in-path needed by MRR is injected from the first materialize, the filter is injected into the first materialize, and the various name setting functions and checks are updated.

Late thought, this looks a lot like a node traverse.
I wonder if we would be better writing this as:
ctx = context.init()
ctx.rewrite_rules.append(MaterializeReadReliability([options like batch size])
-- then the ctx can be reused, the rest of it behaves like I wrote above except MRR becomes a NodeTraverse.

You mentioned something about putting MRR in the context, was this what you meant?

lib/sycamore/sycamore/docset.py
```diff
@@ -1523,6 +1529,31 @@ def execute(self, **kwargs) -> None:
...
+        from sycamore.materialize import Materialize
 
-        for doc in Execution(self.context).execute_iter(self.plan, **kwargs):
-            pass
+        if isinstance(self.plan, Materialize) and self.plan._reliability is not None:
+            mrr = self.plan._reliability
+
+            assert self.plan.child().last_node_materialize(), "first step must be materialize step"
+
+            while True:
```
eric-anderson, line 1539: Minor, but need a hook here to call before each iteration, ideally with an iteration count and the previous round error so that we can clear the jupyter output between rounds, and print out status information (count & prev error).

Without that, too much output makes jupyter wedge up.

lib/sycamore/sycamore/plan_nodes.py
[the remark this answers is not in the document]
dhruvkaliraman7: As per discussion, converted to node traverse

lib/sycamore/sycamore/docset.py
```diff
@@ -1531,29 +1530,45 @@ def execute(self, **kwargs) -> None:
         from sycamore.executor import Execution
         from sycamore.materialize import Materialize
 
-        if isinstance(self.plan, Materialize) and self.plan._reliability is not None:
-            mrr = self.plan._reliability
-
-            assert self.plan.child().last_node_materialize(), "first step must be materialize step"
+        if isinstance(self.plan, Materialize):
```
eric-anderson, line 1533: I can't add the comment where I should, but in the execute docstring include basic instructions for MRR with a pointer to the full documentation in MaterializeReadReliability class.  Something like:
"""
Pipelines can be reliably executed by doing:
ctx = sycamore.init()
ctx.rewrite_rules.append(MaterializeReadReliabliity([batch_size=200]))
ds = ctx.read.<something>.<operation>...
ds.execute()

for more details see the MaterializeReadReliability class.
"""

It would be nice at some point to support reliability with take, take_all and take_streaming, but that's not urgent since nothing needs it.

lib/sycamore/sycamore/docset.py
```diff
@@ -1531,29 +1530,45 @@ def execute(self, **kwargs) -> None:
...
+            mrr = None
```
eric-anderson, line 1534: Can you refactor this to have the reliable execution logic in the same file as MRR?  Then the code can be:
```
if maybe_execute_reliably(self):
   pass # execution already ran
else:
    for doc in Execution(self.context).execute_iter(self.plan, **kwargs):
          pass

with the expected return True/False in maybe_execute_reliably
```

This function is borderline too long and is mixing reliable and unreliable execution.
I think when you refactor, you can also split the function. The first half is figuring out if running into reliability mode so can end with:
```
if mrr is None:
   return False
else
   execute_reliably(docset, mrr, destinationPath)
```

lib/sycamore/sycamore/docset.py
```diff
@@ -1531,29 +1530,45 @@ def execute(self, **kwargs) -> None:
...
+            for rule in self.context.rewrite_rules:
+                if isinstance(rule, MaterializeReadReliability):
+                    mrr = rule
+                    if isinstance(self.plan._orig_path, str):
```
eric-anderson, line 1538: Check to make sure this works with S3 & non S3 paths. I remember some weirdness there.
Single manual test is sufficient.

eric-anderson (approved):

HenryL27 (approved):

eric-anderson (approved):

PR #1095, opened 2025-01-03
Property extraction type cast

baitsguy:
Support some basic type casting. I think we'll want "typed types" eventually and also some column standardization plugins, but I don't want to block on API design

### DIFF
--- a/lib/sycamore/sycamore/transforms/extract_schema.py
+++ b/lib/sycamore/sycamore/transforms/extract_schema.py
@@ -207,7 +207,10 @@ class LLMPropertyExtractor(PropertyExtractor):
 
         for field in self._schema.fields:
             value = fields.get(field.name)
-            result[field.name] = type_cast_functions.get(field.field_type, lambda x: x)(value)
+            if value is None and field.default is None:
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/transforms/extract_schema.py
```diff
@@ -15,6 +15,8 @@
 from sycamore.utils.extract_json import extract_json
 from sycamore.utils.time_trace import timetrace
 
+import dateparser
```
bsowell, line 18: Is this a new dependency, or something we were already using? I was under the impression that we had some logic for certain date formats in the standardizer, as well. Do we want to make sure we are using the same logic here?

bsowell (approved):
One question on the date handling, but otherwise lgtm.

lib/sycamore/sycamore/transforms/extract_schema.py
baitsguy: The standardizer uses it as well, the interface for those classes is a little awkward to use here. I'm hoping we can eventually port that stuff in here/phase it out

PR #1100, opened 2025-01-07
Add support for parallel read in OpenSearchReader

austintlee:
### Background
Today, the base reader implementation for Sycamore connectors for various external data sources fetches full documents into a MaterializedDataset in a single thread.  All downstream transforms wait for this to complete.

### Parallelizing document read - using Point-in-Time with Search Slices
To speed up the read for OpenSearchReader, we make use of the PIT feature along with search slices.  Search results from different slices can be fetched in parallel and OpenSearch recommends for deep pagination, to use PIT to achieve consistent results.

Phase 1 - create a PIT and compute # slices to use.
Phase 2 - each slice is flat mapped to the set of results to fetched from that slice.  Here, we only need doc IDs.
Phase 3 - We map each doc ID to a full document payload.  For reconstruct, we map parent_id to all child docs with the same parent_id.

### Custom partitioning
OpenSearch allows clients to fetch all documents in a slice via pagination.  The "size" + "from" pagination works if the total count for query results <= 10,000.  However, in order to fetch more than 10k results, we need to use "search_after" for deep pagination.  This requires a sort field.  OpenSearch does not allow a text field to be used for sorting.  So, to be able to use "search_after" for pagination, we need documents to be indexed with some numeric sort field (Elastic provides `_shard_doc` for this purpose, but OpenSearch does not).  So, we also implement custom partitioning for cases where we can't use PIT.  To do this, we need to first fetch all doc IDs single-threaded (with PIT + slicing, this can be done in parallel) and split them into N partitions, group by partition and then fetch document payloads.


austintlee:
For local testing, I added four new integration tests and tested with an index with > 20k documents.

lib/sycamore/pyproject.toml
```diff
@@ -85,6 +85,7 @@ transformers = { version = "^4.43.1", optional = true }
 unstructured = { version = "0.10.20", optional = true }
 python-pptx = {version = "^0.6.22", optional = true }
 nanoid = "^2.0.0"
+nltk = "3.8.1"
```
austintlee, line 88: For some reason, upgrading Unstructed did not make this issue (https://github.com/Unstructured-IO/unstructured/issues/3511) go away.  I did this to unblock local development for now.

austintlee:
I'll address mypy and poetry.lock issues after I get some feedback on the PR.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +224,305 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
...
+        # logger.info(f"Applying on {group} ({type(group)}) ...")
+        parent_ids = set()
+        for row in group["parent_id"]:
+            # logging.info(f"Row: {row}: {type(row)}")
+            parent_ids.add(row)
+
+        logger.info(f"Parent IDs: {parent_ids}")
+        return pd.DataFrame([{"_source": {"doc_id": parent_id, "parent_id": parent_id}} for parent_id in parent_ids])
+
+    def reconstruct(self, doc: dict[str, Any]) -> Dict[str, Any]:
```
bsowell, line 382: I'm sure it's covered and I'm just not quite grokking it, but can you explain what is guaranteeing that a parent doc and all of it's chunks are in the same slice, assuming that that is what is happening?

bsowell (commented):
First pass.

lib/sycamore/pyproject.toml
austintlee: I use it as a "dummy" partitioner in tests where I don't care about output quality, though.  Is there a lightweight alternative with minimal dependencies?

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +224,305 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
         return all_elements
 
 
+def get_doc_count(os_client, index_name: str, query: Optional[Dict[str, Any]] = None) -> int:
+    res = os_client.search(index=index_name, body=query, size=0, track_total_hits=True)
```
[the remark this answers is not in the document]
austintlee: Yes.

https://opensearch.org/docs/latest/api-reference/count/

I think I chose this because this can be used for slices as well.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +224,305 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
         return all_elements
 
 
+def get_doc_count(os_client, index_name: str, query: Optional[Dict[str, Any]] = None) -> int:
```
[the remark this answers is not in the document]
austintlee: Ah, I didn't know that!

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +224,305 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
...
+            size = 1000
+            page = 0
+            logger.info(f"Executing {slice_query} against {self._query_params.index_name}")
+
+            query_params = {"_source_includes": ["doc_id", "parent_id", "properties"]}
+            while True:
+                res = os_client.search(
+                    body=slice_query,
+                    size=size,
+                    from_=page * size,
```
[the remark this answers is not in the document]
austintlee: I mentioned in this in the overview above.  `search_after` requires a sort field.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +224,305 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
...
+        finally:
+            if client is not None:
+                client.close()
+
+        records = OpenSearchReaderQueryResponse(results, os_client)
+        docs = records.to_docs(query_params=self._query_params)
+        # logging.info(f"Sample: {docs[:5]}")
+        return [{"doc": doc.serialize()} for doc in docs]
+
+    def map_reduce_parent_id(self, group: pd.DataFrame) -> pd.DataFrame:
```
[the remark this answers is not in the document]
austintlee: I had the same question!

https://docs.ray.io/en/latest/data/api/doc/ray.data.Dataset.groupby.html

The Ray API only works with panda or arrow.  I had to do this weird conversion to get it to work with Ray.  I don't think we want to use DocSet here.  Would that be more efficient?  It took me a half day just debugging this (Ray expecting panda.DataFrame when using groupby -> map_groups), but maybe I still got it wrong?

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +224,305 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
...
+            query = self._query_params.query
+
+            os_client = client._client
+            doc_count = get_doc_count(os_client, index_name, query)
+            num_slices = doc_count // 5000 if doc_count > 5000 else 1
+            # num_workers = self.resource_args.get("parallelism", 2)  # 2 is the minimum number of slices.
+            # num_slices = num_workers
+            logger.info(f"Reading {doc_count} documents from {index_name} in {num_slices} slices")
+
+            res = os_client.create_pit(index=index_name, keep_alive="100m")
```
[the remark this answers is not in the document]
austintlee: Maybe if you have too many PITs created?  I think the current best practice for deep pagination is to use PIT + search_after.  If we have a field that we can reliably use as a sort field, that's what we should do (instead of PIT + size + from).

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
austintlee: I first perform groupby("parent_id") and then fetch full documents.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +224,309 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
...
+                # Step 3: Group by parent_id
+                # Step 4: Deduplicate parent documents by ID
+                # Step 5: Reconstruct parent documents using 'parent_id'
+
+                return (
+                    ds
+                    .flat_map(self._to_parent_doc, **self.resource_args)
+                    .groupby("parent_id")
+                    .map_groups(self.map_reduce_parent_id)
+                    .map(self.reconstruct)
```
austintlee, line 527: @bsowell The formatting of my comments in the code got messed up after I applied poetry black.  Hopefully, this is a bit easier to read and understand.

This is for document reconstruct.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +224,309 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
...
+                )
+            else:
+                # Step 1: Construct slices (pages)
+                # Step 2: For each page, get all documents
+                return ds.flat_map(self._to_doc, **self.resource_args)
```
austintlee, line 532: This is for just fetching documents that match the query (reconstruct = False).

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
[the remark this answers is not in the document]
bsowell: Hmm. I can't quite explain this, I confess. We use groupby on a Ray Dataset in other contexts (https://github.com/aryn-ai/sycamore/blob/e1f0f9ca4183a3de18baf399cc99af42baa6dab9/lib/sycamore/sycamore/query/execution/aggregation.py#L74), but maybe it has to do with the map_groups? Can you at least add a comment explaining the intention here as otherwise it's pretty hard to follow.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -217,8 +246,313 @@ def _get_all_elements_for_doc_ids(self, doc_ids: list[str], index: str) -> list[
...
+
+            if not client.check_target_presence(self._query_params):
+                raise ValueError("Target is not present\n" f"Parameters: {self._query_params}\n")
+
+            index_name = self._query_params.index_name
+            query = self._query_params.query
+
+            os_client = client._client
+            doc_count = get_doc_count(os_client, index_name, query)
+            page_size = 2500
```
bsowell, line 507: Can you add a brief comment about where this num comes from.

bsowell (approved):
Hard to mentally go through all the cases, but the tests look good, and otherwise looks good.

lib/sycamore/pyproject.toml
austintlee: > maybe add it to the "legacy-partitioners" extra?

I'll do this.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
austintlee: OK.  I was aiming for a slice size reasonably below 10k.

PR #1102, opened 2025-01-09
Add `PartitionError` to `aryn_sdk.partition`'s `__init__.py`

MarkLindblad:


HenryL27 (approved):
lgtm. Do we need to do a sdk release?

MarkLindblad:
> lgtm. Do we need to do a sdk release?

I don't think this update is important enough to warrant an aryn-sdk release.

PR #1103, opened 2025-01-09
Prompt update for property extraction

baitsguy:
An attempt to get better json-parseable responses from the LLM with prompt tweaking. In this case we were previously getting python style dict responses.

Current, non-parseable responses with the json lib:
```
...
"someFieldWithNoValue": None
...
```

as opposed to this, which is parseable:

```
...
"someFieldWithNoValue": "None"
...
```

Another approach is to get `extract_json` to better parse these responses with regex/replace style approaches but that felt jankier and I want to avoid that as long as I can

### DIFF
--- a/lib/sycamore/sycamore/llms/prompts/default_prompts.py
+++ b/lib/sycamore/sycamore/llms/prompts/default_prompts.py
@@ -176,8 +176,9 @@ class ExtractPropertiesFromSchemaPrompt(SimplePrompt):
         Document text:
         {text}
         
-        Return your answers as a json dictionary, do not return any extra information.
-        If you cannot find a value, use the provided default or None.
+        Don't return extra information.
+        If you cannot find a value for a requested property, use the provided default or the value 'None'.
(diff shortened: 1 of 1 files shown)


Soeb-aryn (approved):
LGTM, do you think it is worth adding an assert statement to ensure `fields` is of right data type ? 
 https://github.com/aryn-ai/sycamore/blob/52e1c0bad11221ff55862f16e73baa565b6c58ed/lib/sycamore/sycamore/transforms/extract_schema.py#L209

baitsguy:
> LGTM, do you think it is worth adding an assert statement to ensure `fields` is of right data type ?
> 
> https://github.com/aryn-ai/sycamore/blob/52e1c0bad11221ff55862f16e73baa565b6c58ed/lib/sycamore/sycamore/transforms/extract_schema.py#L209

The current transform actually seems to work with a dictionary type response, but also just string style responses. I don't think the latter is necessarily needed but let me think about it outside this PR's context

PR #1105, opened 2025-01-09
Add File Read Reliability

dhruvkaliraman7:
This PR introduces File Read Reliability (FRR), a new component that complements the Materialize Read Reliability (MRR) system. While MRR handles reliability between materialized directories, FRR focuses on the initial data ingestion phase.

**Features:**
Implements reliability checks for raw data reading operations.
Currently supports BinaryScan operation (file scan through the first materialization step).

**Future Work:**
Extend FRR support to additional scan methods, implementation will be handled in subsequent PRs.

This PR is dependent on [Add Materialize Read Reliability](https://github.com/aryn-ai/sycamore/pull/1094)


lib/sycamore/sycamore/tests/unit/test_reader.py
```diff
@@ -46,3 +56,193 @@ def test_manifest(self, tmp_path: Path):
...
+
+            docs = make_docs(10)
+            docs.pop()
+
+            for doc in docs:
+                path = Path(tmpdir) / f"{doc.doc_id}.{doc.properties.get('extension', 'pdf')}"
+                path.write_bytes(b"test content")
+            ctx = sycamore.init(exec_mode=self.exec_mode)
+
+            docset = ctx.read.binary(tmpdir, binary_format="pdf")
```
eric-anderson, line 78: Where's the filtering? I expected a filter here for half the docs (or something like that)

lib/sycamore/sycamore/tests/unit/test_reader.py
```diff
@@ -46,3 +56,193 @@ def test_manifest(self, tmp_path: Path):
...
+            docs.pop()
+            mrr = MaterializeReadReliability(tmpdir2, max_batch=3)
+            mrr = mock_mrr_reset_fn(mrr, counter)
+            paths = []
+            for doc in docs:
+                path = Path(tmpdir1) / f"{doc.doc_id}.{doc.properties.get('extension', 'pdf')}"
+                path.write_bytes(b"test content")
+                paths.append(str(path))
+            ds = (
+                ctx.read.binary(paths, binary_format="pdf", filter_paths=mrr.filter)
```
eric-anderson, line 103: I'd like the same thing from FRR that we can get from MRR where setting reliability only has to happen in the final materialize step.  That means that for FRR, the materialize step would need to inject the reliability into the first node, and check that it supports reliability.  Having the subclasses set self._supports_file_read_reliabliity = True would work for that.

I'm still mixed about having support in the context, but I'm starting to lean toward that since people do re-use contexts in scripts.

lib/sycamore/sycamore/tests/unit/test_reader.py
dhruvkaliraman7: oops, duplicated

lib/sycamore/sycamore/tests/unit/test_reader.py
dhruvkaliraman7: done

lib/sycamore/sycamore/docset.py
```diff
@@ -1611,16 +1611,31 @@ def execute(self, **kwargs) -> None:
...
+        ctx = sycamore.init()
+        ctx.rewrite_rules.append(MaterializeReadReliability(max_batch=200, max_retries=20))
+        ds = ctx.read.materialize()\
+             ... \
+             .materialize()
+        ds.execute()  # Runs with batching, retries, and progress tracking
+
+
+        # Reliable execution with binary read
+        # Make sure to map docid_from_path to set unique docids that would be used to track documents processing
```
eric-anderson, line 1625: The map step should happen automatically if MRR is enabled with a file read.

lib/sycamore/sycamore/materialize.py
```diff
@@ -206,12 +218,14 @@ def _path_to_id(p: Path) -> Optional[str]:
             return None
         return str(p.stem[4:])
 
-    def filter(self, p: str) -> bool:
+    def filter(self, p: str, read: Optional[bool] = False) -> bool:
```
eric-anderson, line 221: Optional[bool] = False doesn't make sense, I'd just do read: bool = False

I'd also make it read_binary since both cases are reads.

lib/sycamore/sycamore/tests/unit/test_reader.py
```diff
@@ -46,3 +55,204 @@ def test_manifest(self, tmp_path: Path):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.exec_mode = ExecMode.LOCAL
+
+    def test_read_with_wrong_nodes(self):
+        with tempfile.TemporaryDirectory() as tmpdir, tempfile.TemporaryDirectory() as tmpdir1:
+
+            docs = make_docs(10)
+            docs.pop()
```
eric-anderson, line 70: why not docs = make_docs(9)?
or remove the docs.pop()

eric-anderson (approved):

lib/sycamore/sycamore/tests/unit/test_reader.py
dhruvkaliraman7: make_docs includes a MetadataDocument, removing that

PR #1107, opened 2025-01-10
Make `lib/poetry-lock/poetry-lock-all.sh` failures more obvious

MarkLindblad:
Previously, first time users who hadn't read the script could see the output abruptly stop and assume the script had completed successfully. Now, it will always either output a SUCCESS or a FAILED near the end of the output.

### DIFF
--- a/lib/poetry-lock/poetry-lock-all.sh
+++ b/lib/poetry-lock/poetry-lock-all.sh
@@ -26,19 +26,19 @@ for i in ${tomls}; do
         (
             echo "--------------------- special casing in $i"
             cd $(dirname "$i")
-            poetry lock --no-update || fail
+            poetry lock --no-update || fail "broke on special case 'poetry lock --no-update' for $i"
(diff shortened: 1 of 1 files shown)


lib/poetry-lock/poetry-lock-all.sh
```diff
@@ -2,6 +2,7 @@
 
 root="$(pwd)"
 fail() {
+    echo FAILED
```
eric-anderson, line 5: I'd do this: `echo "FAILED: $@"`
To make it a little harder to miss.

or if you think people will be bad at reading:
```
echo FAILED FAILED FAILED FAILED
echo "$@"
echo FAILED FAILED FAILED FAILED
```

eric-anderson (approved):

lib/poetry-lock/poetry-lock-all.sh
MarkLindblad: Done.

PR #1112, opened 2025-01-14
Add a lock for ray init to avoid multiple inits

austintlee:


austintlee:
This is the error if you try to call ray.init the second time:

```
error_message = "Perhaps you called ray.init twice by accident?"
>       assert not worker.connected, error_message
E       AssertionError: Perhaps you called ray.init twice by accident?

../../miniconda3/envs/sycamore-ray/lib/python3.12/site-packages/ray/_private/worker.py:2237: AssertionError
```

PR #1114, opened 2025-01-16
Removing Guidance

Soeb-aryn:
This pull request removes the Guidance framework, which previously acted as a wrapper around OpenAI's LLM calls. The removal streamlines the codebase by eliminating unnecessary dependencies and simplifying the interaction with the OpenAI models. As a result, the system directly interfaces with the OpenAI API without the added abstraction layer provided by Guidance.


lib/sycamore/sycamore/llms/reference.py
HenryL27, line 1: empty file?

HenryL27 (commented):
Looks like a bunch of LLM integration tests are failing now

lib/sycamore/sycamore/llms/reference.py
Soeb-aryn: this shouldn't be here. I will delete it

HenryL27 (approved):
lgtm

PR #1115, opened 2025-01-16
add groupby

bohou-aryn:


lib/sycamore/sycamore/transforms/group.py
```diff
@@ -0,0 +1,31 @@
...
+from sycamore import DocSet
+from sycamore.data import Document
+
+
+class GroupedData:
+    def __init__(self, docset: DocSet, key):
+        self._docset = docset
+        self._key = key
+
+    def aggregate(self, f) -> DocSet:
```
baitsguy, line 12: Should we consider subclassing DocSet or using a different type since most of our transforms aren't going to work with the new document structures? I think materialize on this won't work? @eric-anderson ?

HenryL27, line 12: Can I get a type for `f`?

lib/sycamore/sycamore/transforms/group.py
```diff
@@ -0,0 +1,31 @@
...
+        dataset = self._docset.plan.execute().map(Document.from_row)
+        grouped = dataset.groupby(self._key)
+        aggregated = grouped.aggregate(f)
+        m = aggregated.take()
```
HenryL27, line 16: what happens to `m`?

lib/sycamore/sycamore/transforms/group.py
```diff
@@ -0,0 +1,31 @@
...
+
+        def to_doc(row: dict):
+            count = row.pop("count()")
+            doc = Document(row)
+            properties = doc.properties
+            properties["count"] = count
+            doc.properties = properties
+            return doc.to_row()
```
HenryL27, line 24: What's actually in a row at this point? These are ray ops, right? 
Also why not just `doc.properties['count'] = count`?

lib/sycamore/sycamore/transforms/group.py
```diff
@@ -0,0 +1,31 @@
+from ray.data._internal.aggregate import Count
+
+from sycamore import DocSet
+from sycamore.data import Document
+
+
+class GroupedData:
```
HenryL27, line 7: Does this really belong in the `transforms` module? Might be a top-level thing.

HenryL27 (commented):
Any plans for how to implement this in `ExecMode.LOCAL`?

lib/sycamore/sycamore/transforms/group.py
bohou-aryn: yes, would add one

lib/sycamore/sycamore/transforms/group.py
bohou-aryn: used for debug, would remove.

lib/sycamore/sycamore/transforms/group.py
bohou-aryn: yes, might move it out.

baitsguy (approved):
You can address Henry's comments but good to merge after

bohou-aryn:
merged in https://github.com/aryn-ai/sycamore/pull/1123

PR #1120, opened 2025-01-21
[llm unify 1/n] Add consolidated prompt classes

HenryL27:
Creates a SycamorePrompt interface for rendering documents/elements as llm calls and adds a few implementations.
- StaticPrompt: static with no context
- ElementPrompt: renders an element
- ElementListPrompt: renders a document with a (optional) formatted list of elements in it

These are not the only ones to be implemented but these are necessary to move a number of the existing transforms to use llm_map (this prompt interface)

### DIFF
--- a/lib/sycamore/sycamore/llms/prompts/prompts.py
+++ b/lib/sycamore/sycamore/llms/prompts/prompts.py
@@ -209,3 +209,12 @@ class ElementListPrompt(SycamorePrompt):
         if self.user is not None:
             result.messages.append(RenderedMessage(role="user", content=self.user.format(**format_args)))
         return result
+
+    def instead(self, **kwargs) -> "SycamorePrompt":
+        new = copy.deepcopy(self)
+        for k in kwargs:
+            if k in new.__dict__:
+                new.__dict__[k] = kwargs[k]
+            else:
+                new.kwargs[k] = kwargs[k]
+        return new
--- a/lib/sycamore/sycamore/llms/prompts/prompts.py
+++ b/lib/sycamore/sycamore/llms/prompts/prompts.py
@@ -3,7 +3,9 @@ from typing import Any, Union, Optional, Callable
 import copy
 
 import pydantic
+from PIL import Image
 from sycamore.data.document import Document, Element
+from sycamore.utils.pdf_utils import get_element_image
 
 
 @dataclass
@@ -12,11 +14,13 @@ class RenderedMessage:
 
     Args:
         role: the role of this message. Should be one of "user", "system", "assistant"
-        content: the content of this message.
+        content: the content of this message, either a python string or a PIL image.
+        images: optional list of images to include in this message.
     """
 
     role: str
     content: str
+    images: Optional[list[Image.Image]] = None
 
     def to_dict(self):
         return {"role": self.role, "content": self.content}
@@ -59,8 +63,8 @@ class SycamorePrompt:
         """
         raise NotImplementedError(f"render_document is not implemented for {self.__class__.__name__}")
 
-    def render_element(self, elt: Element) -> RenderedPrompt:
-        """Render this prompt, given this element as context.
+    def render_element(self, elt: Element, doc: Document) -> RenderedPrompt:
+        """Render this prompt, given this element and its parent document as context.
         Used in llm_map_elements
 
         Args:
@@ -86,10 +90,12 @@ class SycamorePrompt:
         """Create a new prompt with some fields changed.
 
         Args:
+
             **kwargs: any keyword arguments will get set as fields in the
                 resulting prompt
 
         Returns:
+
             A new SycamorePrompt with updated fields.
 
         Example:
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/llms/prompts/prompts.py
```diff
@@ -0,0 +1,350 @@
...
+        Args:
+            doc: The document to use as context for rendering this prompt
+
+        Returns:
+            A two-message RenderedPrompt containing ``self.system.format()`` and ``self.user.format()``
+            using the format keys as specified above.
+        """
+        format_args = self.kwargs
+        format_args["doc_text"] = doc.text_representation
+        format_args.update({"doc_property_" + k: v for k, v in doc.properties.items()})
```
baitsguy, line 204: anything special we want to do for nested properties? lots of stuff is in properties.entity.* so single level flattening might not do much

baitsguy: Also why are we flattening instead of accepting the dot notation? I think it's more intuitive to send a prompt like:

```python
prompt = ElementListPrompt(
                system = "Hello {name}. This is a prompt about {document.properties.path}"
                ...
```

lib/sycamore/sycamore/llms/prompts/prompts.py
```diff
@@ -0,0 +1,350 @@
...
+            #   {"role": "system", "content": "Hello David Rothschild. This is a prompt about data/mypdf.pdf"},
+            #   {"role": "user", "content": "What do you make of these tables?\\nTables:\\n
+            #               ELEMENT 0: <last table csv>\\nELEMENT 1: <second-last table csv>..."}
+            # ]
+    """
+
+    def __init__(
+        self,
+        *,
+        system: Optional[str] = None,
```
baitsguy, line 160: if we're supporting system and user prompts (i.e. the messages api), shouldn't we be supporting a list then?

baitsguy (commented):
Implementation looks fine, I'm curious about how the existing transforms/uses are going to map to this. I worry a bit about usability, do you think you can transition one of the existing docset transforms to this stuff in this PR to demonstrate?

lib/sycamore/sycamore/llms/prompts/prompts.py
HenryL27: python string.format is rather limited, can't do any of
```python 
"hello {document.properties.name}".format(document=doc, document.properties=props, document.properties.name='henry')
```

- it doesn't do function calls or dict lookups
- can't have dotted param names.

which maybe means we shouldn't be using string.format here.

I did at one point play around with doing something like
```python
s = "Hello {name}. this is a prompt about {document.properties.path}"
fstring = f"f\"{s}\""
formatted = eval(fstring, abunchofglobals)
```

but that seems dangerous bc you could maybe do something like

```python
s = "{os.system('rm -rf /')}"
```

lib/sycamore/sycamore/llms/prompts/prompts.py
```diff
@@ -0,0 +1,350 @@
...
+
+    Args:
+
+        system: The system prompt string. Use {} to reference names that should
+            be interpolated. Defaults to None
+        user: The user prompt string. Use {} to reference names that should be
+            interpolated. Defaults to None
+        element_select: Function to choose which set of elements to include in
+            the prompt. If None, defaults to the first ``num_elements`` elements.
+        element_order: Function to reorder the selected elements. Defaults to
```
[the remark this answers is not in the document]
HenryL27: you right

lib/sycamore/sycamore/llms/prompts/prompts.py
HenryL27: and I don't want to invent a formatting/templating system

lib/sycamore/sycamore/llms/prompts/prompts.py
HenryL27: Sure. I think we can limit to only one system prompt (I don't remember what the providers do but that seems sensible) though

HenryL27:
idk if it makes sense to translate an op until llm_map exists but to sketch it out extract_entity would look like something like 
```python
EntityExtractorFewShotGuidancePrompt = ElementListPrompt(
    system="You are a helpful entity extractor",
    user="""You are given a few text elements of a document. The {entity} of the document is in these few text elements. Here are
    some example groups of text elements where the {entity} has been identified.
    {examples}
    Using the context from the document and the provided examples, FIND, COPY, and RETURN the {entity}. Only return the {entity} as part
    of your answer. DO NOT REPHRASE OR MAKE UP AN ANSWER.
    {elements}
    """,
)
```
(instead of the class construction as in default_prompts)
then 
```python
def extract_entity(self, entity_name, examples):
    prompt = default_prompts.EntityExtractorFewShotGuidancePrompt.instead(
        entity=entity_name, examples=examples
    )
    return self.llm_map(prompt, output_field=entity, llm_and_stuff_that_would_also_be_args_to_extract_entity)
```

lib/sycamore/sycamore/llms/prompts/prompts.py
baitsguy: Yeah that makes sense

baitsguy (approved):
maybe rename `instead` to something else like `populate` or something

PR #1122, opened 2025-01-23
capturing metadata from LLMs 

Soeb-aryn:
Adding LLM request metrics, such as wall_latency, in_token, and out_token, to the docset metadata for each LLM call.


baitsguy (commented):
Think we're repeating the same code a bunch of times, can you consolidate.

Also please add tests

baitsguy (approved):
Looks good, can you add an assertion to an existing integ test too

PR #1126, opened 2025-01-24
[llm unify 2/n] Implement llm_map(_elements) and move extract_entity to it.

HenryL27:
This is hard to break up bc I'm changing APIs of things that already exist and have tests...
- `llm.generate` now takes a RenderedPrompt instead of a dict
- the old api exists as `llm.generate_old`, and it's marked as deprecated.
- most operations have been switched to call `llm.generate_old`
- implement `docset.llm_map` and `docset.llm_map_elements`
- implement a `EntityExtractor.as_llm_map` method that converts all of the functionality of `OpenAIEntityExtractor` into prompts and hooks for llm_map
- `docset.extract_entity()` now uses that to use LLMMap instead of ExtractEntity.

Also changes the prompt rendering apis to produce either a single RenderedPrompt os a sequence of them in order to support a mode of extract entity that uses a tokenizer to make token-limited chunks to try to extract from


HenryL27:
afaict the ITs that are failing are mostly due to `llm.generate_old()` not quite converting `prompt_kwargs` to a `RenderedPrompt` correctly. I could spend a lot of time debugging that but given my goal is to delete `generate_old` entirely I'm tempted to just leave the tests broken and fix them when I translate the transforms they're testing to LLMMap.

lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -66,7 +90,16 @@ class TextSummarizerGuidancePrompt(SimplePrompt):
     var_name = "summary"
 
 
-class SchemaZeroShotGuidancePrompt(SimplePrompt):
+TextSummarizerGuidancePrompt = ElementPrompt(
+    system="You are a helpful text summarizer.",
+    user="""Write a summary of the following. Use onlt the information provided.
```
bsowell, line 95: typo: onlt -> only

lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -291,14 +405,14 @@ def __init__(self, field: str, groups: list[str]):
...
 _deprecated_prompts: dict[str, Type[SimplePrompt]] = {
-    "ENTITY_EXTRACTOR_ZERO_SHOT_GUIDANCE_PROMPT": EntityExtractorZeroShotGuidancePrompt,
-    "ENTITY_EXTRACTOR_ZERO_SHOT_GUIDANCE_PROMPT_CHAT": EntityExtractorFewShotGuidancePrompt,
-    "ENTITY_EXTRACTOR_FEW_SHOT_GUIDANCE_PROMPT_CHAT": EntityExtractorFewShotGuidancePrompt,
-    "ENTITY_EXTRACTOR_FEW_SHOT_GUIDANCE_PROMPT": EntityExtractorFewShotGuidancePrompt,
-    "TEXT_SUMMARIZER_GUIDANCE_PROMPT": TextSummarizerGuidancePrompt,
-    "TEXT_SUMMARIZER_GUIDANCE_PROMPT_CHAT": TextSummarizerGuidancePrompt,
-    "SCHEMA_ZERO_SHOT_GUIDANCE_PROMPT": SchemaZeroShotGuidancePrompt,
-    "SCHEMA_ZERO_SHOT_GUIDANCE_PROMPT_CHAT": SchemaZeroShotGuidancePrompt,
+    "ENTITY_EXTRACTOR_ZERO_SHOT_GUIDANCE_PROMPT": _EntityExtractorZeroShotGuidancePrompt,
```
bsowell, line 408: We now have two levels of translation for backwards compatibility -- the map that was here and then the _stuff that you added. I think it's great to deploy with compatibility like this, but I do wonder when it will be time to break compatibility, at least with the first generation of prompts (the all caps ones). Thoughts? Would it simplify the rest of your code at all? Do you know of any places we are still using them?

lib/sycamore/sycamore/llms/llms.py
```diff
@@ -1,10 +1,23 @@
...
 from sycamore.utils.cache import Cache
 from sycamore.utils.thread_local import ThreadLocalAccess, ADD_METADATA_TO_OUTPUT
 from sycamore.data.metadata import add_metadata
+from sycamore.llms.prompts import RenderedPrompt, RenderedMessage, SimplePrompt
+
+from sycamore.utils.deprecate import deprecated
+
+
+class LLMMode(Enum):
+    UNKNOWN = 0
```
bsowell, line 17: Who passes in the mode? Is it the caller? I'm wondering why/when we would ever get UNKNOWN if if we should disallow it.

lib/sycamore/sycamore/llms/llms.py
```diff
@@ -1,10 +1,23 @@
...
+    SYNC = 1
+    ASYNC = 2
+    BATCH = 3
```
bsowell, line 20: Is this something done by the LLM, or are we batching by combining multiple things into a prompt before sending?

lib/sycamore/sycamore/llms/llms.py
```diff
@@ -28,17 +63,52 @@ def format_image(self, image: Image.Image) -> dict[str, Any]:
...
+            raise ValueError("Either 'prompt' or 'messages' must be specified in prompt_kwargs")
+        return await self.generate_async(prompt=rendered, llm_kwargs=llm_kwargs)
+
     def __str__(self):
         return f"{self.__class__.__name__}({self._model_name})"
 
-    def _llm_cache_key(self, prompt_kwargs: dict, llm_kwargs: Optional[dict] = None) -> str:
+    @staticmethod
+    def _pickleable_response_format(prompt: RenderedPrompt) -> Any:
+        if inspect.isclass(prompt.response_format) and issubclass(prompt.response_format, pydantic.BaseModel):
```
bsowell, line 96: Why do you need inspect here? Does issubclass blow up for strings?

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -313,36 +313,40 @@ def _convert_response_format(self, llm_kwargs: Optional[Dict]) -> Optional[Dict]
...
-        else:
-            raise ValueError("Either prompt or messages must be present in prompt_kwargs.")
+                role = m.role
+            if m.images is None:
+                content: Union[str, list] = m.content
+            else:
+                content = [{"type": "text", "text": m.content}]
+                assert isinstance(content, list)  # mypy!!!
+                for im in m.images:
+                    content.append({"type": "image_url", "image_url": base64_data_url(im)})
```
bsowell, line 346: Isn't this the same as what `format_image` does? Are we replacing that? Or should we use it here?

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -313,36 +313,40 @@ def _convert_response_format(self, llm_kwargs: Optional[Dict]) -> Optional[Dict]
...
+            messages_list.append({"role": role, "content": content})
+
+        kwargs.update({"messages": messages_list})
         return kwargs
 
     def _determine_using_beta(self, response_format: Any) -> bool:
```
bsowell, line 352: I know it's not your PR and we can defer it, but this seems weird and probably outdated, or else it's just a bad method name..

lib/sycamore/sycamore/transforms/base_llm.py
```diff
@@ -0,0 +1,187 @@
...
+
+        child: Child node in the sycamore execution graph
+        prompt: The SycamorePrompt to use to render each document.
+            Must implement the ``render_document`` method.
+        output_field: The name of the field in doc.properties in which
+            to store the llm output
+        llm: The llm to use for inference.
+        llm_mode: How to call the llm - sync/async/batch. All LLMs do not
+            necessarily implement all options.
+        postprocess_fn: function to call on documents after performing the
```
bsowell, line 51: What's the rationale for having a postprocess_fn in this call vs just calling a map afterwards? The multiple prompts thing seems fairly complex.

lib/sycamore/sycamore/transforms/base_llm.py
```diff
@@ -0,0 +1,187 @@
...
+    elif llm_mode == LLMMode.ASYNC:
+        raise NotImplementedError("Haven't done async yet")
+    elif llm_mode == LLMMode.BATCH:
+        raise NotImplementedError("Haven't done batch yet")
+    else:
+        raise NotImplementedError("Unknown LLM Mode")
+
+
+class LLMMap(MapBatch):
+    """The LLMMap transform renders each Document in a docset into
```
bsowell, line 37: Not as part of this PR, but I wonder if eventually it would make sense to have an option to attach the raw LLM request/response the document as well for debugging.

lib/sycamore/sycamore/transforms/base_llm.py
```diff
@@ -0,0 +1,187 @@
...
+        except NotImplementedError as e:
+            raise e
+        except Exception:
+            pass
+
+
+class LLMMapElements(MapBatch):
+    """The LLMMapElements transform renders each Element for each
+    Document in a docset into a prompt for an LLM, calls the LLM,
+    and attaches the output to the document.
```
bsowell, line 111: Is it right to say rather that it attaches the output to the element?

lib/sycamore/sycamore/transforms/base_llm.py
```diff
@@ -0,0 +1,187 @@
+from typing import Optional, Sequence, Callable, Union
+
+from sycamore.llms.llms import LLM, LLMMode
+from sycamore.llms.prompts.prompts import SycamorePrompt, RenderedPrompt
+from sycamore.plan_nodes import Node
+from sycamore.transforms.map import MapBatch
+from sycamore.data import Document, Element
+
+
+def _infer_prompts(
```
bsowell, line 10: I confess I'm having a hard time following all of the prompt sequence stuff. I understand the basic motivation for the token stuff, but I can't help but wonder if there is a cleaner way to do it.

lib/sycamore/sycamore/transforms/llm_filter.py
```diff
@@ -35,6 +35,8 @@ def tokenized_threshold_llm_filter(
     max_tokens: int,
     tokenizer: Tokenizer,
 ) -> bool:
+    print("=" * 80)
+    print(doc)
```
bsowell, line 39: remove the prints in this file.

lib/sycamore/sycamore/docset.py
```diff
@@ -465,6 +467,7 @@ def extract_document_structure(self, structure: DocumentStructure, **kwargs):
         document_structure = ExtractDocumentStructure(self.plan, structure=structure, **kwargs)
         return DocSet(self.context, document_structure)
 
+    @deprecated(version="0.1.31", reason="Use llm_map instead")
```
bsowell, line 470: Is the plan also to deprecate extract_properties?

bsowell (commented):
There's a lot here. For the most part it looks good, I think. Couple of things to think about:

- Is there a reasonable path to replace the generate_old calls at some point, do you think? I think you are doing the right thing doing the replacement incrementally, but I have a suspicion that we are going to live with this for a while. 

- Anything else we need to test to get confidence here? I guess most of our demos don't use sycamore directly anymore. Does AddDoc in DocStore still work? What about the DocPrep generated scripts?

bsowell:
Also, have you gone through the integ test failures to see if they are a related? I did a cursory glance and a few did look llm related.

lib/sycamore/sycamore/llms/prompts/prompts.py
```diff
@@ -113,6 +113,19 @@ def set(self, **kwargs) -> "SycamorePrompt":
                 new.__dict__[k] = v
         return new
 
+    def is_done(self, s: str) -> bool:
```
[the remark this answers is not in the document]
HenryL27: Yep, that's the idea. In extract entities with a tokenizer we set this to `s != "None"`, which is the condition it used to use. I'm not sure that the prompt is the right place for this piece of logic to live, though I guess the prompt is also what decides whether to make a sequence of renderedprompts so maybe?

lib/sycamore/sycamore/llms/prompts/default_prompts.py
HenryL27: I have no idea where the all caps prompts are being used, but I don't think chucking them would really do anything for me.
There's a world where after converting everything over and deleting generate_old we can delete all the bwc prompts, since they adhere to a different (obsolete) interface.

lib/sycamore/sycamore/llms/llms.py
HenryL27: Something done by the LLM - OpenAI and Anthropic have batch apis now: [Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/batch-processing) /  [OpenAI](https://platform.openai.com/docs/guides/batch). I think similar things exist in bedrock/azure as well but less sure

lib/sycamore/sycamore/llms/llms.py
HenryL27: yep looks like you get a TypeError if you give it something that isn't a class

lib/sycamore/sycamore/llms/openai.py
HenryL27: yes it is and yes we should use it :)

lib/sycamore/sycamore/llms/openai.py
HenryL27: yep looks like this is not used and can be dropped

lib/sycamore/sycamore/transforms/base_llm.py
HenryL27: oops

lib/sycamore/sycamore/transforms/llm_filter.py
HenryL27: oops

lib/sycamore/sycamore/docset.py
HenryL27: the plan is to deprecate just about everything on my sprint tasks

lib/sycamore/sycamore/transforms/base_llm.py
HenryL27: I think that makes sense. You'd do something like `doc.properties[f"{self._entity_name}_nth_llm_call"] = i` in the llmmap I guess and then the post-map could just reference that prop. 

I also though about trying to emit metadata docs to keep track of all of this, but I don't know how practical that would be to capture and use on the other side - the whole point is that we want to know which part of the document the answer came from.

I know there was a proposal at some point to separate 'user' properties from 'system' properties. Did that ever go anywhere?

lib/sycamore/sycamore/transforms/base_llm.py
HenryL27: One thought I had which I think is beyond the scope of this is some sort of conditional branching/looping logic in a sycamore pipeline. Like
```python
docset.do()
    .llm_map(<extract entity prompt>)
    .map(if not json set back to 'None')
    .while(lambda d: d.properties.get('entity', 'None') == 'None')
    .continue_processing()
```
I'm sure that creates all sorts of theoretic problems.

But then prompts can go back to being a single prompt and we can increment a counter on the document object and stuff.

I guess we can do the to-render-object counter thing and keep prompts to single-render in a for loop in llm_map which is probably cleaner. I'll try it and see

HenryL27:
> Is there a reasonable path to replace the generate_old calls at some point, do you think?

The goal is for generate_old to be deleted entirely. git grep claims there are 28 uses (10 of them are unit test mocks). Most of the rest of them are in the path of the replacements I have planned. There's 3 more in sycamore query that I'm hoping should be relatively simple to switch over, and the constant warnings should annoy vinayak into doing it (or I can if it isn't done by the time I try to remove generate_old).

> Also, have you gone through the integ test failures to see if they are a related?

Yep. Of the 7 failing ITs I think 4 are directly related to these changes - with generate_old not quite handling images and response_formats correctly. My plan was to let those break and then fix them when I dealt with their associated transforms. Maybe they need pytest skip marks then?

> Anything else we need to test to get confidence here?

The docprep IT's passed (except for the pinecone what that gets rate-limited and dies). I can try to figure out how to test random things in AddDoc - probably should know that anyway. But overall I feel like passing sycamore test suite (barring the 4 guys above) is probably good. Should make sure that AddDoc doesn't hit the bad paths in the failing ITs though.

HenryL27:
update: was able to run Add a doc with docstore using this

bsowell (approved):
I think your plan seems reasonable. You'll have to keep an eye on things when you push this as it potentially has a big blast radius, but I don't think I see any reason to wait.

PR #1130, opened 2025-01-26
Add Opensearch Writer Reliability

dhruvkaliraman7:
Adding a reliability feature to opensearch write which:
1. Only allows Materialize -> Write pipeline ( If multiple datasets, should be present in 1 materialized directory)
2. Deletes an index if it already exists.
3. Asserts that chunks in materialized directory equals chunks in open search.


lib/sycamore/sycamore/writer.py
```diff
@@ -136,7 +137,22 @@ def _convert_to_host_port_list(hostlist: Any) -> list[HostAndPort]:
         client_params = OpenSearchWriterClientParams(**os_client_args)
 
         target_params: OpenSearchWriterTargetParams
-        target_params_dict: dict[str, Any] = {"index_name": index_name}
+        target_params_dict: dict[str, Any] = {
```
eric-anderson, line 140: similar long function comment.  I think:
```
def _get_target_params(self, ...args...) -> OpenSearchWriterTargetParams:
  Lines 140..161
```
would fix it.

lib/sycamore/sycamore/writer.py
```diff
@@ -800,10 +822,17 @@ def json(
...
-    def _maybe_execute(self, node: Node, execute: bool) -> Optional[DocSet]:
+    def _maybe_execute(
+        self, node: Node, execute: bool, client: Optional[OpenSearchWriterClient] = None
+    ) -> Optional[DocSet]:
         ds = DocSet(self.context, node)
         if not execute:
             return ds
 
         ds.execute()
+        if client is not None:
```
eric-anderson, line 833: fix this to just call client.reliabliity_assertor(...) and do the raise in the client base class.

eric-anderson (approved):

PR #1132, opened 2025-01-27
Clean Up Dead Code

karanataryn:


.pre-commit-config.yaml
```diff
@@ -32,28 +32,6 @@ repos:
           args: [--install-types, --explicit-package-bases, --non-interactive]
           exclude: docs/
 
-  - repo: local
-    hooks:
-      - id: pytest-check
-        name: pytest-check
-        # Needed because otherwise it seems to always run pytest in the git root.
-        entry: poetry run pytest lib/sycamore/sycamore/tests/unit apps/crawler/crawler/s3/tests/unit
```
bsowell, line 40: This looks like it removes the sycamore unit tests as well? Do you not just want to remove the crawler part? 

I don't use pre-commit any more, so not sure if this works at all, but if you are keeping it it seems like it makes sense to run the sycamore unit tests.

apps/weaviate/compose.yml
bsowell, line 1: Did we decide to remove this? Can't remember. I have used it when testing Weaviate. Is it used in the integ tests?

baitsguy:
Query-eval stuff looks good to go, can merge after Ben's comments

apps/weaviate/compose.yml
karanataryn: This was used to test if the one in the docs (https://sycamore.readthedocs.io/en/stable/sycamore/connectors/weaviate.html) works correctly, so we don't need this any more.

.pre-commit-config.yaml
karanataryn: Good catch, changed.

bsowell (approved):

sohamkasar19 (approved):

PR #1135, opened 2025-01-28
Add jupyter utils (Finra upstream)

dhruvkaliraman7:
Adding 3 utilities:
1. slow_pprint:  Prints large outputs slowly to prevent Jupyter from dropping output.
2. bound_memory: limit memory usage.
3. reimport: reload a module

### DIFF
--- a/lib/sycamore/sycamore/utils/jupyter.py
+++ b/lib/sycamore/sycamore/utils/jupyter.py
@@ -36,8 +36,11 @@ def bound_memory(gb: int = 4) -> None:
         gb: Memory limit in gigabytes.
     """
     import resource
+    import platform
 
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/utils/jupyter.py
```diff
@@ -0,0 +1,63 @@
...
+def bound_memory(gb: int = 4) -> None:
+    """
+    Limits the process's memory usage.
+
+    Args:
+        gb: Memory limit in gigabytes.
+    """
+    import resource
+
+    # Works on Linux, not on macOS.
```
baitsguy, line 40: can you print a message here if its being used on non-linux

baitsguy (approved):

PR #1138, opened 2025-01-28
Explicitly add tiktoken and relock.

bsowell:


karanataryn (approved):

lib/sycamore/pyproject.toml
```diff
@@ -88,6 +88,7 @@ nltk = { version = "^3.9.0", optional = true }
 
 # LLM Dependencies
 anthropic = { version = "^0.42.0", optional = true }
```
karanataryn, line 90: Is `openai` not an LLM dependency?

HenryL27 (approved):
weird diff in lib/sycamore/poetry.lock

lib/sycamore/pyproject.toml
bsowell: No, it's in the default install for now. I created the anthropic one at the hackathon. We could eventually move openai, but since it's a breaking change, I'd rather be more careful about timing.

PR #1142, opened 2025-01-29
Fix self-reported `aryn-sdk` version

MarkLindblad:


MarkLindblad:
Since we decided not to use importlib to report the version from within `aryn-sdk` so that aryn-sdk code can be copy-pasted by users out of the package, I will add add a test in a future pr that checks our hardcoded self-reported `aryn-sdk` version matches the one importlib expects.

PR #1145, opened 2025-01-30
Add OpenSearch shard related logging

baitsguy:
Ran locally and logged the opposite condition to make sure it's working

### DIFF
--- a/lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
+++ b/lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
@@ -63,8 +63,8 @@ class OpenSearchReaderClient(BaseDBReader.Client):
             query_params.kwargs["_source_includes"] = query_params.doc_reconstructor.get_required_source_fields()
         # No pagination needed for knn queries
         if "query" in query_params.query and "knn" in query_params.query["query"]:
-            response = self._client.search(
-                index=query_params.index_name, body=query_params.query, **query_params.kwargs
+            response = _search_with_logging(
+                self._client, index=query_params.index_name, body=query_params.query, **query_params.kwargs
             )
             hits = response["hits"]["hits"]
             if hits:
@@ -73,8 +73,8 @@ class OpenSearchReaderClient(BaseDBReader.Client):
         else:
             if "scroll" not in query_params.kwargs:
                 query_params.kwargs["scroll"] = "10m"
-            response = self._client.search(
-                index=query_params.index_name, body=query_params.query, **query_params.kwargs
+            response = _search_with_logging(
+                self._client, index=query_params.index_name, body=query_params.query, **query_params.kwargs
             )
             scroll_id = response["_scroll_id"]
             try:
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/connectors/opensearch/utils.py
```diff
@@ -1,10 +1,26 @@
...
+from typing import Optional, Any
+
+from opensearchpy import OpenSearch
 
 from sycamore import Context
 from sycamore.context import context_params
 from sycamore.transforms import Embedder
 
 
+logger = logging.getLogger(__name__)
```
austintlee, line 11: Let's name this logger "opensearch".

lib/sycamore/sycamore/connectors/opensearch/utils.py
```diff
@@ -1,10 +1,26 @@
...
+
+
+class OpenSearchClientWithLogging(OpenSearch):
+    def search(self, **kwargs) -> Any:
+        """Helper method to execute OpenSearch search queries, and silent errors."""
+        response = super().search(**kwargs)
+        shards = response.get("_shards", {})
+        if shards.get("total") != shards.get("successful"):
+            logger.error(f"OpenSearch query skipped shards: {response}")
```
austintlee, line 20: It might be helpful to fish out "query" from kwargs (if it's present) and log it along with the response?  Do you think we need to log the entire response payload?  It could be big.

austintlee (approved):
Looks good.  A couple of suggestions.

lib/sycamore/sycamore/connectors/opensearch/utils.py
baitsguy: Ack

lib/sycamore/sycamore/connectors/opensearch/utils.py
baitsguy: I'll leave this in for now while we're debugging so we know which results actually got back. If it happens frequently enough and we handle it we can clean

lib/sycamore/sycamore/connectors/opensearch/utils.py
```diff
@@ -1,10 +1,26 @@
-from typing import Optional
+import logging
+from typing import Optional, Any
+
+from opensearchpy import OpenSearch
```
HenryL27, line 4: here's the unconditional import

lib/sycamore/sycamore/transforms/query.py
```diff
@@ -26,10 +28,9 @@ def __init__(self, os_client_args: dict) -> None:
 
     @requires_modules("opensearchpy", extra="opensearch")
     def query(self, query: OpenSearchQuery) -> OpenSearchQueryResult:
-        from opensearchpy import OpenSearch
```
HenryL27, line 29: prob need to keep the import sycamore.connectors.opensearch.utils here instead of at the top

lib/sycamore/sycamore/query/client.py
```diff
@@ -131,7 +133,6 @@ def __init__(
         llm: Optional[Union[LLM, str]] = None,
         query_plan_strategy: Optional[QueryPlanStrategy] = None,
     ):
-        from opensearchpy import OpenSearch
```
HenryL27, line 134: also here

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -41,10 +42,8 @@ def __init__(self, client: "OpenSearch"):
     @classmethod
     @requires_modules("opensearchpy", extra="opensearch")
     def from_client_params(cls, params: BaseDBReader.ClientParams) -> "OpenSearchReaderClient":
-        from opensearchpy import OpenSearch
```
HenryL27, line 44: also here

PR #1147, opened 2025-01-31
Aryn connectors for reading and writing docsets

austintlee:


lib/sycamore/sycamore/connectors/aryn/ArynReader.py
```diff
@@ -0,0 +1,83 @@
...
+class ArynQueryParams(BaseDBReader.QueryParams):
+    def __init__(self, docset_id: str):
+        self.docset_id = docset_id
+
+
+class ArynQueryResponse(BaseDBReader.QueryResponse):
+    def __init__(self, docs: list[dict[str, Any]]):
+        self.docs = docs
+
+    def to_docs(self, query_params: "BaseDBReader.QueryParams") -> list[Document]:
```
HenryL27, line 35: Some day I feel like we should turn these lists that we're passing around into iterators/generators (all the way to the ray ds construction) but not right now and that applies to all the readers

lib/sycamore/sycamore/connectors/aryn/ArynWriter.py
```diff
@@ -0,0 +1,79 @@
...
+            files: Mapping = {"doc": doc.serialize()}
+            res = requests.post(url=f"{self.aryn_url}/docsets/write",
+                                params={"docset_id": docset_id},
+                                files=files, headers=headers)
+
+    def create_target_idempotent(self, target_params: "BaseDBWriter.TargetParams"):
+        pass
+
+    def get_existing_target_params(self, target_params: "BaseDBWriter.TargetParams") -> "BaseDBWriter.TargetParams":
+        pass
```
HenryL27, line 72: Is it a huge lift to add docset-creation functionality to this?

lib/sycamore/sycamore/reader.py
```diff
@@ -632,3 +634,23 @@ def qdrant(self, client_params: dict, query_params: dict, **kwargs) -> DocSet:
...
+            kwargs: Keyword arguments to pass to the underlying execution engine.
+        """
+        from sycamore.connectors.aryn.ArynReader import (
+            ArynReader,
+            ArynClientParams,
+            ArynQueryParams,
+        )
+
+        api_key = ArynConfig.get_aryn_api_key()
+        aryn_url = ArynConfig.get_aryn_url()
```
HenryL27, line 653: should be able to pass these in as params here too I think

lib/sycamore/sycamore/reader.py
```diff
@@ -632,3 +634,23 @@ def qdrant(self, client_params: dict, query_params: dict, **kwargs) -> DocSet:
             **kwargs,
         )
         return DocSet(self._context, wr)
+
+    @context_params
+    def aryn(self, docset_id: str, **kwargs) -> DocSet:
```
HenryL27, line 639: What params come from the context?

lib/sycamore/sycamore/writer.py
```diff
@@ -800,6 +804,38 @@ def json(
...
+        """
+
+        from sycamore.connectors.aryn.ArynWriter import (
+            ArynWriter,
+            ArynWriterClientParams,
+            ArynWriterTargetParams,
+        )
+
+        api_key = ArynConfig.get_aryn_api_key()
+        aryn_url = ArynConfig.get_aryn_url()
```
HenryL27, line 825: these guys too

lib/sycamore/sycamore/writer.py
```diff
@@ -800,6 +804,38 @@ def json(
...
+        if docset_id is None and create_new_docset and name is not None:
+            headers = {
+                "Authorization": f"Bearer {api_key}"
+            }
+            res = requests.post(url=f"{aryn_url}/docsets", data={"name": name}, headers=headers)
+            docset_id = res.json()["docset_id"]
```
HenryL27, line 831: I guess docstore allows multiple docsets with the same name?

austintlee: Yes, only IDs are unique.

lib/sycamore/sycamore/reader.py
austintlee: I don't think I need this.  I'll remove it.

lib/sycamore/sycamore/reader.py
austintlee: ok

lib/sycamore/sycamore/writer.py
austintlee: ok

lib/sycamore/sycamore/connectors/aryn/ArynWriter.py
austintlee: If there are multiple writer tasks, will each one attempt to create a docset?  Since `name` is not unique, creating a docset from multiple writer tasks will result in multiple docsets (with the same name) being created.

HenryL27 (approved):

PR #1156, opened 2025-02-05
Make async DocParse methods in `aryn-sdk` not operate on non-DocParse async tasks

MarkLindblad:
The goal is to make `aryn-sdk` harder to use wrong: result won't "see" non-DocParse tasks and cancel won't "see" non-DocParse tasks.

### DIFF
--- a/lib/aryn-sdk/aryn_sdk/partition/partition.py
+++ b/lib/aryn-sdk/aryn_sdk/partition/partition.py
@@ -530,7 +530,15 @@ def partition_file_async_list(
     response = requests.get(
         async_list_url, params=g_parameters, headers=headers, stream=_should_stream(), verify=ssl_verify
     )
-    return response.json()["tasks"]
+
+    # Once filtering has been added to Aryn async, we can stop filtering below and make this logic just delete the path
+    all_tasks = response.json()["tasks"]
+    result = {}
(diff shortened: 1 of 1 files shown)


lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -22,6 +22,7 @@
 _logger.addHandler(logging.StreamHandler(sys.stderr))
 
 g_version = "0.1.13"
+g_parameters = {"path_filter": "/document/partition"}
```
alexaryn, line 25: Do we want `^/v1/document/partition/`

alexaryn (approved):

lib/aryn-sdk/aryn_sdk/partition/partition.py
MarkLindblad: Do we want this to support just v1?

MarkLindblad: And did you mean `^/v1/document/partition$`? These paths do not end in /

MarkLindblad: I've now changed the filter to `^/v1/document/partition$`.

PR #1157, opened 2025-02-06
add pytest-xdist to root sycamore and pre-commit so it goes faster

HenryL27:
sorry about the yaml autoformatting


alexaryn (commented):
This doesn't work on my Linux laptop, but I don't use pre-commit.

HenryL27:
Only saved about 1 minute and causes some people's laptops to explode, so closing

PR #1159, opened 2025-02-06
Ensure parent docs are collected during doc reconstruct

austintlee:


lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -333,8 +333,17 @@ def _to_parent_doc(self, slice_query: dict[str, Any]) -> List[dict[str, Any]]:
...
                     ):
-                        results.append(hit)
+                        # Only add a child doc whose parent_id has not been found, yet.
                         parent_ids.add(hit["_source"]["parent_id"])
+                        results.append(hit)
+                    elif ("parent_id" not in hit["_source"] or hit["_source"]["parent_id"] is None) and hit[
+                        "_id"
+                    ] not in parent_ids:
+                        # Add a parent doc if it's a match.
+                        parent_id = hit["_id"]
```
baitsguy, line 343: do we actually use _id? I thought all our indexing is done by _doc_id

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -333,8 +333,17 @@ def _to_parent_doc(self, slice_query: dict[str, Any]) -> List[dict[str, Any]]:
...
+                        parent_ids.add(parent_id)
+                        hit["_source"]["parent_id"] = parent_id
```
HenryL27, line 345: why do we need to add the loopback?

HenryL27 (approved):
lgtm so long is new ITs pass

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
austintlee: I think we always pass document.doc_id as _id when we call OpenSearch.

PR #1161, opened 2025-02-07
[llm unify 5a/n] Add JinjaPompt and re-convert extract entities

HenryL27:
Also refactors entity extractor to break up the megafunction

### DIFF
--- a/lib/sycamore/sycamore/transforms/extract_entity.py
+++ b/lib/sycamore/sycamore/transforms/extract_entity.py
@@ -254,6 +254,8 @@ class OpenAIEntityExtractor(EntityExtractor):
 
         def postprocess(d: Document) -> Document:
             target_club_idx = d.properties[vars["iteration_var_name"]]
+            if target_club_idx >= len(d.properties[vars["batch_key"]]):
+                return d
             batch = d.properties[vars["batch_key"]][target_club_idx]
             d.properties[vars["source_idx_key"]] = batch
             return d
@@ -283,6 +285,7 @@ class OpenAIEntityExtractor(EntityExtractor):
             llm,
             validate=validate,
             iteration_var=vars["iteration_var_name"],
+            max_tries=100,
             **kwargs,
         )
         nodes.append(head_node)
--- a/lib/sycamore/sycamore/transforms/extract_entity.py
+++ b/lib/sycamore/sycamore/transforms/extract_entity.py
@@ -222,6 +222,7 @@ class OpenAIEntityExtractor(EntityExtractor):
                         curr_tks += tks
                 batches.append(curr_club)
             else:
+                # If no tokenizer, we run a single batch with the first num_of_elements.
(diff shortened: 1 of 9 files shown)


lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -129,133 +130,170 @@ def __init__(
...
+            llm = self._llm
+        assert llm is not None, "Could not find an LLM to use"
+
+        prompt = self._get_prompt()
+        preprocess = self._make_preprocess_fn(prompt)
+        vars = self._get_const_variables()
+
+        def validate(d: Document) -> bool:
+            if self._tokenizer is not None:
+                return d.properties.get(self._entity_name, "None") != "None"
```
baitsguy, line 251: self._entity_name in d.properties?

lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -129,133 +130,170 @@ def __init__(
...
+            else:
+                return True
```
baitsguy, line 253: can just combine the if else

lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -129,133 +130,170 @@ def __init__(
...
-                    if len(last_eclub) > 0 and e.properties[source_idx_key] != last_eclub:
-                        club_idx += 1
-                    last_eclub = e.properties[source_idx_key]
-                    if club_idx == target_club_idx:
-                        d.properties[source_idx_key] = last_eclub
-                        break
-                return d
+                batches.append(curr_club)
+            else:
+                batches = [[i for e, i in elements[: self._num_of_elements]]]
```
baitsguy, line 225: mind adding a comment saying if no tokenizer we process each element separately because we don't know how many we can combine?

lib/sycamore/sycamore/transforms/extract_entity.py
```diff
@@ -1,23 +1,24 @@
 from abc import ABC, abstractmethod
-from typing import Callable, Any, Optional, Union, cast
+from typing import Callable, Any, Optional, Union
 
 from sycamore.context import Context, context_params, OperationTypes
 from sycamore.data import Element, Document
 from sycamore.llms import LLM
 from sycamore.llms.prompts.default_prompts import (
-    EntityExtractorZeroShotGuidancePrompt,
```
baitsguy, line 8: can we start deleting these if we're not using? the set of prompts is growing

baitsguy (commented):
Some cleanup requests but functionally looks good

lib/sycamore/sycamore/transforms/extract_entity.py
baitsguy: discussed offline, we need to check for "None" because that's what the llm returns if no val

PR #1162, opened 2025-02-07
fix extract properties again

HenryL27:


bsowell (approved):
Hard for me to eyeball, but let's get it in and I can test.

PR #1170, opened 2025-02-10
Rename async list endpoints to "action" from "path"

MarkLindblad:


lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -533,7 +533,10 @@ def partition_file_async_list(
 
     result = response.json()["tasks"]
     for v in result.values():
-        del v["path"]
+        if "path" in v:
+            del v["path"]
+        if "action" in v:
+            del v["action"]
```
alexaryn, line 539: Don't worry about "path" because it'll be gone before anyone outside the company can see it.  Also, avoid the double-lookup like this:
```
v.pop("path", None)
```

alexaryn (approved):

lib/aryn-sdk/aryn_sdk/partition/partition.py
MarkLindblad: Done.

PR #1171, opened 2025-02-11
Serialize query strings to avoid Ray Dataset column imputation

austintlee:


HenryL27 (approved):
for all the other sycamore->sycamore transforms we use pickle - 
https://github.com/aryn-ai/sycamore/blob/43758ca85b1120e28df73d56242051babadb01cd/lib/sycamore/sycamore/data/document.py#L182

I'll leave it to you whether to go pickle or json but just flagging. (some items aren't jsonable, but prob not an issue here?)

PR #1176, opened 2025-02-12
Add Gemini LLM and Summarizer

karanataryn:
This pull request introduces support for the Google Gemini language model in the Sycamore project. The changes include adding the necessary dependencies, integrating the Gemini model, adding it as a summarizer, and implementing unit tests to ensure its functionality.


lib/sycamore/sycamore/llms/gemini.py
```diff
@@ -0,0 +1,133 @@
...
+
+class Gemini(LLM):
+    """This is an LLM implementation that uses the Google Gemini API to generate text.
+
+    Args:
+        model_name: The name of the Gemini model to use.
+        cache: A cache object to use for caching results.
+    """
+
+    @requires_modules("google-genai")
```
HenryL27, line 47: I think you also need to add this as an extra in the pyproject.toml and then in this decorator go `extra="google-genai"` or whatev. not 100% certain, it might all default correctly

lib/sycamore/sycamore/llms/gemini.py
```diff
@@ -0,0 +1,133 @@
...
+        """Returns True if the LLM is in chat mode, False otherwise."""
+        return True
+
+    def get_generate_kwargs(self, prompt: RenderedPrompt, llm_kwargs: Optional[dict] = None) -> dict:
+        kwargs: dict[str, Any] = {}
+        config = {
+            "temperature": 0,
+            "candidate_count": 1,
+            **(llm_kwargs or {}),
+        }
```
HenryL27, line 83: RenderedPrompts can do structured outputs and it looks like so can gemini? 
https://ai.google.dev/gemini-api/docs/structured-output?lang=python#supply-schema-in-config

lib/sycamore/sycamore/transforms/summarize_images.py
```diff
@@ -81,6 +82,23 @@ def __init__(self, llm: LLM, prompt: Optional[str] = None, include_context: bool
...
+        Args:
+            image: The image to summarize.
+            context: The context to use for summarization.
+
+        Returns:
+            The summarized image as a string.
+        """
+        messages = []
+        if context is not None and self.include_context:
+            messages = [RenderedMessage(role="system", content=context)]
```
HenryL27, line 97: I think if context is provided you want to include it regardless of is self.include_context - include_context is more a control for automatically picking up data from documents, which doesn't really apply here.

HenryL27 (approved):

lib/sycamore/sycamore/llms/gemini.py
karanataryn: Changed.

lib/sycamore/sycamore/llms/gemini.py
```diff
@@ -0,0 +1,137 @@
...
+        kwargs["config"] = None
+        if config:
+            kwargs["config"] = types.GenerateContentConfig(**config)
+        kwargs["content"] = content
+        return kwargs
+
+    def generate_metadata(self, *, prompt: RenderedPrompt, llm_kwargs: Optional[dict] = None) -> dict:
+        ret = self._llm_cache_get(prompt, llm_kwargs)
+        if isinstance(ret, dict):
+            print(f"cache return {ret}")
```
HenryL27, line 110: needed?

lib/sycamore/sycamore/transforms/summarize_images.py
```diff
@@ -81,6 +82,23 @@ def __init__(self, llm: LLM, prompt: Optional[str] = None, include_context: bool
         self.prompt = prompt
         self.include_context = include_context
 
+    def summarize_image(self, image: Image.Image, context: Optional[str]) -> str:
```
HenryL27, line 85: what do you think about "context" -> "system" ? idk.

HenryL27 (approved):

lib/sycamore/sycamore/transforms/summarize_images.py
karanataryn: I like context for now, we can change later based on usage.

PR #1180, opened 2025-02-13
Move to UV

karanataryn:
Move poetry to uv, which has much faster locking and install times.


lib/uv-lock/poetry-lock-all.sh
HenryL27, line 1: this script does a bunch of poetry stuff right? I think switching needs to be more high-touch here

pyproject.toml
HenryL27, line 1: I had kinda hoped that the root could be turned into a uv workspace - https://docs.astral.sh/uv/concepts/projects/workspaces/

karanataryn: I think I do that, look at the `[tool.uv.workspace]` below. Let me know if you imagined something else.

lib/uv-lock/poetry-lock-all.sh
karanataryn: Good catch, have fixed it.

pyproject.toml
HenryL27: oh look at that I'm blind! I wonder how much of the [dependency-groups] section that lets you delete?

PR #1181, opened 2025-02-13
Bump Ray to Fix Security Issue

karanataryn:
Fixes issue here: https://github.com/aryn-ai/sycamore/security/dependabot/337. Also had to change `apps/jupyter` sycamore dependencies a bit to get locking to work.


poetry.lock
```diff
@@ -7873,25 +7790,6 @@ files = [
     {file = "rfc3986_validator-0.1.1.tar.gz", hash = "sha256:3d44bde7921b3b9ec3ae4e3adca370438eccebc676456449b145d533b240d055"},
 ]
 
-[[package]]
-name = "rich"
-version = "13.9.4"
-description = "Render rich text, tables, progress bars, syntax highlighting, markdown and more to the terminal"
```
HenryL27, line 7879: IIRC sycamore query stuff uses this library. can you confirm that it still works?

HenryL27 (approved):
I think this is right? Little worried about rich.

HenryL27 (changes requested):
Actually it looks like dropping rich killed the integ tests

poetry.lock
karanataryn: Good catch, fixed.

HenryL27 (approved):

PR #1183, opened 2025-02-13
Adding clustering and groupby in luna planner part

bohou-aryn:


lib/sycamore/sycamore/docset.py
```diff
@@ -931,23 +931,31 @@ def kmeans(self, K: int, iterations: int = 20, init_mode: str = "random", epsilo
...
+            doc = Document.from_row(row)
+            return not isinstance(doc, MetadataDocument)
+
         def init_embedding(row):
             doc = Document.from_row(row)
-            return {"vector": doc.embedding, "cluster": -1}
+            return {"vector": doc.embedding, "cluster": -1} if field_name is None else {"vector": doc[field_name], "cluster": -1}
 
-        embeddings = self.plan.execute().map(init_embedding).materialize()
+        embeddings = self.plan.execute().filter(filter_meta).map(init_embedding).materialize()
```
baitsguy, line 946: what's the materialize for? also can we not do

`self.plan.filter(filter_meta).map(init_embedding).execute()`

I think that'll use ray versions of the ops right? so theoretically faster

baitsguy: You can also just pull our the filter_meta method somewhere

lib/sycamore/sycamore/grouped_data.py
```diff
@@ -14,7 +14,12 @@ def __init__(self, docset: DocSet, key):
 
     def aggregate(self, f: "AggregateFn") -> DocSet:
         dataset = self._docset.plan.execute()
-        grouped = dataset.map(Document.from_row).groupby(self._key)
+
+        def filter_meta(row):
```
baitsguy, line 18: pull out somewhere

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -613,6 +615,77 @@ def script(self, input_var: Optional[str] = None, output_var: Optional[str] = No
...
+        clustered = docset.clustering(centroids=centroids, cluster_field_name=cluster_field_name, field_name=field)
+        result = clustered.groupby(cluster_field_name).count().sort(descending, "properties.count", 0).limit(5)
+
+        return result
+
+    def script(self, input_var: Optional[str] = None, output_var: Optional[str] = None) -> Tuple[str, List[str]]:
+        raise Exception("New Top K not implemented for codegen")
+
+
+class SycamoreEmbed(SycamoreOperator):
```
baitsguy, line 658: what's this for? don't think the planner uses it. My current thought is this (embedding) should just be an implementation details within group_by etc?

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -613,6 +615,77 @@ def script(self, input_var: Optional[str] = None, output_var: Optional[str] = No
         return result, []
 
 
+class SycamoreNewTopK(SycamoreOperator):
```
baitsguy, line 618: Let's discuss the new operators we want to add. 

I'd imagine we want: topK, groupBy

We should just be able to nuke the existing topK rather than adding an new one

lib/sycamore/sycamore/docset.py
```diff
@@ -922,7 +922,7 @@ def map(self, f: Callable[[Document], Document], **resource_args) -> "DocSet":
         mapping = Map(self.plan, f=f, **resource_args)
         return DocSet(self.context, mapping)
 
-    def kmeans(self, K: int, iterations: int = 20, init_mode: str = "random", epsilon: float = 1e-4):
+    def kmeans(self, K: int, iterations: int = 20, init_mode: str = "random", epsilon: float = 1e-4, field_name: Optional[str] = None):
```
baitsguy, line 925: the query functionality currently works on reconstructed documents, whereas this is working on exploded docsets (from what I can tell). We probably want the former since we want to group documents?

baitsguy (commented):
Code looks good, some clarifying questions

lib/sycamore/sycamore/data/document.py
```diff
@@ -51,6 +51,13 @@ def __init__(self, document=None, /, **kwargs):
         if "lineage_id" not in self.data:
             self.update_lineage_id()
 
+    def get_by_path(self, path: str):
```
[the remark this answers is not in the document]
bohou-aryn: seems similar, would try to reuse that one.

lib/sycamore/sycamore/docset.py
bohou-aryn: that is for reuse the embeddings, otherwise, just compute twice from beginning, would have one option for configuring this.

lib/sycamore/sycamore/query/execution/sycamore_operator.py
bohou-aryn: inside groupby just means every time run topk, it would do this embedding over all items. Not sure about what role the Luna plays, does it also write?

lib/sycamore/sycamore/docset.py
bohou-aryn: not sure get this one

baitsguy: discussed offline, this is working on document level fields. If the caller wants to run it on text_representation they need to create a top-level field that embeddings all text_representations from the docs elements

lib/sycamore/sycamore/query/operators/top_k.py
```diff
@@ -36,3 +36,24 @@ class TopK(Node):
     """An instruction of what the groups should be about if llm_cluster is True. E.g. if the
     purpose of this operation is to find the top 2 most frequent cities, llm_cluster_instruction
     could be 'Form groups of different food'"""
+
+
+class NewTopK(Node):
```
baitsguy, line 41: GroupByCount

baitsguy (approved):

PR #1190, opened 2025-02-19
Helper script for getting git credentials from the environment

eric-anderson:


apps/git/git-credential-from-env.py
```diff
@@ -0,0 +1,46 @@
...
+if "CUSTOMER_USER" in os.environ and "CUSTOMER_KEY" in os.environ:
+    print("protocol=https")
+    print("host=github.com")
+    print(f"username={os.environ['CUSTOMER_USER']}")
+    print(f"password={os.environ['CUSTOMER_KEY']}")
+    logging.error(f"git-credentials-from-env helper: Customer user was used for {d['path']}")
+    exit(0)
+
+logging.error(f"WARNING from {__file__}: Unable to find CUSTOMER_USER and CUSTOMER_KEY in environ.")
+logging.error(f"WARNING since the helper was enabled, this is probably an error.")
```
Soeb-aryn, line 45: we might not need f-string here.

Soeb-aryn (approved):
LGTM !

apps/git/git-credential-from-env.py
eric-anderson: Fixed.

PR #1192, opened 2025-02-21
[llm unify 7/n] Summarize

HenryL27:
Again, general lack of confidence in this.

Turns summarize document from an iterative folding strategy to a heirarchical strategy.
Uses math + jinja to generate a summary every k elements for the next k elements (and then repeat with k^2 and a stride of k, etc until k^n > n_elements).

Integrates into summarize_data by slightly changing how that reduce happens. Had to make a separate (similar) prompt for that. Should probably factor most of the jinja logic out as fragments.

I probably broke something in luna but all the unittests passed so idk.

Also not sure about some of the names.

### DIFF
--- a/lib/sycamore/sycamore/transforms/summarize.py
+++ b/lib/sycamore/sycamore/transforms/summarize.py
@@ -248,11 +248,8 @@ class MultiStepDocumentSummarizer(Summarizer):
         fields: List of fields to include in each element's representation in the prompt. Specify
             with dotted notation (e.g. properties.title). End the list with `EtCetera` to add all fields
             (previously specified fields go first). Default is [] which includes no fields.
-        max_tokens: token limit for each summarization. Default is 10k (default tokenizer is by character).
         tokenizer: tokenizer to use when computing how many tokens a prompt will take. Default is
             CharacterTokenizer
-        max_rounds: max number of rounds of heirarchical summarization to perform. The number of elements that
-            can be included in the summary is O(e^max_rounds), so can be small. Default is 4.
     """
 
     def __init__(
@@ -408,8 +405,6 @@ class OneStepDocumentSummarizer(Summarizer):
         llm: LLM to use for summarization
         question: Question to use as context for the summary. The llm will attempt to
             use the data provided to answer the question.
-        token_limit: Token limit for the prompt. Default is 10k (default tokenizer is
-            by character)
         tokenizer: Tokenizer to use to count tokens (to not exceed the token limit).
             Default is CharacterTokenizer
         fields: List of fields to include from every element. To include any additional


lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -463,6 +474,154 @@ def __init__(self, question: str, text: str):
...
 
+SummarizeBranchingFactorJinjaPrompt = JinjaElementPrompt(
+    system=textwrap.dedent(
+        """
+        {% if question is defined %}You are a helpful research assistant. You answer questions based on text you are presented with.
+        {% else %}You are a helpful data summarizer. You concisely summarize text you are presented with, including as much detail as possible.
+        {% endif %}
+        """
+    ),
+    user=J_GET_ELEMENT_TEXT_MACRO
```
eric-anderson, line 485: Can we force the question to always be present by having a default question of "What is the summary of this information?"

This feels overly complex, and experience with FINRA was that complexity leads to weird prompts that don't quite do what you want.

lib/sycamore/sycamore/llms/prompts/default_prompts.py
```diff
@@ -463,6 +474,154 @@ def __init__(self, question: str, text: str):
...
+SummarizeDataHeirarchicalPrompt = JinjaElementPrompt(
+    system=textwrap.dedent(
+        """
+        {% if question is defined %}You are a helpful research assistant. You answer questions based on text you are presented with.
+        {% else %}You are a helpful data summarizer. You concisely summarize text you are presented with, including as much detail as possible.
+        {% endif %}
+        """
+    ),
+    user=J_GET_ELEMENT_TEXT_MACRO
+    + textwrap.dedent(
```
eric-anderson, line 521: Do we want dedent or remove all leading whitespace.  It's weird to read code like:
```
if condition_a:
      if condition_ab
"a&b"
      else:
"a&!b"
else:
"!a"
```

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -52,14 +58,13 @@ def math_operation(val1: int, val2: int, operator: str) -> Union[int, float]:
...
     result_description: str,
     result_data: List[Any],
-    use_elements: bool = False,
-    num_elements: int = 5,
-    max_tokens: int = 120 * 1000,
-    tokenizer: Tokenizer = CharacterTokenizer(),
+    summaries_as_text: bool = False,
     context: Optional[Context] = None,
+    docset_summarizer: Optional[Type[Summarizer]] = None,
+    summarizer_kwargs: dict[str, Any] = {},
```
eric-anderson, line 67: Why do we need a class + kwargs rather than passing in an object?

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -52,14 +58,13 @@ def math_operation(val1: int, val2: int, operator: str) -> Union[int, float]:
 @context_params
 def summarize_data(
```
eric-anderson, line 59: Can we simplify this? It feels weird to have a summarize class and a whole ton of parameters in the call.

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -71,94 +76,100 @@ def summarize_data(
...
+        return summarize_data_docsets(
+            llm,
+            question,
+            result_data,
+            docset_summarizer=summarizer,
+            data_description=result_description,
+            summaries_as_text=summaries_as_text,
+        )
+
+    # If data is not DocSets, text is this list here
```
eric-anderson, line 114: Can we force data to always be docsets? If it somehow isn't convert it to a DocSet?

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -71,94 +76,100 @@ def summarize_data(
...
+        result_data = [ds.summarize(HeirarchicalDocumentSummarizer(llm)).map(sum_to_text) for ds in result_data]
+
+    single_docs = [_docset_to_singledoc(ds) for ds in result_data]
+    agged_ds = result_data[0].context.read.document(single_docs).summarize(docset_summarizer)
+    texts = [d.properties["summary"] for d in agged_ds.take_all()]
+    return "\n".join(texts)
+
+
+def _setup_docset_summarizer(summarizer_cls: Type[Summarizer], **kwargs) -> Summarizer:
+    if summarizer_cls is LLMElementTextSummarizer:
```
eric-anderson, line 150: This if class thing can't be the right way to do this.

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -143,7 +141,238 @@ def collapse(text: str, tokens_per_chunk: int, tokenizer: Tokenizer, summarizer_
     return cur_summary
 
 
-class DocumentSummarizer(Summarizer):
+class HeirarchicalDocumentSummarizer(Summarizer):
```
eric-anderson, line 144: It would have been nice to not have this all together in a single PR, this probably could have been separate.

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -143,7 +141,238 @@ def collapse(text: str, tokens_per_chunk: int, tokenizer: Tokenizer, summarizer_
...
+    """
+    Summarizes a document by constructing a heirarchical tree of batches of elements,
```
eric-anderson, line 146: I'm not sure we want this construction; I was expecting something that grouped until it ran out of context window. Do we have  a reason to believe that a multi-stage summarize is better than a single stage one?

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -143,7 +141,238 @@ def collapse(text: str, tokens_per_chunk: int, tokenizer: Tokenizer, summarizer_
...
+            validate=validate,
+            max_tries=20,  # If you hit this then your document has at least 2^20 elements in it.
+        )
+        postmap = Map(llm_map, f=cleanup)
+        comptransform = CompositeTransform(child, [])  # type: ignore
+        comptransform.nodes = [premap, llm_map, postmap]
+        return comptransform
+
+
+class MaxTokensHeirarchicalDocumentSummarizer(Summarizer):
```
eric-anderson, line 252: Big document, spread evenly across context windows of packed with a tail?
Multiple documents/ split only at document boundary or packed.
Split at properties or not?
How to split properties if they exceed context window.
Spread documents evenly or not, e.g. with 10 docs => 5,4,1; or 4,3,3?

lib/sycamore/sycamore/llms/prompts/default_prompts.py
HenryL27: I like to have the prompts render out with indentation for printability (though I think the llm doesn't really see it) - like
```
Element 0:
    properties.state_0: Alabama
    properties.state_1: Alaska
    ...
```

lib/sycamore/sycamore/query/execution/operations.py
HenryL27: According to vinayak if it's not docsets it's a single scalar (output of Count or Math operator). You could, I guess, wrap it in a Document and wrap that in a DocSet. Seems like hunting ducks with a bazooka. Also the data will look very different so you probably can't use the same prompting anyway

lib/sycamore/sycamore/query/execution/operations.py
HenryL27: duh. fixed

lib/sycamore/sycamore/query/execution/operations.py
HenryL27: it's not.

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -61,133 +73,324 @@ def __init__(self, llm: LLM, element_operator: Optional[Callable[[Element], bool
...
+            {%- else -%}
+            Summary: {{ element.properties[intermediate_summary_key] }}
+            {% endif -%}
+        {% endmacro -%}
+
+        {%- macro get_data_description() -%}
+            {%- if data_description is defined -%}
+        {{ data_description }}
+            {%- else -%}
+        some documents
```
eric-anderson, line 127: a set of documents with properties for each document

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -61,133 +73,324 @@ def __init__(self, llm: LLM, element_operator: Optional[Callable[[Element], bool
...
+            "skip_me_key": "_skip_me",
+            "batch_key": "_batch",
+            "round_key": "_round",
+            "intermediate_summary_key": "_summary",
+        }
+
+    def prep_batches(self, doc: Document, round: int = 0) -> Document:
+        vars = self.get_const_vars()
+        for i, elt in enumerate(doc.elements):
+            elt.properties[vars["round_key"]] = round
```
eric-anderson, line 215: I think you can simplify this by calculating a list of batches, but that would require changes to the way llmmapelements works, so lets chat.  Without that, this will need a bunch of explanation.

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -61,133 +73,324 @@ def __init__(self, llm: LLM, element_operator: Optional[Callable[[Element], bool
...
+                e2.properties[vars["skip_me_key"]] = True
+        return doc
+
+    def cleanup(self, doc: Document) -> Document:
+        if len(doc.elements) == 0:
+            return doc
+        vars = self.get_const_vars()
+        doc.properties["summary"] = doc.elements[0].properties[vars["intermediate_summary_key"]]
+        for e in doc.elements:
+            for v in vars:
```
eric-anderson, line 240: e.properties.pop(v, None)

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -61,133 +73,324 @@ def __init__(self, llm: LLM, element_operator: Optional[Callable[[Element], bool
...
+            if elt.properties[vars["skip_me_key"]]:
+                continue
+            this_batch = [i]
+            elt.properties[vars["batch_key"]] = this_batch
+            for j in range(i + 1, len(doc.elements)):
+                e2 = doc.elements[j]
+                if e2.properties.get(vars["skip_me_key"], False):
+                    continue
+                this_batch.append(j)
+                tks = self.prompt.render_element(elt, doc).token_count(self.tokenizer)
```
eric-anderson, line 227: If this was rendering only a single element, I would expect total_tks += self.prompt.render_element(elt, doc); to make the max_tokens check work.
If it's rendering the entire doc, it's O(n^2).

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -61,133 +73,324 @@ def __init__(self, llm: LLM, element_operator: Optional[Callable[[Element], bool
...
+        doc.properties["summary"] = doc.elements[0].properties[vars["intermediate_summary_key"]]
+        for e in doc.elements:
+            for v in vars:
+                if v in e.properties:
+                    del e.properties[v]
+        return doc
+
+    def as_llm_map(self, child: Optional[Node], **kwargs) -> Node:
+        vars = self.get_const_vars()
+        if self.fields is not None:
```
eric-anderson, line 247: self.prompt = self.prompt.fork(ignore_None=True, fields=self.fields, question=self.question, data_description=self.data_description)

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -61,133 +73,324 @@ def __init__(self, llm: LLM, element_operator: Optional[Callable[[Element], bool
...
+                child=prep_round,
+                prompt=self.prompt,
+                output_field=vars["intermediate_summary_key"],
+                llm=self.llm,
+            )
+            nodes.extend([prep_round, llm_round])
+            last = llm_round
+        cleanup = Map(child=last, f=self.cleanup)
+        nodes.append(cleanup)
+        ct = CompositeTransform(child, [])  # type: ignore
```
eric-anderson, line 267: This works but looks weird. I'd do return CompositeTransform(child, nodes=nodes) and fix the constructor.

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -61,133 +73,324 @@ def __init__(self, llm: LLM, element_operator: Optional[Callable[[Element], bool
...
+                if p in fields:
+                    continue
+                fields.append(p)
+                last = self.prompt.render_document(doc)
+                ntk = last.token_count(self.tokenizer)
+                if ntk > self.token_limit:
+                    fields.pop()
+                    return doc
+        doc.properties[vars["numel_key"]] += 1
+        this = self.prompt.render_document(doc)
```
eric-anderson, line 365: This algorithm also feels N^2.  Let's chat.

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -71,123 +80,72 @@ def summarize_data(
...
+    if summaries_as_text:
+        result_data = [ds.summarize(docset_summarizer).map(sum_to_text) for ds in result_data]
+
+    single_docs = [_docset_to_singledoc(ds) for ds in result_data]
+    agged_ds = result_data[0].context.read.document(single_docs).summarize(docset_summarizer)
+    texts = [d.properties["summary"] for d in agged_ds.take_all()]
+    return "\n".join(texts)
+
+
+def _docset_to_singledoc(ds: DocSet) -> Document:
```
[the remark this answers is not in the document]
HenryL27: lol. I used to have a more complicated implementation that did something fancy when in ray but realized it was unnecessary

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -283,7 +282,7 @@ def script(self, input_var: Optional[str] = None, output_var: Optional[str] = No
         input_str = input_var or get_var_name(self.logical_node.input_nodes()[0])
         output_str = output_var or get_var_name(self.logical_node)
         result = f"""
-prompt = LlmFilterMessagesPrompt(filter_question='{self.logical_node.question}').as_messages()
+prompt = LlmFilterMessagesJinjaPrompt.set(filter_question='{self.logical_node.question}')
```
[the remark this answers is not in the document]
HenryL27: I like fork. My initial name was `instead` but that was kinda confusing. Also wanted `with` but that's a reserved word in python so it woulda had to be something like `with_`

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -1,17 +1,24 @@
...
+)
 from sycamore.plan_nodes import NonCPUUser, NonGPUUser, Node
 from sycamore.llms import LLM
-from sycamore.llms.prompts.default_prompts import _TextSummarizerGuidancePrompt
 from sycamore.transforms.map import Map
-from sycamore.utils.time_trace import timetrace
+from sycamore.transforms.base import CompositeTransform
+from sycamore.transforms.base_llm import LLMMapElements, LLMMap
 
 NUM_DOCS_GENERATE = 60
```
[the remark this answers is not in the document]
HenryL27: yep

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -61,133 +73,324 @@ def __init__(self, llm: LLM, element_operator: Optional[Callable[[Element], bool
...
+    system=textwrap.dedent(
+        """
+        {% if question is defined %}You are a helpful research assistant. You answer questions based on
+        text you are presented with.
+        {% else %}You are a helpful data summarizer. You concisely summarize text you are presented with,
+        including as much detail as possible.
+        {% endif %}
+        """
+    ),
+    user=textwrap.dedent(
```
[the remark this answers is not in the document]
HenryL27: yes. in the static-width hierarchy implementation it wasn't bc there was just one llm map call that did all the iteration internally, and that one also used the get_text macro which had been factored out to jinja_fragments.py

lib/sycamore/sycamore/transforms/summarize.py
HenryL27: yep, turns out that n^2 for gpt-mini sized context windows is unbearably slow. fixing

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -52,14 +63,12 @@ def math_operation(val1: int, val2: int, operator: str) -> Union[int, float]:
 @context_params
 def summarize_data(
     llm: LLM,
-    question: str,
+    question: Optional[str],
     result_description: str,
     result_data: List[Any],
```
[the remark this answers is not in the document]
HenryL27: upon further inspection I apparently hallucinated that this appears anywhere else. 
changed here to data_description and input_data.

lib/sycamore/sycamore/data/document.py
```diff
@@ -328,6 +329,51 @@ def split_data_metadata(all: list[Document]) -> tuple[list[Document], list[Metad
     )
 
 
+class SummarizeDocument(Document):
```
eric-anderson, line 332: Summarized?  (Summarize implies it's actively happening).

I'd also put this in the summarize file since I think it only is meaningful for that function (and tests).

lib/sycamore/sycamore/data/document.py
```diff
@@ -328,6 +329,51 @@ def split_data_metadata(all: list[Document]) -> tuple[list[Document], list[Metad
...
+            raise ValueError("Cannot set elements directly in a SummarizeDocument")
+        super().__init__(document, **kwargs)
+        if self.data.get("sub_docs") is None:
+            self.data["sub_docs"] = []
+        elif not isinstance(sd := self.data["sub_docs"], list):
+            raise ValueError(f"sub_docs must be a list of Document, found {sd}")
+        else:
+            subdocs = self.data["sub_docs"]
+            for sd in subdocs:
+                if not isinstance(sd, (dict, UserDict, Document)):
```
eric-anderson, line 344: Is this for convenience?  I would think you could force the caller to make everything a Document.

lib/sycamore/sycamore/data/document.py
```diff
@@ -328,6 +329,51 @@ def split_data_metadata(all: list[Document]) -> tuple[list[Document], list[Metad
...
+
+    @sub_docs.setter
+    def sub_docs(self, sub_docs: list[Document]):
+        self.data["sub_docs"] = sub_docs
+
+    @sub_docs.deleter
+    def sub_docs(self) -> None:
+        self.data["sub_docs"] = []
+
+    @property
```
eric-anderson, line 360: Not sure why you have elements in here. There's a weirdness that the deleter doesn't really delete the elements since it will revert back to the elements of the sub-doc.  I would think the deleter should do self.data["_elements"] = [] as the closest approximation, but maybe the reason will be clear in later code.

lib/sycamore/sycamore/docset.py
```diff
@@ -733,10 +733,8 @@ def summarize(self, summarizer: Summarizer, **kwargs) -> "DocSet":
                     .partition(partitioner=ArynPartitioner())
                     .summarize(summarizer=summarizer)
         """
-        from sycamore.transforms import Summarize
-
-        summaries = Summarize(self.plan, summarizer=summarizer, **kwargs)
-        return DocSet(self.context, summaries)
+        map = summarizer.as_llm_map(self.plan, **kwargs)
```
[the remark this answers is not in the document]
eric-anderson: If you think this is fixable, add a TODO. If not resolve it.

lib/sycamore/sycamore/llms/prompts/prompts.py
```diff
@@ -45,8 +45,9 @@ class RenderedPrompt:
     response_format: ResponseFormat = None
 
     def token_count(self, tokenizer: Tokenizer) -> int:
-        total_text = " ".join(m.content for m in self.messages)
-        return len(tokenizer.tokenize(total_text))
+        if len(self.messages) == 0:
```
eric-anderson, line 48: If you add @cache onto the tokenize method in the tokenizer class I think it will auto-cache this so repeated calls are fine. I don't know if you still have the repeated calls.

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -5,21 +5,37 @@
...
 )
 
 log = structlog.get_logger(__name__)
+# multistep
+DEFAULT_DOCSET_SUMMARIZE_CLASS = MultiStepDocumentSummarizer  # type: ignore
+
+DEFAULT_DOCSET_SUMMARIZE_KWARGS: dict[str, Any] = {
+    "fields": "*",
+    "tokenizer": OpenAITokenizer("gpt-4o"),
+    "max_tokens": 80_000,
```
eric-anderson, line 29: can you move the token limit into the tokenizer? 80_000 is correct for 4o, but if that is changed then it would be wrong. having the class itself know the max tokens seems like a cleaner design.

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -5,21 +5,37 @@
...
+    "llm_mode": LLMMode.SYNC,
```
eric-anderson, line 30: why default to sync?

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -69,125 +82,61 @@ def summarize_data(
...
-
-        else:
-            text += str(result) + "\n"
-
-    final_summary = collapse(text, max_tokens, tokenizer, QuestionAnsweringSummarizer(llm, question))
-    return final_summary
+    single_docs = [SummarizeDocument(sub_docs=ds.take_all()) for ds in input_data]
+    agged_ds = input_data[0].context.read.document(single_docs).summarize(docset_summarizer)
+    texts = [d.properties["summary"] for d in agged_ds.take_all()]
+    return "\n".join(texts)
```
eric-anderson, line 142: I'd return the list here and do the join in the caller. That makes this function more general and makes it clear that it's summarize that's collapsing all the summaries into one block.

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -57,137 +58,375 @@ class LLMElementTextSummarizer(Summarizer):
...
+        max_tokens: token limit for each summarization. Default is 10k (default tokenizer is by character).
+        tokenizer: tokenizer to use when computing how many tokens a prompt will take. Default is
+            CharacterTokenizer
+        max_rounds: max number of rounds of heirarchical summarization to perform. The number of elements that
+            can be included in the summary is O(e^max_rounds), so can be small. Default is 4.
+    """
 
-class QuestionAnsweringSummarizer:
-    def __init__(self, llm: LLM, question: str):
+    def __init__(
```
eric-anderson, line 174: Take a look at https://github.com/aryn-ai/sycamore/commit/03f5766e0fd4680ec1b151be2019780344646639
That's roughly the structure I was expecting.  I think the fake documents makes it easier to reason about as it reduces the layers of indirection.

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -57,137 +58,375 @@ class LLMElementTextSummarizer(Summarizer):
...
+        # If fields is specified these are always included, so this will be our starting token total
+        doc.properties[vars["fields_key"]] = fields
+        curr_ntks = prompt.render_document(doc).token_count(self.tokenizer)
+        if etc:
+            for p in all_element_property_names:
+                if p in fields:
+                    continue
+                doc.properties[vars["fields_key"]] = [p]
+                ntk = prompt.render_document(doc).token_count(self.tokenizer) - data_independent_ntk
+                if curr_ntks + ntk < self.token_limit:
```
eric-anderson, line 382: Add a TODO to make this not O(n^2); I'm not sure how much it can hurt. The bad case is a lot of very small properties.

Either fix it to not be O(n^2), which I think will be a little messy since you need a "render only property and render only element controls"; or add a test case with 1000 documents, 26^2 2 character key -> 1 character value properties and a token limit of ~500k. Lets see how slow it is as to whether it's necessary to fix. If it can do that under 5s, the O(n^2) doesn't matter for now.

lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -57,137 +58,375 @@ class LLMElementTextSummarizer(Summarizer):
...
+                    return doc
+        # We added all the fields, now add as many elements as possible
+        final_numel = 0
+        doc.properties[vars["numel_key"]] = 1
+        doc.properties[vars["fields_key"]] = []
+        # This is complicated bc we might get a SummarizeDocument or a Document
+        max_numel = max(len(d.data.get("elements", [])) for d in doc.data.get("sub_docs", doc.elements))
+        # If elements can fit there's a little additional fluff added, so recompute baseline tokens
+        # with no elements (but the element introduction fluff)
+        doc.properties[vars["startel_key"]] = max_numel + 10
```
eric-anderson, line 396: Why + 10; I would think +1 is sufficient.

lib/sycamore/sycamore/data/document.py
HenryL27: will move it. It gets used when I'm about to summarize, so... maybe SummaryDocument?

lib/sycamore/sycamore/transforms/summarize.py
eric-anderson: I mis-read this. Yea, it's incremental. No need to write a test until we run into a problem for real.

lib/sycamore/sycamore/query/execution/operations.py
HenryL27: Yep, with TODO to make the param required everywhere we use tokenizers

lib/sycamore/sycamore/query/execution/operations.py
HenryL27: well, not all llms implement async, but here I've set openai explicitly so updating

lib/sycamore/sycamore/transforms/summarize.py
HenryL27: lol. not sure why I did that

lib/sycamore/sycamore/data/document.py
HenryL27: I find it useful to be able to refer to the elements of the summarize doc and set it in various places without having to worry about whether I'm a SummarizeDocument or normal document, e.g. in MultiStep element token counting.

I don't think the deleter ever gets used so maybe that can just be dropped.

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -5,21 +5,35 @@
...
 log = structlog.get_logger(__name__)
+# multistep
+DEFAULT_DOCSET_SUMMARIZE_CLASS = MultiStepDocumentSummarizer  # type: ignore
+
+DEFAULT_DOCSET_SUMMARIZE_KWARGS: dict[str, Any] = {
+    "fields": "*",
+    "tokenizer": OpenAITokenizer("gpt-4o", max_tokens=80_000),
+    "llm_mode": LLMMode.ASYNC,
+}
+# onestep
```
eric-anderson, line 31: Won't these assignments override the previous ones?
Maybe do:
```
 MULTISTEP_SUMMARIZE: Tuple[Summarizer, dict[str, Any]] = (MultiStepDocumentSummarizer, { 
    "fields": "*",
    "tokenizer": OpenAITokenizer("gpt-4o", max_tokens=80_000),
    "llm_mode": LLMMode.ASYNC,
})
ONESTEP_SUMMARIZE : Tuple[Summarizer, dict[str, Any]] = (OneStepDocumentSummarizer, { 
    "fields": [],
    "tokenizer": OpenAITokenizer("gpt-4o", max_tokens=80_000),
})

DEFAULT_SUMMARIZE = ONESTEP_SUMMARIZE
```

lib/sycamore/sycamore/query/execution/operations.py
```diff
@@ -5,21 +5,35 @@
...
+DEFAULT_DOCSET_SUMMARIZE_CLASS = OneStepDocumentSummarizer  # type: ignore
+DEFAULT_DOCSET_SUMMARIZE_KWARGS = {
+    "fields": [],
```
eric-anderson, line 34: why is this not "*"?

eric-anderson (approved):

lib/sycamore/sycamore/query/execution/operations.py
HenryL27: I changed both to use list of fields optionally terminated by EtCetera for specification. Updated to use [EtCetera] here though

PR #1195, opened 2025-02-24
Get random hits when filtering properties in sycamore query

dhruvkaliraman7:
When indexing multiple datasets with different properties in finra, sycamore client was dropping properties from the dataset that was ingested later as only the first 1000 hits in the index are used to filter properties. This PR ensures random hits are selected.


HenryL27 (approved):
Ok.
TIL there is also such thing as a `match_none` query.

PR #1196, opened 2025-02-25
Implement Pass Through using ASDK

karanataryn:
Passes through to the partitioner using the Aryn SDK, reducing maintenance work in the future.


lib/sycamore/pyproject.toml
```diff
@@ -46,6 +46,8 @@ pydantic = "^2.8.2"
 
 typing-extensions = "^4.12.2"
 
+aryn-sdk = "*"
```
HenryL27, line 49: what does * mean in this context?

HenryL27 (approved):

PR #1200, opened 2025-02-26
LLM Async mode

HenryL27:
Implements async mode for llm_map and llm_map_elements.
Also adds asynchronous llm implementations for anthropic and gemini (+exponential backoff for openai)

I left out bedrock bc bedrock async looks like "start job, poll get job until job is complete, then download result from s3 (oh by the way tell me where in s3 to put the result" and I didn't want to deal with that. (there might be better ways  - there's a converse stream client method that can give you an async iterator over the response that might do the trick but that's a whole 'nother API to deal with)


bsowell (approved):
As far as I can tell this looks fine. Is there a useful test we can write?

PR #1201, opened 2025-02-27
updating run-jupyter.sh file 

Soeb-aryn:


apps/jupyter/run-jupyter.sh
```diff
@@ -7,7 +7,13 @@ die() {
 
 mkdir -p $HOME/.jupyter
 if [[ -f /.dockerenv ]]; then
-    APP_DIR=/app
+    if [[ $(whoami) == "aryn" ]]; then
```
eric-anderson, line 10: This feels weird to need. Can we consolidate on just having APP_DIR=/app?
The only thing it seems to be used for is WORK_DIR and /app/work makes more sense anyway, although I'd probably drop that as well; no one uses the jupyter containers except in customer envs and we don't have persistent storage there anyway.

apps/jupyter/run-jupyter.sh
```diff
@@ -36,13 +42,32 @@ if [[ "${JUPYTER_CONFIG_RESET}" == yes ]]; then
 fi
 
 if [[ ! -f "${JUPYTER_CONFIG_DOCKER}" ]]; then
-    TOKEN=$(openssl rand -hex 24)
+    s3_token="s3://${JUPYTER_S3_BUCKET}/${JUPYTER_S3_PREFIX}/jupyter_token"
+    [[ ${s3_token} == "s3:///jupyter_token" ]] && s3_token=""
+    if [[ ${s3_token} != "" ]]; then
```
eric-anderson, line 47: You can consolidate this:
```
if [[ ${s3_token} != "" && ! -f "${VOLUME}/token" ]]; then
```

apps/jupyter/run-jupyter.sh
```diff
@@ -130,5 +155,6 @@ fi
 
 trap "kill $!" EXIT
 
-cd "${WORK_DIR}"
-poetry run jupyter lab "${SSLARG[@]}" --no-browser --ip 0.0.0.0 "$@"
+cd "${ARYN_DIR}"
+poetry run jupyter labextension disable "@jupyterlab/apputils-extension:announcements"
+poetry run jupyter lab "${SSLARG[@]}" --no-browser --ip 0.0.0.0 "$@"
```
eric-anderson, line 160: nit trailing newline.

eric-anderson (approved):

PR #1202, opened 2025-02-27
LLM Batch inference

HenryL27:
Adds batch inference modes for openai and anthropic.
I didn't do bedrock or gemini bc those involve dealing with s3 and gcs/bigquery.

OpenAI batch is pretty slow - to be able to test it I ended up using 3.5 turbo as it has far less demand and batch inferences are low priority (expires only after 24h). Anthropic haiku 3 was decently fast (competitive with async!) although that may be the same effect. I did not test it with a more modern / powerful claude.

### DIFF
--- a/lib/sycamore/sycamore/llms/anthropic.py
+++ b/lib/sycamore/sycamore/llms/anthropic.py
@@ -4,6 +4,8 @@ import logging
 from typing import Any, Optional, Union
 import asyncio
 import random
+from tqdm import tqdm
+import time
 
 from PIL import Image
 
@@ -203,6 +205,7 @@ class Anthropic(LLM):
         start = datetime.now()
         done = False
         retries = 0
+        response = None
         while not done:
             try:
                 response = await self._async_client.messages.create(model=self.model.value, **kwargs)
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/llms/anthropic.py
```diff
@@ -218,3 +221,48 @@ async def generate_async(self, *, prompt: RenderedPrompt, llm_kwargs: Optional[d
...
+        batch = self._client.messages.batches.create(requests=calls)
+
+        prog = tqdm(total=len(calls))
+        ctr = 0
+        while batch.processing_status == "in_progress":
+            batch = self._client.messages.batches.retrieve(batch.id)
+            counts = batch.request_counts
+            diff = counts.succeeded + counts.errored + counts.canceled + counts.expired - ctr
+            prog.update(n=diff)
+            ctr += diff
```
karanataryn, line 252: What is this counter indicate? A running total of counts across all batches. Would be useful to have a comment here.

lib/sycamore/sycamore/llms/anthropic.py
```diff
@@ -218,3 +221,48 @@ async def generate_async(self, *, prompt: RenderedPrompt, llm_kwargs: Optional[d
...
+            if batch.processing_status == "ended":
+                break
+            time.sleep(10)
```
karanataryn, line 255: Might want to make this configurable?

lib/sycamore/sycamore/llms/openai.py
```diff
@@ -488,3 +493,56 @@ async def _generate_awaitable_using_openai_structured(
...
+        f = io.BytesIO()
+        for i, c in enumerate(calls):
+            f.write(json.dumps(c).encode("utf-8"))
+            if i != len(calls) - 1:
+                f.write(b"\n")
+        client = self.client_wrapper.get_client()
+        starttime = datetime.now()
+        batch_in_file = client.files.create(file=f, purpose="batch")
+        batch = client.batches.create(
+            input_file_id=batch_in_file.id, endpoint="/v1/chat/completions", completion_window="24h"
```
karanataryn, line 517: Also make configurable?

lib/sycamore/sycamore/transforms/base_llm.py
```diff
@@ -37,7 +37,7 @@ def _infer_prompts(
             res[i] = rs
         return res
     elif llm_mode == LLMMode.BATCH:
-        raise NotImplementedError("Haven't done batch yet")
+        return llm.generate_batch(prompts=prompts)
```
karanataryn, line 40: Yay!

karanataryn (approved):
LGTM with few suggestions. Can we add a unit test of some sort here? Integration tests would be a bit tough.

lib/sycamore/sycamore/llms/anthropic.py
HenryL27: probably the right thing to do is just drop the progress bars altogether

lib/sycamore/sycamore/llms/anthropic.py
HenryL27: yep

PR #1203, opened 2025-02-28
Add param to control how model is selected in hybrid table extractor

HenryL27:
add a model_selection param to hybrid table extractor that accepts expressions in this spec:
(take from the docstring)
```
        Parse a model selection expression. Model selection expressions are of the form:
            "metric cmp threshold -> model; metric cmp threshold -> model; model;"
        That is, any number of conditional expression selections followed by up to one unconditional
        selection expression, separated by semicolons. Expressions are processed from left to right.

        - Supported metrics are "pixels" - the number of pixels in the larger dimension of the table, and
        "chars" - the number of characters in the table, as detected by the partitioner's text_extractor.
        - Supported comparisons are the usual set - <, >, <=, >=, ==, !=.
        - The threshold must be numeric (and int or a float)
        - The model must be either "deformable_detr" or "table_transformer"

        Args:
            selection: the selection string.

        Returns:
            a function that can be used to select a model given the pixels and chars metrics.
```

### DIFF
--- a/lib/aryn-sdk/aryn_sdk/partition/partition.py
+++ b/lib/aryn-sdk/aryn_sdk/partition/partition.py
@@ -85,11 +85,12 @@ def partition_file(
             default: English
         extract_table_structure: extract tables and their structural content.
             default: False
-        table_extraction_options: Specify options for table extraction, currently only supports boolean
-            'include_additional_text': if table extraction is enabled, attempt to enhance the table
-            structure by merging in tokens from text extraction. This can be useful for tables with missing
-            or misaligned text, and is False by default.
-            default: {}
+        table_extraction_options: Specify options for table extraction. Only enabled if table extraction
+            is enabled. Default is {}. Options:
+            - 'include_additional_text': Attempt to enhance the table structure by merging in tokens from
+                text extraction. This can be useful for tables with missing or misaligned text. Default: False
+            - 'model_selection': expression to instruct DocParse how to choose which model to use for table
+                extraction. See https://docs.aryn.ai/docparse/processing_options for more details.
         extract_images: extract image contents in ppm format, base64 encoded.
             default: False
         selected_pages: list of individual pages (1-indexed) from the pdf to partition
--- a/lib/aryn-sdk/aryn_sdk/partition/partition.py
+++ b/lib/aryn-sdk/aryn_sdk/partition/partition.py
@@ -90,7 +90,8 @@ def partition_file(
             - 'include_additional_text': Attempt to enhance the table structure by merging in tokens from
                 text extraction. This can be useful for tables with missing or misaligned text. Default: False
             - 'model_selection': expression to instruct DocParse how to choose which model to use for table
-                extraction. See https://docs.aryn.ai/docparse/processing_options for more details.
+                extraction. See https://docs.aryn.ai/docparse/processing_options for more details. Default:
+                "pixels > 500 -> deformable_detr; table_transformer"
         extract_images: extract image contents in ppm format, base64 encoded.
             default: False
         selected_pages: list of individual pages (1-indexed) from the pdf to partition
--- a/lib/sycamore/sycamore/transforms/table_structure/extract.py
+++ b/lib/sycamore/sycamore/transforms/table_structure/extract.py
@@ -239,6 +239,14 @@ class HybridTableStructureExtractor(TableStructureExtractor):
 
     _model_names = ("table_transformer", "deformable_detr")
     _metrics = ("pixels", "chars")
+    _comparisons = (
+        "==",
+        "<=",
+        ">=",
+        "!=",
+        "<",
+        ">",
+    )
 
     def __init__(
         self,
@@ -255,8 +263,8 @@ class HybridTableStructureExtractor(TableStructureExtractor):
         doc_image: Image.Image,
         model_selection: str,
     ) -> Union[TableTransformerStructureExtractor, DeformableTableStructureExtractor]:
-        """If the absolute size of the table is > 500 pixels in any dimension, use deformable.
-        Otherwise, use TATR"""
+        """Use the model_selection expression to choose the model to use for table extraction.
+        If the expression returns None, use table transformer."""
         if element.bbox is None:
             return self._tatr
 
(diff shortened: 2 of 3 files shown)


lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -85,11 +85,12 @@ def partition_file(
             default: English
         extract_table_structure: extract tables and their structural content.
             default: False
-        table_extraction_options: Specify options for table extraction, currently only supports boolean
-            'include_additional_text': if table extraction is enabled, attempt to enhance the table
-            structure by merging in tokens from text extraction. This can be useful for tables with missing
-            or misaligned text, and is False by default.
-            default: {}
+        table_extraction_options: Specify options for table extraction. Only enabled if table extraction
+            is enabled. Default is {}. Options:
```
karanataryn, line 89: Looks like these are on different lines?

lib/aryn-sdk/aryn_sdk/partition/partition.py
```diff
@@ -85,11 +85,12 @@ def partition_file(
...
+            - 'include_additional_text': Attempt to enhance the table structure by merging in tokens from
+                text extraction. This can be useful for tables with missing or misaligned text. Default: False
+            - 'model_selection': expression to instruct DocParse how to choose which model to use for table
+                extraction. See https://docs.aryn.ai/docparse/processing_options for more details.
```
karanataryn, line 93: No default? Also different lines

lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -245,25 +250,47 @@ def __init__(
...
+
+        select_fn = self.parse_model_selection(model_selection)
+
         width, height = doc_image.size
         bb = element.bbox.to_absolute(width, height)
         padding = 10
         max_dim = max(bb.width, bb.height) + 2 * padding
-        if max_dim > 500:
+
+        nchars = len("".join(tok["text"] for tok in element.tokens or [{"text": ""}]))
```
karanataryn, line 270: We don't need to calculate this if `chars` is not in `model_selection` right? Might be worth it as an optimization given this is expensive. Could do the same for `max_dim` but that's less urgent.

lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -245,25 +250,47 @@ def __init__(
...
+
+        selection = select_fn(max_dim, nchars)
+        print("=" * 80)
+        print(selection)
+        if selection == "table_transformer":
+            return self._tatr
+        elif selection == "deformable_detr":
             return self._deformable
-        return self._tatr
+        elif selection is None:
```
karanataryn, line 279: I think we should state in the docstring and docs that if the value doesn't meet any of the specified constraints, it defaults to `tatr`.

lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -276,9 +303,141 @@ def extract(self, element: TableElement, doc_image: Image.Image, union_tokens=Fa
...
+    def parse_model_selection(cls, selection: str) -> Callable[[float, int], Optional[str]]:
         """
-        model = self._pick_model(element, doc_image)
-        return model.extract(element, doc_image, union_tokens)
+        Parse a model selection expression. Model selection expressions are of the form:
+            "metric cmp threshold -> model; metric cmp threshold -> model; model;"
+        That is, any number of conditional expression selections followed by up to one unconditional
+        selection expression, separated by semicolons. Expressions are processed from left to right.
+
+        - Supported metrics are "pixels" - the number of pixels in the larger dimension of the table, and
```
karanataryn, line 320: Why not total number of pixels? We should explain the intuition behind this choice.

karanataryn (commented):
Comments and suggestions. I like the query syntax, but wonder if this is a bit too complex for a relatively obscure option. We'll need to document this well.

lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -276,9 +303,139 @@ def extract(self, element: TableElement, doc_image: Image.Image, union_tokens=Fa
...
+                if (rv := c(pixels, chars)) is not None:
+                    return rv
+            return None
+
+        return select_fn
+
+    @classmethod
+    def parse_comparison(cls, comparison: str) -> tuple[str, Callable[[Num, Num], bool], Union[int, float]]:
+        cmp_pieces = []
+        if "!=" in comparison:
```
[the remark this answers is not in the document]
HenryL27: I thought about that but I want to avoid sticking evals in, especially when the user is writing the doce here. I think it's probably safe but I'd rather not open the door for potential RCEs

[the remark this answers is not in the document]
karanataryn: This code only gets executed if we find that comparator in the string, so I think it's safe? Alternatively, how about something like this:

```
def apply_operator(op_string, a, b):
    operators = {
        "==": lambda a, b: a == b,
        "!=": lambda a, b: a != b,
        "<": lambda a, b: a < b,
        "<=": lambda a, b: a <= b,
        ">": lambda a, b: a > b,
        ">=": lambda a, b: a >= b,
    }
    
    if op_string in operators:
        return operators[op_string](a, b)
    else:
        raise ValueError(f"Unsupported operator: {op_string}")
```

lib/sycamore/sycamore/transforms/table_structure/extract.py
HenryL27: Are these expensive? esp. given I'm about to plug all this stuff into a neural network and then do a superlinear computation on them? I guess I can at lease compute nchars as a sum of lens rather than a len of sums but I'm unconvinced that the bit of performance from (doing a linear scan for 'chars' btw and) maybe not computing this is worth the complexity of providing dummy values into the selection call or modifying the signature of it

karanataryn: Ah, I was simply imagining checking if `chars` is in the `model_selection` string. That feels like a good approximation to me. Agree that it's not worth much more effort.

lib/sycamore/sycamore/transforms/table_structure/extract.py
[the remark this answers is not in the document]
HenryL27: I like that. I'm taking it

karanataryn (approved):
LGTM!

PR #1204, opened 2025-02-28
Add entity name in grouped result, also add materialize in groupbycount operator

bohou-aryn:


lib/sycamore/sycamore/docset.py
```diff
@@ -1338,10 +1338,10 @@ def llm_query(self, query_agent: LLMTextQueryAgent, **kwargs) -> "DocSet":
         queries = LLMQuery(self.plan, query_agent=query_agent, **kwargs)
         return DocSet(self.context, queries)
 
-    def groupby(self, key: Union[str, list[str]]) -> "GroupedData":
+    def groupby(self, grouped_key: Union[str, list[str]], entity: str=None) -> "GroupedData":
```
HenryL27, line 1341: What does this entity do?

lib/sycamore/sycamore/grouped_data.py
```diff
@@ -1,41 +1,42 @@
...
-        aggregated = grouped.aggregate(f)
+        def group_udf(batch):
+            import numpy as np
+            result = {"count": np.array([len(batch["properties"])])}
+            if self._entity:
+                result[self._entity] = np.array([batch["properties"][0][self._entity]])
+            return result
+
+        grouped = dataset.filter(filter_meta).map(Document.from_row).groupby(self._grouped_key)
+        aggregated = grouped.map_groups(group_udf)
```
HenryL27, line 26: what happens if a group is bigger than the batch size?

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -649,10 +649,14 @@ def execute(self) -> Any:
...
-        docset = self.inputs[0].embed(embedder)
+        import tempfile
+        temp_dir = tempfile.mkdtemp()
+
+        docset = self.inputs[0].embed(embedder).materialize(path=f"{temp_dir}", source_mode=MATERIALIZE_USE_STORED)
+        docset.execute()
         centroids = docset.kmeans(K=K * 2, field_name=embed_name)
         clustered = docset.clustering(centroids=centroids, cluster_field_name=cluster_field_name, field_name=embed_name)
-        result = clustered.groupby(cluster_field_name).count().sort(descending, "properties.count", 0).limit(K)
+        result = clustered.groupby(cluster_field_name, entity_name.split(".")[-1]).count().sort(descending, "properties.count", 0).limit(K)
```
HenryL27, line 659: why the split and take last segment of key?

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -649,10 +649,14 @@ def execute(self) -> Any:
         descending = logical_node.descending
         K = logical_node.K
 
-        docset = self.inputs[0].embed(embedder)
+        import tempfile
+        temp_dir = tempfile.mkdtemp()
+
+        docset = self.inputs[0].embed(embedder).materialize(path=f"{temp_dir}", source_mode=MATERIALIZE_USE_STORED)
```
HenryL27, line 655: One of the trickinesses I faced using this the other day was getting the embedder to embed the field I wanted to cluster by - I had to sneak in a pre_process_document hook to the embedder object and also temporarily hide the element text representations and stick in a dummy document text representation to get it to do the right thing. Is that solved?

lib/sycamore/sycamore/grouped_data.py
```diff
@@ -1,41 +1,42 @@
...
             return not isinstance(doc, MetadataDocument)
 
-        grouped = dataset.filter(filter_meta).map(Document.from_row).groupby(self._key)
-        aggregated = grouped.aggregate(f)
+        def group_udf(batch):
+            import numpy as np
+            result = {"count": np.array([len(batch["properties"])])}
+            if self._entity:
+                result[self._entity] = np.array([batch["properties"][0][self._entity]])
+            return result
```
HenryL27, line 23: This is putting the first instance of the entity in the result? Why is this useful?

HenryL27 (commented):
What is this trying to do?

lib/sycamore/sycamore/docset.py
bohou-aryn: entity name, need map back to the field to do the embedding, otherwise, the result is only the count and cluster id, since our groupby is indirect on cluster id.

lib/sycamore/sycamore/grouped_data.py
bohou-aryn: That would overflow.

lib/sycamore/sycamore/query/execution/sycamore_operator.py
bohou-aryn: our entity name is inside properties.

lib/sycamore/sycamore/query/execution/sycamore_operator.py
bohou-aryn: yes, that is mitigated in the embedder, you could point to some other field to embed instead of the text_representation.

lib/sycamore/sycamore/grouped_data.py
bohou-aryn: for each group, choose one entity in that group.

lib/sycamore/sycamore/query/execution/sycamore_operator.py
HenryL27: so entity_name is something like "properties.field"? What about "properties.nested.field"?

HenryL27 (approved):

PR #1208, opened 2025-03-04
Bump Dependencies to Fix Security Issues

karanataryn:


lib/aryn-sdk/poetry.lock
```diff
@@ -141,92 +128,118 @@ all = ["flake8 (>=7.1.1)", "mypy (>=1.11.2)", "pytest (>=8.3.2)", "ruff (>=0.6.2
 
 [[package]]
 name = "numpy"
-version = "1.26.4"
+version = "2.0.2"
```
HenryL27, line 131: At some point I had issues (using uv) with installing sycamore in a venv with numpy > 2. we good now?

lib/aryn-sdk/poetry.lock
```diff
@@ -475,41 +489,42 @@ use-chardet-on-py3 = ["chardet (>=3.0.2,<6)"]
...
-    {file = "tzdata-2024.2-py2.py3-none-any.whl", hash = "sha256:a48093786cdcde33cad18c2555e8532f34422074448fbc874186f0abd79565cd"},
-    {file = "tzdata-2024.2.tar.gz", hash = "sha256:7d85cc416e9382e69095b7bdf4afd9e3880418a2413feec7069d533d6b4e31cc"},
+    {file = "tzdata-2025.1-py2.py3-none-any.whl", hash = "sha256:7e127113816800496f027041c570f50bcd464a020098a3b6b199517772303639"},
+    {file = "tzdata-2025.1.tar.gz", hash = "sha256:24894909e88cdb28bd1636c6887801df64cb485bd593f2fd83ef29075a81d694"},
 ]
 
 [[package]]
 name = "urllib3"
-version = "1.26.20"
+version = "2.3.0"
```
HenryL27, line 514: another major version update, flagging but prob fine

lib/poetry-lock/pyproject.toml
```diff
@@ -7,7 +7,7 @@ readme = "README.md"
 repository = "https://github.com/aryn-ai/sycamore.git"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.13"
+python = ">=3.9.2,<3.13"
```
HenryL27, line 10: huh?

lib/sycamore/poetry.lock
```diff
@@ -2171,63 +2268,74 @@ mccabe = ">=0.6.0,<0.7.0"
 pycodestyle = ">=2.8.0,<2.9.0"
 pyflakes = ">=2.4.0,<2.5.0"
 
+[[package]]
+name = "flatbuffers"
```
HenryL27, line 2272: what pulled this in?

lib/sycamore/poetry.lock
```diff
@@ -3843,6 +4052,36 @@ files = [
 [package.dependencies]
 six = "*"
 
+[[package]]
+name = "layoutparser"
```
HenryL27, line 4056: what's this for?

HenryL27 (approved):
1 - adding onnxruntime could be big overhead. prob fine given all our other deps
2 - can we really drop support for python 3.9.1 and 3.9.0?
3 - flagged a couple major version upgrades in case you missed them

lib/poetry-lock/pyproject.toml
karanataryn: This is due to cryptography: https://github.com/aryn-ai/sycamore/security/dependabot/327

lib/sycamore/poetry.lock
karanataryn: `onnxruntime` :(

lib/sycamore/poetry.lock
karanataryn: Required by `unstructured-inference`

lib/aryn-sdk/poetry.lock
karanataryn: This is only for `aryn-sdk`, so I think it should be fine.

karanataryn:
> 1 - adding onnxruntime could be big overhead. prob fine given all our other deps 2 - can we really drop support for python 3.9.1 and 3.9.0? 3 - flagged a couple major version upgrades in case you missed them

1 is due to continued support for unstructured, 2 is due to the newest versions of cryptography not allowing for 3.9.0 and 3.9.1.

lib/aryn-sdk/poetry.lock
HenryL27: well we just made sycamore depend on asdk right?

bsowell (approved):
I think I'm good with this. I think we should try to get rid of Unstructured, but should do that separately.

lib/aryn-sdk/poetry.lock
karanataryn: Perhaps this might be an issue haha.

PR #1216, opened 2025-03-06
Rewrite llm query operator

bohou-aryn:
This is part of the series of tasks for converting llm operator to use Henry's new LLMMap or LLMElementMap.


lib/sycamore/sycamore/llms/prompts/prompts.py
```diff
@@ -662,3 +662,54 @@ def render_element(self, elt: Element, doc: Document) -> RenderedPrompt:
...
+        render_args = copy.deepcopy(self.kwargs)
+        render_args["elt"] = elt
+        render_args["doc"] = doc
+        render_args["prev"] = prev
+
+        result = render_templates(self._sys_template, self._user_templates, render_args)
+        if self.include_image and len(result.messages) > 0:
+            from sycamore.utils.pdf_utils import get_element_image
+
+            result.messages[-1].images = [get_element_image(elt, doc)]
```
HenryL27, line 712: might also want to add the image of prev here?

lib/sycamore/sycamore/transforms/base_llm.py
```diff
@@ -207,6 +209,14 @@ def llm_map_elements(self, documents: list[Document]) -> list[Document]:
             for e, _ in elt_doc_pairs:
                 e.properties[self._iteration_var] = 0
 
+        skips = []
+        counter = 0
+        for e, _ in elt_doc_pairs:
+            if self._filter(e) and (not self._number_of_elements or counter < self._number_of_elements):
+                counter += 1
+            else:
+                skips.append(e)
```
HenryL27, line 218: skips should be an array of bools (with length = num of elements) if the rest of its usage remains the same

HenryL27 (commented):
Thanks! Now can you delete execute_query / _query_text_object (assuming this thing is equivalent)?

lib/sycamore/sycamore/transforms/base_llm.py
```diff
@@ -207,6 +209,14 @@ def llm_map_elements(self, documents: list[Document]) -> list[Document]:
             for e, _ in elt_doc_pairs:
                 e.properties[self._iteration_var] = 0
 
+        skips = []
+        counter = 0
+        for e, _ in elt_doc_pairs:
+            if self._filter(e) and (not self._number_of_elements or counter < self._number_of_elements):
```
dhruvkaliraman7, line 215: Why are we filtering twice when I want to run Table Merger? Once here for table and then again in render_elements.

lib/sycamore/sycamore/llms/prompts/prompts.py
```diff
@@ -662,3 +662,54 @@ def render_element(self, elt: Element, doc: Document) -> RenderedPrompt:
...
+        self.user = user
+        self.include_image = include_image
+        self.response_format = response_format
+        self.kwargs = kwargs
+        self._env = SandboxedEnvironment(extensions=["jinja2.ext.loopcontrols"])
+        self._sys_template: Optional[Template] = None
+        self._user_templates: Union[None, list[Template]] = None
+
+    def render_element(self, elt: Element, doc: Document) -> RenderedPrompt:
+        filtered = [e for e in doc.elements if e.type == "table"]
```
dhruvkaliraman7, line 691: Am I understanding this wrong or for every document we have O(len(doc.elements)^2)?

lib/sycamore/sycamore/transforms/base_llm.py
bohou-aryn: For table merger, I think mostly, the prompt side does not have the filter or skip information that's collected in the llm_map_elements.

dhruvkaliraman7 (approved):
Only transform where you care about index of prev filtered element. Can revisit later if processing is too slow.

lib/sycamore/sycamore/llms/prompts/prompts.py
dhruvkaliraman7: @HenryL27 no wiggle room in render_elements for this?

HenryL27: I think the alternative is to do a single pass beforehand and write down a pointer on each element to prev. 
then in render elements you're O(1) right?

HenryL27: You might be able to avoid needing this particular class entirely since you can do that lookup in Jinja - something like
```python
"""
{%- set prev = doc.elements[elt.properties["_prev_table"]] if "_prev_table" in elt.properties else None -%}
"""
```

PR #1217, opened 2025-03-07
Force all `List-item` content onto one line

MarkLindblad:
While this makes partitioned output containing `List-item`s more understandable to LLMs, it may not be desirable to change the output so significantly as this does.

### DIFF
--- a/lib/sycamore/sycamore/transforms/detr_partitioner.py
+++ b/lib/sycamore/sycamore/transforms/detr_partitioner.py
@@ -135,10 +135,10 @@ class ArynPDFPartitioner:
                 for m_index, m in enumerate(matched):
                     matches.append(m)
                     if text_to_add := m.text_representation:
-                        if (
-                            is_list_item and m_index + 1 < num_matched and text_to_add[-1] == "\n"
-                        ):  # special case for list items
-                            text_to_add = text_to_add[:-1]
+                        if is_list_item:  # special case for list items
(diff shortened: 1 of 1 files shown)


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -135,10 +135,10 @@ def _supplement_text(inferred: list[Element], text: list[Element], threshold: fl
                 for m_index, m in enumerate(matched):
                     matches.append(m)
                     if text_to_add := m.text_representation:
-                        if (
-                            is_list_item and m_index + 1 < num_matched and text_to_add[-1] == "\n"
-                        ):  # special case for list items
-                            text_to_add = text_to_add[:-1]
+                        if is_list_item:  # special case for list items
+                            text_to_add = text_to_add.rstrip().lstrip().replace("\n", " ")
```
HenryL27, line 139: string.lstrip().rstrip() == string.strip()

HenryL27 (approved):
now we might be in the territory of 'add a flag and make this opt-in/out' though idk, this seems reasonable to me

lib/sycamore/sycamore/transforms/detr_partitioner.py
MarkLindblad: Done.

MarkLindblad:
After discussion, we have decided to not make this change now and to possibly add it behind an opt-in flag in the future.

PR #1219, opened 2025-03-07
[llm unify 8/8] llm documentation

HenryL27:


docs/source/sycamore/APIs/llm.rst
```diff
@@ -3,6 +3,20 @@
 LLM
 ================
 
-.. automodule:: sycamore.llms.llms.LLM
+.. autoclass:: sycamore.llms.llms.LLM
    :members:
-   
\ No newline at end of file
+   :exclude-members: generate_old
```
karanataryn, line 8: Can we not delete the old code and docstrings?

docs/source/conf.py
```diff
@@ -26,6 +26,8 @@
     "sphinx.ext.autodoc.typehints",
     "sphinx.ext.autosummary",
     "sphinx.ext.viewcode",
+    "enum_tools.autoenum",
```
karanataryn, line 29: Where we using these two new libraries?

HenryL27: Autoenum is used for documenting enums (e.g. LLMMode shows LLMMode.SYNC, LLMMode.ASYNC, LLMMode.BATCH - without it it's just a subclass of Enum). I use autovariables to document the ResponseFormat type variable.

docs/source/sycamore/APIs/llm.rst
HenryL27: generate_old isn't completely excised from the library so I don't want to drop it yet.

karanataryn (approved):
LGTM!

PR #1220, opened 2025-03-12
Markdown conversion now takes options for header/footer/pagenum

alexaryn:


lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -73,10 +85,13 @@ def elements_to_markdown(elems: list[Element]) -> str:
                 sio.write(f"{indent}-{t}\n")
             else:
                 sio.write(f"- {text}\n")
-        elif type in ("caption", "footnote"):
+        elif type in {"caption", "footnote", "page-header", "page-footer"}:
             sio.write(f"\n{text}\n\n")
```
karanataryn, line 89: Why do we specifically need two newline characters for these types?

lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -33,17 +32,30 @@ def escape_str(s: str) -> str:
     return sio.getvalue()
 
 
-def elements_to_markdown(elems: list[Element]) -> str:
+def elements_to_markdown(elems: list[Element], opts: Optional[dict[str, Any]] = None) -> str:
     """
     This is the main function of interest.
     Assumes elements are sorted as per bbox_sort.bbox_sort_document().
     """
+    skip_types = {"image"}
```
karanataryn, line 40: We don't need to right now, but would be worth thinking about returning the image in an encoded format if `extract_images=True` and adding the b64 version of it: https://stackoverflow.com/questions/50817518/hard-code-markdown-images

lib/sycamore/sycamore/utils/markdown.py
```diff
@@ -73,10 +85,13 @@ def elements_to_markdown(elems: list[Element]) -> str:
                 sio.write(f"{indent}-{t}\n")
             else:
                 sio.write(f"- {text}\n")
-        elif type in ("caption", "footnote"):
+        elif type in {"caption", "footnote", "page-header", "page-footer"}:
             sio.write(f"\n{text}\n\n")
         else:
             sio.write(f"\n{text}\n")
+        blank = False
+    if (pagenum is not None) and (not blank):
```
karanataryn, line 93: Shouldn't this be before the for loop? Looks like we add page numbers at the beginning of the page.

lib/sycamore/sycamore/utils/markdown.py
alexaryn: It's so they don't get folded in as part of a paragraph when rendered in a browser.

lib/sycamore/sycamore/utils/markdown.py
alexaryn: Officially Markdown does not support inline images.  This is some quasi-feature.  We can revisit, but I don't see it being useful at the moment.

lib/sycamore/sycamore/utils/markdown.py
alexaryn: This is to make sure the final page number gets omitted.  We're emitting page numbers at the end of pages, when we see the page number change to the next one.

karanataryn: Ah, I see.

karanataryn (approved):

PR #1230, opened 2025-03-24
Add extract_image_format option.

bsowell:
Currently this can be set to either PPM and PNG, and controls the format an image is encoded by the partitioner when extract_images is True. The default is PPM, which is the current behavior, but PPMs are not displayable on the console, so we provide an alternative option.

### DIFF
--- a/lib/sycamore/sycamore/transforms/detr_partitioner.py
+++ b/lib/sycamore/sycamore/transforms/detr_partitioner.py
@@ -159,6 +159,7 @@ class ArynPDFPartitioner:
         table_structure_extractor=None,
         table_extractor_options: dict = {},
         extract_images=False,
+        extract_image_format: Literal["PPM", "PNG"] = "PPM",
         batch_size: int = 1,
         use_partitioning_service=True,
         aryn_api_key: str = "",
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -316,6 +318,7 @@ def _partition_pdf_batched(
         table_structure_extractor=None,
         table_extractor_options: dict = {},
         extract_images: bool = False,
+        extract_image_format: Literal["PPM", "PNG"] = "PPM",
```
karanataryn, line 321: Can't we support any PIL `image.save` accepted format here?

karanataryn (approved):
Quick question, but not blocking.

lib/sycamore/sycamore/transforms/detr_partitioner.py
bsowell: Yeah, I suppose so. For the managed service API I would probably be in favor of restricting it more so we can provide better error messages/etc, but that's less of a concern here.

PR #1236, opened 2025-03-25
Switch to Base64 Encoding

karanataryn:


karanataryn:
Closed since we don't want to change defaults right now.

PR #1237, opened 2025-03-26
Synth data gen

dtecuci:
Added synthetic data generation


examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
...
+                    },
+                    "type": "knn_vector",
+                }
+            }
+        },
+    }
+}
+
+# The number of documents to generate.
+max_docs = int(args.maxdocs)
```
eric-anderson, line 70: call this num_docs in the arguments then (and be consistent on the usage of _, both in the argument and var or neither)

examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
...
+                        "parameters": {},
+                    },
+                    "type": "knn_vector",
+                }
+            }
+        },
+    }
+}
+
+# The number of documents to generate.
```
eric-anderson, line 69: put this in the help string for the argument.

examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
...
+max_docs = int(args.maxdocs)
+
+# The maximum number of elements in each document.
+max_elems = int(args.maxelems)
+
+
+def make_element(text):
```
eric-anderson, line 76: I'd remove the docstring and just call this element_from_text. If you keep the docstring for something this simple, I expect it for all the functions.

examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
...
+
+    Args:
+        text (str): The text representation of the element.
+
+    Returns:
+        dict: A dictionary representing the element with the following keys:
+            - 'properties': A dictionary of properties associated with the element.
+            - 'text_representation': The text representation of the element.
+    """
+#    return Element(text_representation=text)
```
eric-anderson, line 88: shouldn't this return an Element?
Prefer to add type-checks to the function definition so mypy will catch this for you.

examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
...
+        list: A list of Document objects, each containing a random set of elements.
+    """
+    docset = []
+    for i in range(max_docs):
+        doc = generate_doc(str(i), keywords, probabilities, max_elems)
+        docset.append(doc)
+    return docset
+
+
+keywords = ['cat', 'dog']
```
eric-anderson, line 151: Weird to have this as two parallel arrays. It makes it hard to be sure that the counts are the same and which keyword goes with which name. 
Make this a dictionary.
```
keyword_to_probability = {
   'cat": 0.1,
   'dog': 0.2,
}
```

examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
...
+     Nullam auctor magna nec mollis ultricies. Mauris non pulvinar turpis. Pellentesque dictum 
+     tortor ac elit feugiat varius. Nulla sollicitudin, lectus non venenatis finibus, nulla ex 
+     tempor urna, id ornare est nisl et lacus. Pellentesque commodo turpis in nibh aliquet viverra. 
+     Curabitur varius, elit nec finibus scelerisque, nibh eros congue dui, et ultrices justo magna 
+     eu sem. Sed ultricies posuere posuere. Ut euismod neque at diam interdum, sed rhoncus quam 
+     iaculis. Morbi sit amet tempus orci, a dictum metus. Donec a imperdiet sem. Vestibulum tincidunt 
+     gravida turpis vitae aliquet.
+    """
+    for i in range(num):
+        text = "base_text" + str(i)
```
eric-anderson, line 107:
```
# choose keywords to include based on the probability of each keyword
```

examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
...
+        selected_keywords = ""
+        for keyword, probability in zip(keywords, probabilities):
+            if random.random() < probability:
+                selected_keywords = selected_keywords + " " + keyword
```
eric-anderson, line 111: fwiw, this is O(n); Probably doesn't matter, but the cleaner way would be to do it as a list. I think:
```
   " ".join([k for k in keyword_to_probability if random.random() < keyword_to_probability[k]])
```
does it in a single list comprehension.

examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
...
+probabilities = [0.1, 0.2]
+docset = generate_docset(keywords, probabilities, max_docs, max_elems)
+
+print(f"Generated docset containing {len(docset)} documents")
+print(docset[0])
+
+context = sycamore.init()
+tokenizer = HuggingFaceTokenizer("thenlper/gte-small")
+llm = OpenAI(OpenAIModels.GPT_3_5_TURBO.value)
+
```
eric-anderson, line 161: up to you, but I'd do the entire pipeline as a single statement:
```
(
  sycamore.init()
  .read.document(docs=docset)
  .explode()
  .write.opensearch(
      ...
  )
)
```

   .

eric-anderson (approved):

examples/query/synth_data_llm_filter.py
```diff
@@ -0,0 +1,169 @@
+import sycamore
```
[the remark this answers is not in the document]
dtecuci: Fixed.

PR #1238, opened 2025-03-26
deserialize a summaryDocument as a SummaryDocument, not a Document

HenryL27:
Managed to run into this issue with the multistep summarizer too. The real fix is to deserialize it correctly rather than cast in place I guess


dhruvkaliraman7 (approved):
nice

PR #1241, opened 2025-03-27
Add access method for materialize docset

bohou-aryn:


lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -121,6 +121,40 @@ def script(self, input_var: Optional[str] = None, output_var: Optional[str] = No
         )
 
 
+class SycamoreQueryBookmark(SycamoreOperator):
+    """
+    Note: Currently only supports an OpenSearch scan load implementation.
```
baitsguy, line 126: Fix to s3

baitsguy (approved):

PR #1242, opened 2025-03-27
Fix limit and add subquerying

dhruvkaliraman7:
1. Fix limit to not count Metadata Documents.
2. Add hacky sub querying for a certain set of queries where you want to limit using a field on unique values rather than each row.

### DIFF
--- a/lib/sycamore/sycamore/transforms/basics.py
+++ b/lib/sycamore/sycamore/transforms/basics.py
@@ -1,8 +1,9 @@
-from typing import Callable, TYPE_CHECKING
+from typing import Callable, TYPE_CHECKING, Optional
 
-from sycamore.data import Document
+from sycamore.data import Document, MetadataDocument
 from sycamore.plan_nodes import Node, NonGPUUser, NonCPUUser, Transform
 from sycamore.transforms.map import MapBatch
+from sycamore.utils.nested import dotted_lookup
 
 if TYPE_CHECKING:
     from ray.data import Dataset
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/transforms/basics.py
```diff
@@ -24,16 +25,59 @@ class Limit(NonCPUUser, NonGPUUser, Transform):
...
+        self._field = field
 
     def execute(self, **kwargs) -> "Dataset":
+        import ray
+
         dataset = self.child().execute()
-        return dataset.limit(self._limit)
+        rayDocs = []
+        if self._field:
+            unique_docs = set()
```
HenryL27, line 39: stupid naming nit: unique_field_values

lib/sycamore/sycamore/transforms/basics.py
```diff
@@ -24,16 +25,59 @@ class Limit(NonCPUUser, NonGPUUser, Transform):
...
+                    unique_docs.add(field_val)
+                    if len(unique_docs) > self._limit:
+                        break
+                rayDocs.append(doc)
+            else:
+                if not isinstance(deser_doc, MetadataDocument):
+                    count += 1
+                    if count > self._limit:
+                        break
+                rayDocs.append(doc)
```
HenryL27, line 55: I'm worried that this logic might not be right.
Case 1: field is set to something, expected behavior is to take 5 documents with different values. 
What I get: 7 documents of which 3 have the same value. But in my docset there are hundreds of duplicates of each value. (bc the 8th document gave me my 6th unique field)

Case 2: field is set to something, expected behavior is to take all documents with one of 5 different values.
Same issue but the other way around

Case 3: field is unset, expected behavior is to get 5 documents
What I get: 5 documents, correctly, but a random subset of the metadata documents. Probably I want either all the metadata docs or just the ones pertaining to the documents I have selected.

lib/sycamore/sycamore/transforms/basics.py
```diff
@@ -24,16 +25,59 @@ class Limit(NonCPUUser, NonGPUUser, Transform):
...
+                    unique_docs.add(field_val)
+                    if len(unique_docs) > self._limit:
+                        break
+                filtered_docs.append(doc)
+            else:
+                if not isinstance(doc, MetadataDocument):
+                    count += 1
+                    if count > self._limit:
+                        break
+                filtered_docs.append(doc)
```
HenryL27, line 78: might be able to factor out the inner logic like so
```python
def limit_iterable(doc_iterable):
    for doc in doc_iterable:
        ...

def execute() -> "Dataset":
    doc_iter = map(dataset.iter_rows(), lambda row: Document.deserialize(row['doc']))
    limit_iterable(doc_iter)

def local_execute() -> list[Document]:
    doc_iter = all_docs
    limit_iterable(doc_iter)
```

HenryL27 (commented):
I'm not sure this does what you want it to do?

lib/sycamore/sycamore/transforms/basics.py
dhruvkaliraman7: Case 1: I used the Limit node with field in Finra env after a sort node.  Made sure to get all instances of unique value.
Case 3: I thought about this, is the correct behavior to go through all docs and then find matching lineage ids for getting correct metadata docs?

HenryL27: I wonder if the thing to do then is to make this an entirely different docset method or something - call it `limit_by_field` - and have it tack on both the sort and the custom uniqueness limit you have here

PR #1243, opened 2025-03-28
spread property data, not references

HenryL27:
if everything has to go into properties.entity this is a requirement


bsowell (approved):
Looks reasonable.

PR #1244, opened 2025-03-31
fallback to tatr if deformable fails

HenryL27:
With the two-stage approach it's possible that the encoder does not generate sufficient region proposals, which causes the inner topk to error out. We should fix that eventually but for now let's hack it so that if deformable fails, we fall back to tatr which (should) not have this issue


austin-aryn-ai (approved):
I tested out this change locally and it worked!

PR #1247, opened 2025-04-02
Bump Lint Dependencies and Relint

karanataryn:
I went down a rabbit hole after realizing our version of `ruff` is very out of date.


HenryL27 (approved):
oh man. nice.

PR #1248, opened 2025-04-03
Add Claude 3.7 Sonnet

karanataryn:


HenryL27 (approved):
Who knew adding claude 3.7 would break every test?
);

PR #1251, opened 2025-04-07
Fix clustring flaky test

bohou-aryn:


lib/sycamore/sycamore/tests/unit/transforms/test_clustering.py
```diff
@@ -11,6 +11,7 @@ class TestKMeans:
 
     @pytest.mark.skip(reason="flaky")
```
karanataryn, line 12: Let's remove the skip to test this?

alexaryn (approved):
Makes sense.

PR #1253, opened 2025-04-09
add default llm_mode to llms

HenryL27:


lib/sycamore/sycamore/llms/anthropic.py
```diff
@@ -145,15 +146,20 @@ def __init__(
...
 
     def __reduce__(self):
         def deserializer(kwargs):
             return Anthropic(**kwargs)
 
-        kwargs = {"model_name": self.model_name, "cache": self._cache}
+        kwargs = {"model_name": self.model_name, "cache": self._cache, "default_mode": self._default_mode}
         return deserializer, (kwargs,)
 
+    def default_mode(self) -> LLMMode:
```
eric-anderson, line 158: I think you can simplify this. Have each subclass pass in default_mode as a parameter to the LLM superclass (already done), and then have the LLM class return self._default_mode always.

lib/sycamore/sycamore/llms/llms.py
```diff
@@ -23,9 +23,16 @@ class LLMMode(Enum):
 class LLM(ABC):
     """Abstract representation of an LLM instance. and should be subclassed to implement specific LLM providers."""
 
-    def __init__(self, model_name, cache: Optional[Cache] = None):
+    def __init__(self, model_name, cache: Optional[Cache] = None, default_mode: Optional[LLMMode] = None):
```
eric-anderson, line 26: make default_mode not optional here.  You could do  = LLMMode.SYNC, but given how few subclasses we have I think it's better to have it be explicit.

eric-anderson (approved):

PR #1256, opened 2025-04-16
Allow override of OpenSearch user/password in run_plan().

alexaryn:


lib/sycamore/sycamore/tests/unit/test_context.py
```diff
@@ -33,6 +33,14 @@ def test_get_val_from_context():
     assert get_val_from_context(Context(params={}), "missing") is None
 
 
+def test_modified_context():
+    params = {"paramKeyA": {"llm": 1}, "paramKeyB": {"llm": ["llm1", "llm2"]}, "default": {"llm": "openai"}}
+    c0 = Context(params=params)
+    c1 = modified_context(c0, "default", llm="dromedary")
```
baitsguy, line 39: can you add a test to update a non-default context as well since access paths are different

lib/sycamore/sycamore/query/client.py
```diff
@@ -216,16 +216,22 @@ def generate_plan(
...
         return plan
 
-    def run_plan(self, plan: LogicalPlan, dry_run=False, codegen_mode=False) -> SycamoreQueryResult:
+    def run_plan(
+        self, plan: LogicalPlan, dry_run=False, codegen_mode=False, os_client_args: Optional[dict] = None
+    ) -> SycamoreQueryResult:
         """Run the given logical query plan and return a tuple of the query ID and result."""
         assert self.context is not None, "Running a plan requires a configured Context"
+        if os_client_args:
+            my_ctx = modified_context(self.context, "default", os_client_args=os_client_args)
```
baitsguy, line 225: fine for now, I wonder if we'll need to scope this to the "opensearch" namespace

baitsguy (approved):

lib/sycamore/sycamore/query/client.py
alexaryn: It seemed a bit murky as to how a piece of code like this could know what downstream namespace was relevant, or even how namespaces would be used.

lib/sycamore/sycamore/tests/unit/test_context.py
alexaryn: Coming right up!

PR #1257, opened 2025-04-16
Delete PITs after we're done reading

austin-aryn-ai:


austin-aryn-ai:
> lgtm... does this get parallelized and does opensearch give you an error if you try to delete a PIT that's already been deleted?

It doesn't get parallelized.

PR #1259, opened 2025-04-17
Inspect serialization issues only if TypeError is raised

austin-aryn-ai:
A quick search on github on how 'inspect_serializability' is used shows that this is usually called when an exception of type TypeError is raised.  This way, we avoid calling this (which calls colorama.init()) prematurely.

### DIFF
--- a/lib/sycamore/sycamore/connectors/base_reader.py
+++ b/lib/sycamore/sycamore/connectors/base_reader.py
@@ -5,6 +5,7 @@ from abc import ABC, abstractmethod
 
 from sycamore.data.document import Document
 from sycamore.plan_nodes import Scan
+from sycamore.utils.ray_utils import handle_serialization_exception
 from sycamore.utils.time_trace import TimeTrace
 
 if TYPE_CHECKING:
@@ -73,10 +74,8 @@ class BaseDBReader(Scan):
             client.close()
         return docs
 
+    @handle_serialization_exception("_client_params", "._query_params")
     def execute(self, **kwargs) -> "Dataset":
-        from sycamore.utils.ray_utils import check_serializable
-
-        check_serializable(self._client_params, self._query_params)
         from ray.data import from_items
 
         with TimeTrace("Reader"):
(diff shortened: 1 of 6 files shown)


HenryL27 (approved):
lgtm. We may still want to patch colorama to not overwrite stderr/stdout. But thanks for finding this.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -518,17 +519,14 @@ def _execute(self, **kwargs) -> "Dataset":
                 .map(lambda d: {"doc": Document(**d).serialize()})
             )
 
+    @handle_serialization_exception("_client_params", "._query_params")
```
HenryL27, line 522: ooh typo I think

lib/sycamore/sycamore/connectors/base_reader.py
```diff
@@ -73,10 +74,8 @@ def read_docs(self) -> list[Document]:
             client.close()
         return docs
 
+    @handle_serialization_exception("_client_params", "._query_params")
```
HenryL27, line 77: another one

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
austin-aryn-ai: Good catch!

lib/sycamore/sycamore/tests/unit/utils/test_ray_utils.py
```diff
@@ -1,15 +1,35 @@
...
 
 def test_non_serializable():
     import threading
 
-    lock = threading.Lock()
+    lock1 = threading.Lock()
+    lock2 = threading.Lock()
     with pytest.raises(ValueError):
-        check_serializable(lock)
+        check_serializable(lock1, lock2)
```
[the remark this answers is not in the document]
austin-aryn-ai: Ok, just wanted a test that passes more than one object.

PR #1261, opened 2025-04-18
Summarize the group name for groupby

bohou-aryn:


lib/sycamore/sycamore/query/operators/top_k.py
```diff
@@ -47,13 +47,23 @@ class GroupByCount(Node):
...
     embed_name: Optional[str] = None
+    """The field for entity embedding."""
+
     """The database field to find the top K occurences for."""
 
     cluster_field_name: str = "centroids"
+    """The centroid field used for clustering"""
 
     K: int = 5
+    """The number of top frequency occurences to look for (e.g. top 2 most common, K=2)."""
```
akarshgupta7, line 61: Can you change this comment to something like "Number of groups to form"?

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -687,9 +693,20 @@ def execute(self) -> Any:
         docset.execute()
         centroids = docset.kmeans(K=K * 2, field_name=embed_name)
```
akarshgupta7, line 694: Wondering if we can find a better way to decide what the number of clusters are

lib/sycamore/sycamore/query/operators/clustering.py
```diff
@@ -0,0 +1,24 @@
...
+
+class KMeanClustering(Node):
+    """Group documents based on a particular field.
+
+    Returns a database with ONLY 2 FIELDS: "properties.key" (which corresponds to unique values of
+    *field*) and "properties.count" (which contains the counts corresponding to unique values
+    of *field*).
+    """
+
+    entity_name: Optional[str] = None
```
baitsguy, line 14: can we call these field like in other ops, e.g. https://github.com/aryn-ai/sycamore/blob/348b9a0822cffc5917e491d17029a48599add784/lib/sycamore/sycamore/query/operators/llm_filter.py#L26

lib/sycamore/sycamore/query/operators/clustering.py
```diff
@@ -0,0 +1,24 @@
...
+    """The database field to find the top K occurences for."""
+
+    embed_name: Optional[str] = None
```
baitsguy, line 17: we probably don't need this in the query operator? at least until the planner can also compute embeddings and have a new field?

lib/sycamore/sycamore/query/operators/top_k.py
```diff
@@ -38,22 +38,16 @@ class TopK(Node):
...
     Returns a database with ONLY 2 FIELDS: "properties.key" (which corresponds to unique values of
     *field*) and "properties.count" (which contains the counts corresponding to unique values
     of *field*).
     """
 
     entity_name: Optional[str] = None
-    embed_name: Optional[str] = None
     """The database field to find the top K occurences for."""
 
     cluster_field_name: str = "centroids"
```
baitsguy, line 52: should only need 1 field right? since clustering is a separate op

lib/sycamore/sycamore/query/operators/clustering.py
```diff
@@ -0,0 +1,24 @@
+from typing import Optional
```
akarshgupta7, line 1: I think that this operator should be with all the other group by in lib/sycamore/sycamore/query/operators/top_k.py. Maybe we rename top_k.py to group_by.py or something?

lib/sycamore/sycamore/query/operators/aggregate.py
```diff
@@ -0,0 +1,22 @@
...
+    of *field*).
+    """
+
+    llm_summary: bool = False
+    """If true, use llm_summary to summarize each group; If False, randomly select one entity from each group."""
+
+    llm_summary_instruction: Optional[str] = Field(default=None, json_schema_extra={"exclude_from_comparison": True})
+    """An instruction of what the group name should be about if llm_summary is True. E.g. if the
+    purpose of this operation is to group by cities, llm_cluster_instruction
+    could be 'The city name of this group'"""
```
akarshgupta7, line 22: should we combine this with the existing count operator? I have no strong opinions.

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -692,10 +693,91 @@ def execute(self) -> Any:
...
+        llm = None
+        if logical_node.llm_summary:
+            llm = get_val_from_context(
+                context=self.context, val_key="llm", param_names=[OperationTypes.BINARY_CLASSIFIER.value]
+            )
+            assert isinstance(llm, LLM), "GroupBy with llm summary requies an 'llm' configured on the Context"
+
+        result = self.inputs[0].collect()
+
+        if llm:
```
akarshgupta7, line 767: why are we finding the summary in SycamoreAggregateCount? I think we should find the summary in different GroupBy operators and this operator just count those grouped entities and add it to properties.count field in the docset.

lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -692,10 +693,91 @@ def execute(self) -> Any:
...
+        assert isinstance(logical_node, AggregateCount)
+
+        llm = None
+        if logical_node.llm_summary:
+            llm = get_val_from_context(
+                context=self.context, val_key="llm", param_names=[OperationTypes.BINARY_CLASSIFIER.value]
+            )
+            assert isinstance(llm, LLM), "GroupBy with llm summary requies an 'llm' configured on the Context"
+
+        result = self.inputs[0].collect()
```
akarshgupta7, line 765: Does this automatically add a new field `properties.count` to the docset?

lib/sycamore/sycamore/query/operators/top_k.py
akarshgupta7, line 52: we should remove cluster_field_name as well from this operator

lib/sycamore/sycamore/query/operators/aggregate.py
akarshgupta7: nvm I think it's best to keep them separate

lib/sycamore/sycamore/query/operators/top_k.py
bohou-aryn: no, cluster_field_name is for the embedding used for groupby while the entity name is for the original content, which you want to output, you may not want output the a list embedding floats.

lib/sycamore/sycamore/query/operators/clustering.py
bohou-aryn: not sure I understand this one, do you mean embedding is separated out from the clustering?

lib/sycamore/sycamore/query/operators/clustering.py
bohou-aryn: sure, you mean field instead of name, right?

akarshgupta7: I think he means just field like we have for other operators like LLMFilter etc

lib/sycamore/sycamore/query/operators/top_k.py
akarshgupta7: This is a traditional group by right where we don't do clustering?

lib/sycamore/sycamore/query/operators/clustering.py
baitsguy: yep

lib/sycamore/sycamore/query/operators/clustering.py
baitsguy: I didn't understand what this field means?

bohou-aryn: for embedding

lib/sycamore/sycamore/query/operators/top_k.py
bohou-aryn: no

akarshgupta7: then what's the difference between this and the KmeansClustering operator?

baitsguy (approved):

PR #1262, opened 2025-04-18
better defaults for aryn writer

HenryL27:
- name change: autoschema -> update_schema
- update_schema defaults to True. May need to set it to false in some of the JS scripts
- default aryn_url is now "https://api.aryn.ai/v1/storage" instead of ""


lib/sycamore/sycamore/utils/aryn_config.py
```diff
@@ -7,6 +7,7 @@
 
 DEFAULT_PATH = os.path.join(pathlib.Path.home(), ".aryn", "config.yaml")
 _DEFAULT_PATH = DEFAULT_PATH
+DEFAULT_STORAGE_URL = "https://api.aryn.ai/v1/storage"
 ## ToDO: remove after 2024-08-31
```
austin-aryn-ai, line 11: Still here!

austin-aryn-ai (approved):

PR #1268, opened 2025-04-23
Don't include original_elements in queries

baitsguy:
^

### DIFF
--- a/lib/sycamore/sycamore/query/execution/sycamore_operator.py
+++ b/lib/sycamore/sycamore/query/execution/sycamore_operator.py
@@ -75,7 +75,7 @@ class SycamoreOperator(PhysicalOperator):
         return self.get_node_args()
 
 
-def remove_extra_fields(doc):  # these are fields we don't want to pass to query operations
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/query/execution/sycamore_operator.py
```diff
@@ -75,6 +75,12 @@ def get_execute_args(self) -> Dict:
         return self.get_node_args()
 
 
+def remove_extra_fields(doc):  # these are fields we don't want to pass to query operations
+    if "_original_elements" in doc.properties:
```
austin-aryn-ai, line 79: You want to get these fields from settings?  Or just hard code this one field?

austin-aryn-ai: Or you can just call this `remove_original_elements`.

baitsguy: I suspect we're going to need more so left it. settings is a lot of layers to plumb through sadly :|

austin-aryn-ai (approved):

PR #1269, opened 2025-04-24
Detect article-like pages entirely in two columns when generating markdown

MarkLindblad:


lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -131,16 +131,49 @@ def bbox_sort_based_on_tags(elems: list[Element]) -> None:
...
-        elem.data["_coltag"] = col_tag(elem)
-    tag_two_columns(elems)
-    bbox_sort_based_on_tags(elems)
-    for elem in elems:
-        elem.data.pop("_coltag", None)  # clean up tags
+    is_entirely_two_columns, left_col, _ = check_entirely_two_columns(elems)
+    if is_entirely_two_columns:
+        # if the entire page is two columns, sort the left column first
+        elems.sort(key=generate_elem_prefer_left(left_col))
+    else:
```
alexaryn, line 170: I'd just return here, making the special case self-contained and not needing the else and indentation.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -131,16 +131,49 @@ def bbox_sort_based_on_tags(elems: list[Element]) -> None:
...
 
 
+def check_entirely_two_columns(elems: list[Element]) -> tuple[bool, list[Element], list[Element]]:
+    """
+    Check if the elements are entirely two columns. If so, return True and the left and right column elements.
+    """
+    left_col = []
+    right_col = []
+    for elem in elems:
+        if elem.type == "Page-footer" or elem.type == "Page-header":
```
alexaryn, line 141: Should probably also exclude titles.  Not sure about section headers.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -131,16 +131,49 @@ def bbox_sort_based_on_tags(elems: list[Element]) -> None:
...
+            continue
+        x1, _, x2, _ = elem.data["bbox"]
+        if x1 < 0.6 and x2 < 0.6:
+            left_col.append(elem)
+        elif x1 > 0.4 and x2 > 0.4:
+            right_col.append(elem)
```
alexaryn, line 147: Conceptually, this gets fooled by a 5-column newspaper-ish layout.  It puts the first three columns in left.  It also calls something two-column if it's one column on the side, but that may end up OK after processing.

The .4 and .6 constants are kind-of magic.  I'm sure there will be false positives and false negatives.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -131,16 +131,49 @@ def bbox_sort_based_on_tags(elems: list[Element]) -> None:
         bbox_sort_two_columns(elems, lidx, len(elems))
 
 
+def check_entirely_two_columns(elems: list[Element]) -> tuple[bool, list[Element], list[Element]]:
+    """
+    Check if the elements are entirely two columns. If so, return True and the left and right column elements.
+    """
+    left_col = []
+    right_col = []
```
alexaryn, line 139: Nothing uses right_col.  Might as well nuke it, unless you plan to use it.  Perhaps if either array is empty, return false and sort by y.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -131,16 +131,49 @@ def bbox_sort_based_on_tags(elems: list[Element]) -> None:
...
+        if elem.type == "Page-footer" or elem.type == "Page-header":
+            continue
+        x1, _, x2, _ = elem.data["bbox"]
+        if x1 < 0.6 and x2 < 0.6:
+            left_col.append(elem)
+        elif x1 > 0.4 and x2 > 0.4:
+            right_col.append(elem)
+        else:
+            return False, [], []
+    return True, left_col, right_col
```
alexaryn, line 150: Instead of returning a list of "lefts", we could use the existing tagging approach used in this file.  Then we might not need a fancy closure-based sort key.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -131,16 +131,49 @@ def bbox_sort_based_on_tags(elems: list[Element]) -> None:
...
-    for elem in elems:  # tag left/right/full based on width/position
-        elem.data["_coltag"] = col_tag(elem)
-    tag_two_columns(elems)
-    bbox_sort_based_on_tags(elems)
-    for elem in elems:
-        elem.data.pop("_coltag", None)  # clean up tags
+    is_entirely_two_columns, left_col, _ = check_entirely_two_columns(elems)
+    if is_entirely_two_columns:
+        # if the entire page is two columns, sort the left column first
+        elems.sort(key=generate_elem_prefer_left(left_col))
```
alexaryn, line 169: What would happen if we just used `elem_left_top` here?

MarkLindblad:
While this solution improves robustness, we have chosen an even more robust solution here: #1273.

PR #1270, opened 2025-04-24
Add metadata loading to materialize

HenryL27:
1. Adds a Materialize.load_metadata() that loads all metadata documents in memory for the materialize
2. Adds a docset.get_materializes() that returns a list of all the materializes in the docset plan (param for fifo/lifo)
3. Adds a QueryResponse.get_metadata() that returns the metadata documents for the last bits of the inner sycamore pipeline

### DIFF
--- a/lib/sycamore/sycamore/materialize.py
+++ b/lib/sycamore/sycamore/materialize.py
@@ -539,6 +539,34 @@ class Materialize(UnaryNode):
 
         return ret
 
+    def load_metadata(self) -> list[MetadataDocument]:
+        self._verify_has_files()
+        if not self._fshelper.file_exists(self._success_path()):
+            logging.warning(f"materialize.success not found in {self._orig_path}. Returning partial data")
+        from sycamore.utils.sycamore_logger import LoggerFilter
+
+        limited_logger = logging.getLogger(__name__ + ".limited_local_source")
+        limited_logger.addFilter(LoggerFilter())
+        ret = []
+        count = 0
+        for fi in self._fshelper.list_files(self._root):
+            if self._path_filter is not None and not self._path_filter(fi.path):
+                continue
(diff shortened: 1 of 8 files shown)


lib/sycamore/sycamore/materialize.py
```diff
@@ -539,6 +539,34 @@ def local_source(self) -> list[Document]:
...
         return ret
 
+    def load_metadata(self) -> list[MetadataDocument]:
+        self._verify_has_files()
+        if not self._fshelper.file_exists(self._success_path()):
+            logging.warning(f"materialize.success not found in {self._orig_path}. Returning partial data")
+        from sycamore.utils.sycamore_logger import LoggerFilter
+
+        limited_logger = logging.getLogger(__name__ + ".limited_local_source")
+        limited_logger.addFilter(LoggerFilter())
```
eric-anderson, line 549: Can we rename LoggerFilter to RateLimitLogger?  I had to look at the code to figure out what that was doing.

lib/sycamore/sycamore/materialize.py
```diff
@@ -539,6 +539,34 @@ def local_source(self) -> list[Document]:
 
         return ret
 
+    def load_metadata(self) -> list[MetadataDocument]:
+        self._verify_has_files()
+        if not self._fshelper.file_exists(self._success_path()):
+            logging.warning(f"materialize.success not found in {self._orig_path}. Returning partial data")
+        from sycamore.utils.sycamore_logger import LoggerFilter
+
+        limited_logger = logging.getLogger(__name__ + ".limited_local_source")
```
eric-anderson, line 548: I'd name this something like .load_metadata; since the primary use would be to control that logger separately from other ones.

eric-anderson (approved):

PR #1273, opened 2025-04-25
Improve bbox sorting by allowing `center` and `tolerance` to be customized

MarkLindblad:


lib/sycamore/sycamore/tests/unit/utils/test_margin.py
```diff
@@ -0,0 +1,43 @@
...
+@pytest.mark.parametrize(
+    "original_bboxes, expected_final_coordinates",
+    [
+        (
+            [(0.0, 0.0, 0.0, 0.0), (1.0, 1.0, 1.0, 1.0), (0.1, 0.2, 0.3, 0.4)],
+            [(0.0, 0.0, 0.0, 0.0), (1.0, 1.0, 1.0, 1.0), (0.1, 0.2, 0.3, 0.4)],
+        ),
+        (
+            [(0.1, 0.1, 0.9, 0.9)],
+            [(0.0, 0.1, 1.0, 0.9)],
```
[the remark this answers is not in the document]
MarkLindblad: Done.

lib/sycamore/sycamore/tests/unit/utils/test_margin.py
```diff
@@ -0,0 +1,43 @@
...
+    original_bboxes: list[tuple[float, float, float, float]],
+    expected_final_coordinates: list[tuple[float, float, float, float]],
+) -> None:
+    elements = [Element({"bbox": bbox}) for bbox in original_bboxes]
+    margin_transform_page(elements, leave_original_tags=True)
+    for element, expected_bbox, original_bbox in zip(elements, expected_final_coordinates, original_bboxes):
+        if (eb := element.bbox) is not None:
+            bbox = tuple(eb.to_list())
+            if bbox != expected_bbox:
+                raise AssertionError(f"When inspecting .bbox, expected {expected_bbox}, got {bbox}")
```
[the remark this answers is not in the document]
MarkLindblad: No, we don't. I've now fixed this to be simpler.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,77 @@
...
+    for elem in elems:
+        if elem.data.get("bbox") is not None and (old_bbox := elem.data.get(margin_tag)) is not None:
+            elem.data["bbox"] = old_bbox
+            del elem.data[margin_tag]
+
+
+def find_reasonable_margin_page(elements: list[Element]) -> tuple[float, float, float, float]:
+    left, top, right, bottom = find_margin_page(elements)
+    left = min(left, 0.4)
+    right = max(right, 0.6)
```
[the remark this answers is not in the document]
MarkLindblad: Done.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,77 @@
...
+    :return: tuple of (left, top, right, bottom)
+    """
+    left = float("inf")
+    top = float("-inf")
+    right = float("-inf")
+    bottom = float("inf")
+    set_at_least_once = False
+
+    for elem in elements:
+        if elem.type in ("Page-header", "Image", "Page-footer"):
```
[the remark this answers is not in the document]
MarkLindblad: Let's add this as a follow-up.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -143,6 +145,14 @@ def bbox_sort_page(elems: list[Element]) -> None:
         elem.data.pop("_coltag", None)  # clean up tags
 
 
+def bbox_margin_sort_page(elems: list[Element]):
+    if len(elems) < 2:
+        return
+    margin_transform_page(elems, leave_original_tags=True)
+    bbox_sort_page(elems)
+    revert_margin_transform_page(elems)
```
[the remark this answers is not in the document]
MarkLindblad: I've now changed it to use that approach.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
+
+from sycamore.data import Element, BoundingBox
+
+margin_tag = "_pre_margin_transform_bbox"
```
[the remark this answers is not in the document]
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -57,7 +58,7 @@ def col_tag(elem: Element) -> Optional[str]:
         width = right - left
         if width > 0.6 or elem.type == "Page-footer":
             return "full"
-        elif (width < 0.1) or (width >= 0.45):
+        elif (width < 0.1) or (width >= 0.5):
```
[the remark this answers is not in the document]
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -49,15 +56,16 @@ def collect_pages(elems: list[Element]) -> list[list[Element]]:
     return rv
 
 
-def col_tag(elem: Element) -> Optional[str]:
-    bbox = elem.data.get("bbox")
-    if bbox:
+def col_tag(elem: Element, transform: Optional[np.ndarray] = None, max_width: float = 0.45) -> Optional[str]:
```
alexaryn, line 59: Let's keep this function simple and true to its name.  Let's use a different key in `elem.data` to cache the transformed bbox.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -49,15 +56,16 @@ def collect_pages(elems: list[Element]) -> list[list[Element]]:
...
+    cached_bbox = get_bbox_prefer_cached(elem, transform)
```
alexaryn, line 60: Rather than `get_bbox_prefer_cached()` let's have a function`get_transformed_bbox()` which can handle caching internally.  Caching is a performance optimization/implementation detail that callers needn't know/worry about.  What is critical for callers, though, is to know they're getting the transformed coordinates.

No need to rename the variable `cached_bbox` as it's the only `bbox`.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -131,16 +140,27 @@ def bbox_sort_based_on_tags(elems: list[Element]) -> None:
...
+def bbox_sort_page(elems: list[Element], transform: Optional[np.ndarray] = None, max_width: float = 0.45) -> None:
     if len(elems) < 2:
         return
-    elems.sort(key=elem_top_left)  # sort top-to-bottom, left-to-right
+    if transform is None:
+        transform = np.eye(3)
+    elems.sort(key=generate_elem_top_left(transform))  # sort top-to-bottom, left-to-right
     for elem in elems:  # tag left/right/full based on width/position
-        elem.data["_coltag"] = col_tag(elem)
+        elem.data["_coltag"] = col_tag(elem, transform, max_width)
```
alexaryn, line 150: Let's leave this alone.  It's wasteful to slap the same transform and max_width onto every element.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
+from typing import Optional
+import numpy as np
+
+from sycamore.data import Element, BoundingBox
+from sycamore.utils.sycamore_logger import get_logger, setup_logger
+
+setup_logger()
+g_logger = get_logger()
+
+cached_bbox_tag = "_bbox_accounting_for_margins"
```
alexaryn, line 10: I'd call this `_transformed_bbox` as that's most accurate and future-proof.  It also belongs in the other file where matters of caching should be located.  This file produces the transformation matrix; the other file utilizes it.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def find_transform_page(elems: list[Element]) -> tuple[bool, np.ndarray]:
+    is_reasonable, left, top, right, bottom = find_margins_and_check_are_reasonable(elems)
```
alexaryn, line 14: This is a lot of data being returned.  I think it would be cleaner to have a "margin finder" or "active region locator" class that would have these five values as data members.  Positional tuple unpacking can be error-prone.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+    width = right - left
+    height = bottom - top
+    if is_reasonable:
```
alexaryn, line 17: I don't think this code needs to return `is_reasonable` to its caller.  This code is in a better position to decide what to do than the caller is.  Even if it's a tie, less dirty laundry is better.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+        # fmt: off
+        return is_reasonable, np.array([[1/width, 0,          -left/width],
+                                        [0,         1/height, -top/height],
+                                        [0,         0,        1]])
+        # fmt: on
+    else:
+        return is_reasonable, np.eye(3)
+
+
+def get_bbox_prefer_cached(elem: Element, transform: Optional[np.ndarray]) -> Optional[BoundingBox]:
```
alexaryn, line 27: This functionality belongs in the other file.  It should be `get_transformed_bbox()`.  I'm pretty sure `transform` will never be None.  We should use a different name than "transform" in Sycamore to avoid confusion.  Maybe "matrix".

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+    if (cached := elem.data.get(cached_bbox_tag)) is not None:
+        return cached
+    elif (bbox := elem.bbox) is not None:
+        cache = apply_transform(bbox, transform)
```
alexaryn, line 31: I'd call this `result` instead of `cache` to improve clarity.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+        elem.data[cached_bbox_tag] = cache
+        return cache
+    else:
+        return None
+
+
+def apply_transform(bbox: BoundingBox, transform: Optional[np.ndarray]) -> BoundingBox:
+    g_logger.info(f"Applying transform {transform} to bbox {bbox}")
+    if transform is None:
```
alexaryn, line 40: Again, I think the matrix will never be None.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -6,17 +6,24 @@
...
-def elem_top_left(elem: Element) -> tuple:
-    bbox = elem.data.get("bbox")
-    if bbox:
-        return (bbox[1], bbox[0])
-    return (0.0, 0.0)
+def generate_elem_top_left(transform: np.ndarray) -> Callable[[Element], tuple[float, float]]:
+    def elem_top_left(elem: Element) -> tuple:
+        cached_bbox = get_bbox_prefer_cached(elem, transform)
+        if cached_bbox:
+            bbox = cached_bbox.to_list()
```
alexaryn, line 22: The existing code didn't need to call `to_list()` all the time.  The new code should figure out how to avoid it the same way.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+                                        [0,         0,        1]])
+        # fmt: on
+    else:
+        return is_reasonable, np.eye(3)
+
+
+def get_bbox_prefer_cached(elem: Element, transform: Optional[np.ndarray]) -> Optional[BoundingBox]:
+    if (cached := elem.data.get(cached_bbox_tag)) is not None:
+        return cached
+    elif (bbox := elem.bbox) is not None:
```
alexaryn, line 30: I think the secret here is to call `elem.data.get("bbox")` to avoid needless conversions.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+        cache = apply_transform(bbox, transform)
+        elem.data[cached_bbox_tag] = cache
+        return cache
+    else:
+        return None
+
+
+def apply_transform(bbox: BoundingBox, transform: Optional[np.ndarray]) -> BoundingBox:
```
alexaryn, line 38: I think this function will end up taking a tuple and returning a tuple.  We should probably rename "transform" to "matrix"

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+    g_logger.info(f"Applying transform {transform} to bbox {bbox}")
+    if transform is None:
+        return bbox
+    x1, y1, x2, y2 = bbox.to_list()
+    # fmt: off
+    old_coords = np.array([[x1, x2],
```
alexaryn, line 44: Nice fancy two-at-once matrix multiplication.  Might deserve a comment.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+                            [y1, y2],
+                            [1,  1]])
+    # fmt: on
+    new_coords = np.dot(transform, old_coords)
+    new_x1, new_x2 = new_coords[0]
+    new_y1, new_y2 = new_coords[1]
+    return BoundingBox(new_x1, new_y1, new_x2, new_y2)
+
+
+def clear_cached_bboxes(elems: list[Element]) -> None:
```
alexaryn, line 54: This should be in the other file.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+    for elem in elems:
+        elem.data.pop(cached_bbox_tag, None)
+
+
+def find_margins_and_check_are_reasonable(elements: list[Element]) -> tuple[bool, float, float, float, float]:
```
alexaryn, line 59: In terms of clean and evolve-able code, I'd make a function called `margins_are_reasonable()` where the kitchen sink of checks can live without cluttering the readability of the main logic.

lib/sycamore/sycamore/utils/margin.py
```diff
@@ -0,0 +1,102 @@
...
+    if right < 0.6:
+        is_reasonable = False
+    if top > 0.4:
+        is_reasonable = False
+    if bottom < 0.6:
+        is_reasonable = False
+    return is_reasonable, left, top, right, bottom
+
+
+def find_margin_page(elements: list[Element]) -> tuple[float, float, float, float]:
```
alexaryn, line 73: This name seems to imply that it finds the margin of one page.  How about `find_margin_of_pages()`?

alexaryn (commented):
This is heading in the right direction, but it still seems like the code wants to be organized into a lower-complexity state.  Once that's attained, the code will be easier to understand, test, debug, maintain, and extend.  It'll also be more efficient and likely smaller.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -6,17 +6,24 @@
...
 """
 
-from typing import Optional
+from typing import Callable, Optional
+
+import numpy as np
 
 from sycamore.data import Document, Element
 from sycamore.data.document import DocumentPropertyTypes
+from sycamore.utils.margin import clear_cached_bboxes, find_transform_page, get_bbox_prefer_cached
```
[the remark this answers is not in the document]
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -6,17 +6,24 @@
...
 
 
-def elem_top_left(elem: Element) -> tuple:
-    bbox = elem.data.get("bbox")
-    if bbox:
-        return (bbox[1], bbox[0])
-    return (0.0, 0.0)
+def generate_elem_top_left(transform: np.ndarray) -> Callable[[Element], tuple[float, float]]:
+    def elem_top_left(elem: Element) -> tuple:
```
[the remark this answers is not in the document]
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -49,15 +56,16 @@ def collect_pages(elems: list[Element]) -> list[list[Element]]:
...
+    cached_bbox = get_bbox_prefer_cached(elem, transform)
+    if cached_bbox:
+        bbox = cached_bbox.to_list()
         left = bbox[0]
         right = bbox[2]
         width = right - left
         if width > 0.6 or elem.type == "Page-footer":
             return "full"
-        elif (width < 0.1) or (width >= 0.45):
+        elif (width < 0.1) or (width >= max_width):
```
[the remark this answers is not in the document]
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -131,16 +140,27 @@ def bbox_sort_based_on_tags(elems: list[Element]) -> None:
...
-        elem.data["_coltag"] = col_tag(elem)
+        elem.data["_coltag"] = col_tag(elem, transform, max_width)
     tag_two_columns(elems)
     bbox_sort_based_on_tags(elems)
     for elem in elems:
         elem.data.pop("_coltag", None)  # clean up tags
+    clear_cached_bboxes(elems)
+
+
+def bbox_margin_sort_page(elements: list[Element]) -> None:
```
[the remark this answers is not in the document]
MarkLindblad: Do you really want to force the consumers of `bbox_sort.py` who want margin based sorting to call `find_transform_page` manually like this?
```python
bbox_sort_page(elements, find_transform_page(elements))
```

lib/sycamore/sycamore/utils/bbox_sort.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
MarkLindblad: I've now done this in a different, more readable way.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: I've now done something similar.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/margin.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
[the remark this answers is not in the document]
MarkLindblad: Did you intend for users who want bbox sorting that accounts for margins to import a `bbox_margin_sort_page` from `margin.py`?

lib/sycamore/sycamore/utils/bbox_sort.py
MarkLindblad: Done.

lib/sycamore/sycamore/utils/bbox_sort.py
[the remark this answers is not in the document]
alexaryn: I have no problem with those additional bits of code.  The thought is that the correct calculation of the transformation matrix could become complex and dependent on various options and document properties.  It may or may not vary from page to page.  This is all too much complexity for `bbox_sort.py` to sign up for.  The cleaner approach is to just make the caller in charge.  This is an example of "separation of concerns".  There are also aspects of "separability" and "modularity" involved in the sense that we may end up with multiple alternative margin finders.  We may also have other places in the code where finding margins is useful.

MarkLindblad:
Superseded by #1301

PR #1275, opened 2025-04-25
Put os_client_args in default namespace to avoid surprise.

alexaryn:
Added some checks to prevent an override with no default value under it.  Also now log how we connect to OpenSearch (without password in log).


baitsguy (approved):
tests before merging pls, code looks good

PR #1277, opened 2025-04-29
Add better view_pdf function

HenryL27:
Adds `view_pdf` function which displays the pdf as an iframe or a list of links to the files. Configurable with display_name- and url- generation functions, which can be set globally at the module level if you want. I also have a default for local files which runs a python stdlib simple http server (`file:///path/to/doc.pdf` didn't work bc jupyter sandboxes its html thingy)

### DIFF
--- a/lib/sycamore/sycamore/tests/unit/utils/test_jupyter.py
+++ b/lib/sycamore/sycamore/tests/unit/utils/test_jupyter.py
@@ -1,0 +1,17 @@
+import sycamore.utils.jupyter as suj
+from sycamore.data import Document
+
+
+def test_init_viewpdf():
+    def foo(doc: Document) -> str:
+        return "foo"
+
+    def bar(doc: Document) -> str:
+        return "bar"
+
+    suj.init_viewpdf(foo, bar)
+    assert suj._doc_to_url is foo  # type: ignore
+    assert suj._doc_to_display_name is bar  # type: ignore
+
+    delattr(suj, "_doc_to_url")
+    delattr(suj, "_doc_to_display_name")
--- a/lib/sycamore/sycamore/tests/unit/utils/test_jupyter.py
+++ b/lib/sycamore/sycamore/tests/unit/utils/test_jupyter.py
@@ -1,17 +1,29 @@
-import sycamore.utils.jupyter as suj
-from sycamore.data import Document
+from sycamore.utils.jupyter import LocalFileViewer
+import socket
+import pytest
+import time
+import gc
 
 
-def test_init_viewpdf():
-    def foo(doc: Document) -> str:
-        return "foo"
+def test_localfileviewer_no_leaks():
+    port = 2647
+    viewpdf = LocalFileViewer(port=port)
 
-    def bar(doc: Document) -> str:
-        return "bar"
+    # Wait for subprocess server to start
+    time.sleep(0.1)
+    with pytest.raises(Exception):
+        socket.create_server(("localhost", port)).close()
 
-    suj.init_viewpdf(foo, bar)
-    assert suj._doc_to_url is foo  # type: ignore
-    assert suj._doc_to_display_name is bar  # type: ignore
+    newport = 2648
+    viewpdf = LocalFileViewer(port=newport)
+    # gc the original viewpdf
+    gc.collect()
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/tests/unit/utils/test_jupyter.py
```diff
@@ -0,0 +1,17 @@
+import sycamore.utils.jupyter as suj
+from sycamore.data import Document
+
+
+def test_init_viewpdf():
```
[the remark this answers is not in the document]
HenryL27: I think this is a useful thing to have outside of the context of a QueryEval, but ack on the globals thing. plan:
1. make a PDFViewer class in sycamore.utils.jupyter (maybe set the __call__ method of that to view_pdf(docset))
2. add one as an optional field to QueryEval (in managed-service repo) + a pass-through method

lib/sycamore/sycamore/utils/jupyter.py
```diff
@@ -57,10 +63,200 @@ def reimport(module_name: str) -> None:
...
+                s = socket.create_server(("localhost", port))
+                s.close()
+                self._server_owner = True
+            except Exception:
+                if not suppress_used_port_warning:
+                    logging.warning(
+                        f"Port {port} is already in use. Will assume that it is an already-running LocalFileViewer server."
+                    )
+        else:
+            max_tries = 10
```
eric-anderson, line 215: Alex pointed out:
`for port in random.sample(range(1025, 65536), 10):`
does the sampling without replacement.

lib/sycamore/sycamore/utils/jupyter.py
```diff
@@ -57,10 +63,200 @@ def reimport(module_name: str) -> None:
...
+            tried: set[int] = set()
+            while len(tried) < max_tries:
+                port = random.randrange(1025, 65536)
+                if port in tried:
+                    continue
+                try:
+                    s = socket.create_server(("localhost", port))
+                    self._port = port
```
eric-anderson, line 223: same thing here, I think just a local port rather than class port works.

lib/sycamore/sycamore/utils/jupyter.py
```diff
@@ -57,10 +63,200 @@ def reimport(module_name: str) -> None:
...
     except Exception as e:
         print(f"Error reloading module '{module_name}': {e}")
+
+
+class PDFViewer(ABC):
+    """
+    Base class for viewing PDF files in a jupyter notebook.
+    """
+
+    @abstractmethod
```
eric-anderson, line 80: You're going to need another subclass to make this work in a customer container, doc_to_url will write the binary out to a file at a specified path and then return a magic URL that nginx can serve. doc_to_display_name will need to fetch the customer name property, but that should be easy to do when you downstream it.

eric-anderson (approved):

lib/sycamore/sycamore/utils/jupyter.py
HenryL27: fancy

lib/sycamore/sycamore/utils/jupyter.py
HenryL27: want it on the class so the url generator knows where to point

lib/sycamore/sycamore/utils/jupyter.py
```diff
@@ -57,10 +63,200 @@ def reimport(module_name: str) -> None:
...
+                    continue
+                try:
+                    s = socket.create_server(("localhost", port))
+                    self._port = port
+                    s.close()
+                    self._server_owner = True
+                    break
+                except Exception:
+                    tried.add(port)
+            if not self._server_owner:
```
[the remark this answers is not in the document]
HenryL27: Don't wanna have to dupe the subprocess code in each branch (port specified vs unspecified). But can be a local var. Actually I factor out the port logic entirely that's prob cleanest

lib/sycamore/sycamore/utils/jupyter.py
HenryL27: yep. I added it to the redownstreaming list in my upstreaming list

PR #1286, opened 2025-05-06
Add instructions to split entities in LLM extract entity.

akarshgupta7:


bohou-aryn:
Cool, still, a manual test validating the result would be great.

akarshgupta7:
> Cool, still, a manual test validating the result would be great.

How do we do manual testing?

bohou-aryn (approved):
Since there is no easy way for you to test this, let's push to test and validate. This needs to guarantee no regression happen for other queries and also generate newline separated entities for >> 2 cases.

PR #1287, opened 2025-05-06
Add heuristics to compute K for kmeans based on docset size.

akarshgupta7:


bohou-aryn (approved):
We need better algorithm than this to figure K, let's merge this in to test end to end first.

akarshgupta7:
> We need better algorithm than this to figure K, let's merge this in to test end to end first.

This is what people in ML community typically use for means

PR #1296, opened 2025-05-12
Minor fixups from testing with managed-service.

eric-anderson:
llms/__init__.py: Remove exporting the "models". Some uses expected them to be classes, and
  inherited. That didn't work so switch to not exporting.

operations.py: only make one tokenizer. Making them is kinda slow.

sycamore_operator.py: importing get_knn_query is slow. Only do it on demand

utils/__init__.py: Move the requires module logic inside the fucntion so that if
  someone specifies the device they want we don't end up loading torch.


HenryL27 (approved):
lgtm

PR #1301, opened 2025-05-15
Initial draft of X-Y Cut reading order for Sycamore (Hackathon)

alexaryn:
This isn't hooked up to live code yet.  It'll need an option to select it (or kill-switch).


lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def finalize(self) -> "NodeLeaf":
+        if not self.elists[-1]:
+            self.elists.pop()
+        return self
+
+    def cansplit(self) -> bool:
+        try:
+            elem = self.elists[0][0]
+            next = self.elists[1]  # noqa: F841
```
eric-anderson, line 106: return len(self.elists) >= 2 and isinstance(self.elists[0][0], Element)

if isinstance(self.elists, int) shouldn't go to False, that should be an assertion.

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+            return isinstance(elem, Element)
+        except (IndexError, TypeError):
+            return False
+
+    def to_str_impl(self, pfx: str, sio: StringIO) -> None:
+        for idx, elist in enumerate(self.elists):
```
eric-anderson, line 112: for elist in self.elists:`

idx is unused.

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+            elems.extend(elist)
+
+
+###############################################################################
+
+
+def get_bbox(elem: Element) -> tuple[float, float, float, float]:
+    if bbox := elem.data.get("bbox"):
+        return bbox
+    return (1.0, 1.0, 1.0, 1.0)
```
eric-anderson, line 135: I'd use 
```
return (1,1,1,1) # range on bbox is [0, 1] this sorts at the end.
```

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+    s = bbox_to_str(get_bbox(elem))
+    if tr := elem.text_representation:
+        nonl = tr.replace("\n", " ")[:32]
+        s += f"[{nonl}]"
+    return s
+
+
+def make_begin_end(elems: ElemList, axis: int) -> BeginEndList:
+    """Returns array of (coord, isopen, elem)"""
+    ary: BeginEndList = []
```
eric-anderson, line 153: ary isn't meaningful, it's just the variable type (which should be lst). 
Use either bel (match the type name); or ret (how it's used); or order (what you used later).

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+    for elem in elems:
+        bbox = get_bbox(elem)
+        aa = bbox[axis]
```
eric-anderson, line 156: nit: a & b?  Less characters.

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+        bb = bbox[axis + 2]
+        if bb < aa:
```
eric-anderson, line 158: I feel like we should be guaranteeing a <b, and just have assert a <= b.  Fix it elsewhere if we get it wrong.

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+from io import StringIO
+from typing import Generator, Optional
+
+from sycamore.data import Document, Element
+from sycamore.utils.bbox_sort import collect_pages
+
+ElemList = list[Element]
+BeginEndList = list[tuple[float, int, Element]]
+
+XAXIS = 0
```
eric-anderson, line 16: `# The enum values are important bbox[XAXIS], bbox[XAXIS+2] need to be the left and right sides of the bbox.`

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+        bb = bbox[axis + 2]
+        if bb < aa:
+            aa, bb = bb, aa
+        ary.append((aa, OPEN, elem))
+        ary.append((bb, CLOSE, elem))
+    ary.sort()
+    return ary
+
+
+def gen_overlap(ary: BeginEndList) -> Generator[tuple, None, None]:
```
eric-anderson, line 166: ary -> bel or order or input

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+    """Yields tuple (coord, isopen, elem, count, width)"""
+    count = 0
+    cur = ary[0]
+    n = len(ary)
+    for ii in range(n):
```
eric-anderson, line 171:
```
for i, cur in enumerate(bel):
    width = -1.0
    if cur[1] == OPEN:
        count += 1
    else:
        count -= 1
        if count == 0 and i+1 < n:
            width = bel[i+1][0] - cur[0]
     yield (*cur, count, width)
```

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+                next = ary[ii + 1]
+                if count == 0:
+                    width = next[0] - cur[0]
+        yield (*cur, count, width)
+        cur = next
+
+
+def widest_cut(order: BeginEndList) -> tuple[float, Element]:
+    rv = (-1.0, order[0][2])
+    if len(order) < 3:  # order is twice as long as elems
```
eric-anderson, line 188: `# order is twice as long as elems; if there is only 1 element there is no widest cut`

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def choose_axis(elems: ElemList) -> Optional[tuple[BeginEndList, Element]]:
+    xorder = make_begin_end(elems, XAXIS)
+    yorder = make_begin_end(elems, YAXIS)
+    xw, xe = widest_cut(xorder)
+    yw, ye = widest_cut(yorder)
+    if max(xw, yw) < 0.0:
+        return None
+    if xw < yw:
```
eric-anderson, line 203: remove debugging or change it to:
```
if False:
    ... debugging stuff ...
```
to make it easier to turn on and to have typechecking check it.

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+
+def cleave_elems(elems: ElemList) -> NodeLeaf:
+    """Binary split across widest gap."""
+    node = NodeLeaf()
+    if len(elems) < 2:
+        node.extend(elems)
+    elif (choice := choose_axis(elems)) is None:
+        node.extend(elems)
+    else:
+        order, cut_after = choice
```
eric-anderson, line 223: `TODO: we could improve the performance by only computing the overlap once; not worth it for now. Most documents have moderate elements/page`

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+        for _, isopen, elem, cnt, width in gen_overlap(order):
+            if not isopen:
+                node.append(elem)
+                if elem == cut_after:
+                    node.advance()
+    return node.finalize()
```
eric-anderson, line 229: Explain in slack to me how we can need to finalize; If the width >= 0 then gen_overlap found a cut point before and after since on the last element it will generate a width of -1, which would fall under the choice is None branch.

Once I understand, add a comment explaining to future readers.

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def divide_node(node: NodeLeaf) -> NodeInner:
+    """Return replacement Node.  Assume input is leaf."""
+    inner = NodeInner()
+    for elist in node.elists:
+        subnode = cleave_elems(elist)
+        # print("....................")
```
eric-anderson, line 237:
```
if False:
   debugstuff
```
or remove.

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+        # print(subnode)
+        # print("^^^^^^^^^^^^^^^^^^^^")
+        if subnode.cansplit():
+            inner.append(divide_node(subnode))  # recursive step
+        else:
+            inner.append(subnode)
```
eric-anderson, line 243: bbox sort the remaining elements if you can't split; or document why we leave the elements in the original order.

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+    return inner
+
+
+def xycut_sorted_page(elems: ElemList) -> ElemList:
+    if len(elems) < 2:
+        return elems
+    flat = NodeLeaf().extend(elems)
+    # print("VVVVVVVVVVVVVVVVVVVV")
```
eric-anderson, line 251: `if False: ...`

lib/sycamore/sycamore/utils/xycut.py
```diff
@@ -0,0 +1,271 @@
...
+    # print(flat)
+    # print("AAAAAAAAAAAAAAAAAAAA")
+    tree = divide_node(flat)
+    return tree.to_elems()
+
+
+def xycut_sorted_elements(elements: ElemList, update_indices: bool = True) -> ElemList:
```
eric-anderson, line 258: Document why you might not want to update indices.  If you don't update indices then round-tripping the document through opensearch will revert the order.
I could imagine a preserve_original_indices, but I have no use case for that so still wouldn't include it.

eric-anderson (approved):

lib/sycamore/sycamore/data/element.py
```diff
@@ -115,6 +115,16 @@ def field_to_value(self, field: str) -> Any:
 
         return dotted_lookup(self, field)
 
+    def __lt__(self, other) -> bool:
+        sidx = self.element_index or 0
```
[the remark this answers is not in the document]
alexaryn: I'm thinking...
```
sidx = -1 if self.element_index is None else self.element_index
```

lib/sycamore/sycamore/transforms/detr_partitioner.py
```diff
@@ -223,7 +225,10 @@ def partition_pdf(
                         promote_title(page, title_candidate_elements)
                     else:
                         promote_title(page)
-                bbox_sort_page(page)
+                if sort_mode and sort_mode == "xycut":
```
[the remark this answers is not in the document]
alexaryn: I'll take a look.  The semantics of sorted() versus sort() are tricky if we want to keep efficiency.

lib/sycamore/sycamore/transforms/partition.py
```diff
@@ -14,6 +14,7 @@
 from sycamore.utils import choose_device
 from sycamore.utils.aryn_config import ArynConfig
 from sycamore.utils.bbox_sort import bbox_sort_document
+from sycamore.utils.xycut import xycut_sort_document
```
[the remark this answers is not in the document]
alexaryn: I had a similar feeling when I first put bbox_sort in.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: The reason to double characters is to make later search and replace work better.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: I opted for robustness here.  My code is fine with them reversed and ML models sometimes do strange things.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: Do you mean a simple sort by Y if we can't split, or calling the algorithm we're replacing.  Aside from the top-level call, the inputs are already sorted for the previous cut.  If we can't make a single cut, the current code leaves the order alone.  Absent any cuts, is it safe to sort arbitrarily by Y?  I could sort by Y as a pre-step.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: The reason for finalize is so the NodeLeaf will end up with 0 elements if nothing is ever appended.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: I'm copying behavior somebody added to bbox_sort.  I have no idea.

lib/sycamore/sycamore/data/element.py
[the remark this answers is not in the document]
eric-anderson: Also works. It felt kinda repetitive to me so I thought of making a function.

lib/sycamore/sycamore/utils/xycut.py
eric-anderson: I was assuming that the fix would happen in the code that calls the ML model. I don't expect other users to handle bbox out of order, so I don't see a good reason to handle it here (other than assert).

lib/sycamore/sycamore/utils/xycut.py
eric-anderson: I think that sorting should define the order, and xycut with bbox fallback is a complete ordering.   Otherwise the order is potentially random (and we've had problems in the past with failing to preserve element ordering)

lib/sycamore/sycamore/utils/xycut.py
eric-anderson: Ask them and then document in both places.  Code where no one knows why its there doesn't make sense.

eric-anderson (approved):

lib/sycamore/sycamore/data/element.py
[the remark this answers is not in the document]
alexaryn: Neither is great.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: There's more going on and I'm loath to change it right now.  It's true there's some by-catch in the exception handling, but this formulation of the code was more compact, straightforward, and performant than the bunch of if-statements I previously had.  Also, it's Pythonic to ask forgiveness, not permission.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: Added it into the f-string.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: For a given sublist of elements, the code tries to calculate the overlap once.  It's possible to do a splicing/adjusting thing, but it seems like it would be complex and, as you say, not likely significant.  If we end up with a performance problem, empirical evidence will lead us to the right code to optimize.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: If we can't split at this point, it's because there aren't enough elements.  This is the exit from the recursion.

Separately, I can try a fallback to bbox_sort if the top-level divide can't find any cuts.  That may confuse debugging because it's not clear which algorithm ran, but that's a smaller problem.

lib/sycamore/sycamore/utils/xycut.py
alexaryn: Unused.  Nuking.  Easy to revive.

PR #1304, opened 2025-05-16
Add VLMTableStructureExtractor for table structure extraction.

bsowell:
This calls an LLM to determine the cells of a table.

### DIFF
--- a/lib/sycamore/sycamore/transforms/table_structure/extract.py
+++ b/lib/sycamore/sycamore/transforms/table_structure/extract.py
@@ -472,8 +472,7 @@ class VLMTableStructureExtractor(TableStructureExtractor):
         if element.bbox is None:
             return element
 
-        width, height = doc_image.size
-        cropped_image, crop_box = _crop_bbox(doc_image, element.bbox)
+        cropped_image, _ = _crop_bbox(doc_image, element.bbox)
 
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -449,6 +457,49 @@ def parse_comparison(cls, comparison: str) -> tuple[str, Callable[[Num, Num], bo
...
+            element.table = table
+        except Exception as e:
+            logging.warning(f"Not able to parse table from HTML: {e}")
+            return element
+
+        # Convert cell bounding boxes to be relative to the original image.
+        for cell in table.cells:
+            if cell.bbox is None:
+                continue
+            cell.bbox.translate_self(crop_box[0], crop_box[1]).to_relative_self(width, height)
```
HenryL27, line 498: shouldn't all the cell bboxes be null?

HenryL27 (approved):
1 q but lgtm

lib/sycamore/sycamore/transforms/table_structure/extract.py
bsowell: Lol, yes. I guess I was just on auto-pilot. I'll go ahead and remove this. Part of me just wanted to leave it as a defense mechanism, but I can't think of a way that an llm could hallucinate bounding boxes in a way we would interpret.

PR #1309, opened 2025-05-20
Upgrade setuptools.

bsowell:
Addresses https://github.com/aryn-ai/sycamore/security/dependabot/375

### DIFF
--- a/lib/remote-processors/pyproject.toml
+++ b/lib/remote-processors/pyproject.toml
@@ -12,7 +12,6 @@ pyyaml = "^6.0.1"
 grpcio = "^1.60.0"
 cbor2 = "^5.6.0"
 sycamore-ai = "^0.1.13"
-#protobuf = "4.25.3" # See poetry-lock/pyproject.toml
 
 [tool.poetry.group.test.dependencies]
 pytest = "^7.4"
--- a/lib/remote-processors/pyproject.toml
+++ b/lib/remote-processors/pyproject.toml
@@ -12,7 +12,7 @@ pyyaml = "^6.0.1"
 grpcio = "^1.60.0"
 cbor2 = "^5.6.0"
 sycamore-ai = "^0.1.13"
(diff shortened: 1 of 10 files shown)


lib/remote-processors/pyproject.toml
```diff
@@ -12,7 +12,7 @@ pyyaml = "^6.0.1"
 grpcio = "^1.60.0"
 cbor2 = "^5.6.0"
 sycamore-ai = "^0.1.13"
-protobuf = "4.25.3" # See poetry-lock/pyproject.toml
```
karanataryn, line 15: Are we sure we don't need this anymore? Would rather delete it.

lib/sycamore/pyproject.toml
```diff
@@ -33,7 +33,7 @@ tenacity = "^8.4.1"
 diskcache = "^5.6.3"
 overrides = "^7.7.0"
 structlog = "^24.2.0"
-setuptools = ">70.0.0"
+setuptools = ">78.1.1"
```
karanataryn, line 36: Nit: > or >= ? Looks like the latter

karanataryn (approved):

lib/remote-processors/pyproject.toml
bsowell: We don't use remote-processors anymore. poetry-lock was just broken in this project with this. I can delete it, but I confess I don't know whether there will be issues.

lib/sycamore/pyproject.toml
bsowell: Sure.

lib/remote-processors/pyproject.toml
karanataryn: Ok got it. I'd suggest adding a comment stating that at least.

bsowell: I removed it. Don't think it matters much either way since I don't think anyone knows or is using this code.

PR #1311, opened 2025-05-21
improve model selection error message

HenryL27:
previously:
```
ValueError("Invalid statement: 'default'. Did not find '->', so this is assumed to be a static statement, but the model_name was not in ('table_transformer', 'deformable_detr'))
```
Now:
```
ValueError("Bad model selection: default
Invalid statement: 'default'. Did not find '->', so this is assumed to be a static statement, but the model_name was not in ('table_transformer', 'deformable_detr'))
```

### DIFF
--- a/lib/sycamore/sycamore/transforms/table_structure/extract.py
+++ b/lib/sycamore/sycamore/transforms/table_structure/extract.py
@@ -370,6 +370,7 @@ class HybridTableStructureExtractor(TableStructureExtractor):
             if "->" not in statement:
                 if statement not in cls._model_names:
                     raise ValueError(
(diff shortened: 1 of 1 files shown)


austin-aryn-ai:
How about:
```
Bad model selection: 'default'.  Valid values are ('table_transformer', 'deformable_detr')
```

Btw, are those values even documented?  They sound way too technical...

HenryL27:
> How about: ...

Issue is you can also do stuff like `model_selection="chars > 1000 -> table_transformer; deformable_detr"`

Documentation is [here](https://docs.aryn.ai/docparse/processing_options)

PR #1315, opened 2025-05-22
Add a split_pdf function.

bsowell:
I found this function useful in my own development. It splits a pdf Document (or each pdf in a DocSet) into multiple Documents, where each resulting Document has the specified number of pages, except possibly the last. Both the pdf binary and elements are split.


HenryL27 (approved):
@Soeb-aryn you were looking for something like this right?

PR #1316, opened 2025-05-26
Fix fast notebook tests

bsowell:
This PR combines two commits in an attempt to the get the fast notebook tests to pass. 

Previously the _infer_prompt method would call asyncio.run to start up an
event loop in the case that an LLM was in async mode. This causes issues in
environments like Jupyter notebooks that already have a running event loop. To
avoid conflicts, we modify this to start a new thread and run the new event
loop there.

I also switched from GPT_3_5_TURBO_INSTRUCT to GPT_4O_MINI to avoid context errors.

### DIFF
--- a/lib/sycamore/sycamore/transforms/base_llm.py
+++ b/lib/sycamore/sycamore/transforms/base_llm.py
@@ -1,3 +1,4 @@
+import threading
 from typing import Optional, Sequence, Callable, Union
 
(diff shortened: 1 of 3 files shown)


HenryL27 (approved):
Oh man, thanks for fixing this. It's been bothering me that I couldn't get something like it to work.

PR #1319, opened 2025-05-29
Handle case where query is a compound query

austin-aryn-ai:


lib/sycamore/sycamore/tests/integration/connectors/opensearch/test_opensearch_read.py
```diff
@@ -781,27 +781,31 @@ def test_result_filter_on_property(self, setup_index, os_client):
         expected = {"p3"}
         assert expected == {d.doc_id for d in retrieved_docs}
 
-        query_with_filter = {
```
eric-anderson, line 784: keep the old test, it still seems valid. Make a new test test_compound_query.

lib/sycamore/sycamore/connectors/opensearch/opensearch_reader.py
```diff
@@ -257,15 +257,39 @@ def get_doc_count_for_slice(os_client, slice_query: dict[str, Any]) -> int:
...
+    return False
 
 def add_filter_to_query(query: dict[str, Any], filter: dict[str, Any]):
-    actual_query = query["query"]
-    query["query"] = {
-        "bool": {
-            "must": [actual_query],
-            "filter": [{"terms": filter}],
+
+    if is_compound_query(query):
```
eric-anderson, line 280: Per IM discussion, don't do this, instead always wrap -- add a test for the wrapping of a compound query.

https://docs.opensearch.org/docs/latest/query-dsl/compound/index/
says you can have multiple levels "Compound queries serve as wrappers for multiple leaf or compound clauses either to combine their results or to modify their behavior."

eric-anderson (approved):

PR #1324, opened 2025-05-30
Enable pdfminer vertical text grouping.

bsowell:
Pdfminer has an option to support vertical text in its layout analysis. This commit simply enables that option, which correctly extracts vertical text.


lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
```diff
@@ -167,3 +167,35 @@ def test_detr_pdfminer_object_type(self):
         objects_text = "".join(el.text_representation for el in objects_page if el.text_representation is not None)
 
         assert lines_text == objects_text
+
+    def test_pdfminer_vertical_text(self):
```
karanataryn, line 171: Can we split this into two tests? one for boxes and one for lines. It makes it easier to understand

karanataryn (approved):

lib/sycamore/sycamore/tests/unit/transforms/test_detr_partitioner.py
bsowell: Sure. All the other pdfminer tests have both in one method, but I will split this one if you think it's clearer.

PR #1327, opened 2025-06-03
Add some useful planner processors

HenryL27:
From upstreaming
- AlwaysSummarize: makes sure the plan has a summarizedata or count on the end
- OnlyRetrieval: Optimizes plan for only caring about the documents that come back
- LogPlan: Print the plan while processing; useful debug tool
- LLMRewriteQuestion: Use a llm + prompt to rewrite the question

### DIFF
--- a/lib/sycamore/sycamore/query/strategy.py
+++ b/lib/sycamore/sycamore/query/strategy.py
@@ -196,6 +196,7 @@ class AlwaysSummarize(LogicalPlanProcessor):
         prev_result = plan.result_node
         plan.result_node += 1
         plan.nodes[plan.result_node] = SummarizeData(
+            node_type="SummarizeData",
             node_id=plan.result_node,
             description=f"Summarize the answer to the question {question}",
             inputs=[prev_result],
--- a/lib/sycamore/sycamore/query/strategy.py
+++ b/lib/sycamore/sycamore/query/strategy.py
@@ -173,7 +173,7 @@ class RemoveVectorSearchForAnalytics(LogicalPlanProcessor):
 class AlwaysSummarize(LogicalPlanProcessor):
     """
     A logical plan processor that makes sure every plan ends with a
-    summarize or a count. If the plan already ends with a summarize or count,
+    summarize data. If the plan already ends with a summarize data,
     nothing happens. If the plan ends with a sort, we drop the sort as it
     does not matter when summarizing. Then we add a summarize node to the
     end of the plan.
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/query/strategy.py
```diff
@@ -169,6 +170,119 @@ def postprocess_llm_helper(self, user_message: str) -> str:
...
+
+    def __call__(self, plan: LogicalPlan) -> LogicalPlan:
+        n = plan.nodes[plan.result_node]
+        while n.node_type in ("Sort", "SummarizeData", "LlmExtractEntity", "TopK"):
+            # SummarizeData is bad, it can fail to execute the entire pipeline so we don't get
+            # a list of documents back because there is not materialize success dir
+            # LlmExtractEntity without anything after it is just slow
+            # TopK can sometimes fail if the LLM returns garbage and doesn't help with getting
+            # the list of documents -- we just have to ignore it
+            assert len(n.inputs) == 1, f"Trailing {n.node_type} node must have exactly one input"
```
baitsguy, line 227: you'll need to update dependencies here right?

baitsguy: A more explicit way would be to choose which operators stay in vs out

lib/sycamore/sycamore/query/strategy.py
```diff
@@ -169,6 +170,119 @@ def postprocess_llm_helper(self, user_message: str) -> str:
...
+            penultimate = n.inputs[0]
+            del plan.nodes[plan.result_node]
+            plan.result_node = penultimate
+            n = plan.nodes[plan.result_node]
+        return plan
+
+
+class LogPlan(LogicalPlanProcessor):
```
baitsguy, line 235: print plan?

baitsguy: is it better to just add a .print() method to the plan object? then the .print can be called by planner who is orchestrating the processors

HenryL27: If I'm trying to assemble a strategy that does the right thing then my programming language is effectively the list of processors so one that prints the current state is useful.

lib/sycamore/sycamore/query/strategy.py
HenryL27: wdym update the dependencies? I delete the bad tail node and update the plan result_node to the one before it.

HenryL27: I considered letting it be 'RemoveTailNodesByType' or some such. The only use-case I could come up with for it was this one though, so I decided to go with the less general thing as it's clearer why you want it.

baitsguy (approved):

PR #1328, opened 2025-06-04
Add code to turn materialize off for all nodes after and incl Sort.

akarshgupta7:


baitsguy:
Also add tests please

PR #1331, opened 2025-06-04
Fix materialize commit

bohou-aryn:
Replace the dataset operators with docset operators to handle
materialize success handling.

### DIFF
--- a/lib/sycamore/sycamore/docset.py
+++ b/lib/sycamore/sycamore/docset.py
@@ -1407,21 +1407,10 @@ class DocSet:
     def llm_generate_group(self, llm: LLM, instruction: str, field: str, **kwargs):
         # Not all documents will have a value for the given field, so we filter those out.
 
-        def filter_meta(row):
-            doc = Document.from_row(row)
-            return not isinstance(doc, MetadataDocument)
+        count = self.count()
+        samples = self if count < 1000 else self.random_sample(1000.0 / count)
 
-        def extract_entities(row):
-            doc = Document.from_row(row)
-            value = doc.field_to_value(field)
-            return {"key": value}
-
-        values = self.plan.execute().filter(filter_meta).map(extract_entities).materialize()
-
-        count = values.count()
-        samples = values if count < 1000 else values.random_sample(1000.0 / count)
-
-        field_values = [sample["key"] for sample in samples.take_all()]
+        field_values = [sample.field_to_value(field) for sample in samples.take_all()]
         text = ", ".join([str(v) for v in field_values if v is not None])
 
         messages = LlmClusterEntityFormGroupsMessagesPrompt(
--- a/lib/sycamore/sycamore/grouped_data.py
+++ b/lib/sycamore/sycamore/grouped_data.py
@@ -1,5 +1,90 @@
+from typing import TYPE_CHECKING
+
 from sycamore import DocSet
 from sycamore.data import Document, MetadataDocument
+from sycamore.plan_nodes import NonCPUUser, NonGPUUser, Transform, Node
+
+if TYPE_CHECKING:
+    from ray.data import Dataset
+
+
+def filter_meta(row):
+    doc = Document.from_row(row)
+    return not isinstance(doc, MetadataDocument)
+
+
+class AggregateCount(NonCPUUser, NonGPUUser, Transform):
+    def __init__(self, child: Node, key: str):
+        super().__init__(child)
+        self._grouped_key = key
+
+    def group_udf(self, batch):
+        import numpy as np
+
+        result = {"count": np.array([len(batch["properties"])])}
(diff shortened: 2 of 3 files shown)


lib/sycamore/sycamore/docset.py
```diff
@@ -1407,21 +1407,10 @@ def top_k(
...
-        def extract_entities(row):
-            doc = Document.from_row(row)
-            value = doc.field_to_value(field)
-            return {"key": value}
-
-        values = self.plan.execute().filter(filter_meta).map(extract_entities).materialize()
-
-        count = values.count()
-        samples = values if count < 1000 else values.random_sample(1000.0 / count)
+        count = self.count()
```
eric-anderson, line 1410: I think this code is going to force execution twice. 
You don't know the count unless you execute the entire pipeline. I think that self.count() will run through the standard Executor so will materialize (if that's enabled) and at least not re-do the full effort.  

A more efficient implementation would be to do a streaming read and use the choose k out of n algorithm from https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle; so you store the 1000 documents, and process all of them exactly once.   I'd probably just put this in as a TODO.

lib/sycamore/sycamore/transforms/dataset_scan.py
```diff
@@ -22,3 +22,22 @@ def format(self):
 
     def __str__(self):
         return f"DatasetScan({self._dataset})"
+
+
+class DatasetTransplant(UnaryNode):
```
eric-anderson, line 27: I don't understand what this is for. I think what is happening is it's assuming that some other work happened to produce dataset from node, and that we return that, but I don't think this will interact correctly with our plan rewriting stuff. I think the dataset is computed earlier (and I can't quite tell exactly when it is forced to be computed. but if plan rewriting inserts some nodes, those inserts wouldn't be respected.

Probably just need to chat.

eric-anderson: docs = sycamore.init().read().map().map().filter().llm_generate_group().aggregate().
   take_all() { actual_node = rewrite(node); return [ Document(d) for d in actual_node.execute() ] }
add a node rewriter that adds a filter step right after read (for security purposes)
The filter will be ignored because llm_generate_group will have calculated the pipeline when the filter step doesn't exist.
```
def aggregate(self, child):
    ds = child.execute() # this computes a ray dataset, which binds the execution tree to what is happening with child at the time of the call
    return DataSetTransplant(ds, child)
```
 
1. node tree up to aggregate
2. transform that node tree into a ray dataset
3. add a node after the initial read
4. call take_all which ignores the node added in step 3 -- security is broken here because the filter doesn't happen

```
child = [1,2,3,4]
dataset = map(lambda x: x + 1, child)
child[3] = 17
# dataset is unchanged
```
execute() function to turn a node into a ray_dataset

```
If node -> ray form was effectively 
  ray_dataset = ray_dataset.ray_apply(lambda: node_as_function(node))
  this would let you update node and have take effect until ray runs.
  but you couldn't insert a node.
  
but the actual code looks more like:
  node = Filter(...)
  ray_dataset = ray_dataset.filter(fn = node.fn)
  now updating node.fn has no effect

nodes = Scan Node -> Entity Extraction -> GroupBy [not a docset here] -> Aggregate.
ds = agg() # DocSet
class DocSet:
     def agg(self):
           return DocSet(Node(...))

ctx = sycamore.init()
ds1 = ctx.read()
# here is where the rewrite rule would insert the filter
ds2 = ds1.ee()
# Eric guesses adding rule here will work
gb = ds2.gb()
# Here might work
ds3 = gb.collect()
# Here won't work because collect has already called execute on ds2, so the rule is ignored.
docs = ds.take_all()
```
-----
This is hard to figure out what it is doing.
Write a test with a rewrite rule that inserts a filter to remove one of the groups and verify the group vanishes with the rule enabled.

lib/sycamore/sycamore/docset.py
eric-anderson: Remember to add the TODO.

lib/sycamore/sycamore/grouped_data.py
```diff
@@ -1,82 +1,100 @@
...
+def filter_meta(row):
+    doc = Document.from_row(row)
+    return not isinstance(doc, MetadataDocument)
 
-    def count(self) -> DocSet:
-        dataset = self._docset.plan.execute()
 
-        def group_udf(batch):
-            import numpy as np
+class AggregateCount(NonCPUUser, NonGPUUser, Transform):
```
eric-anderson, line 16:
```
# TODO: share a base class between AggregateCount and AggregateCollect. The respective execute functions are identical.
```

eric-anderson (approved):

PR #1349, opened 2025-06-12
Remove default parameter for Gemini max_output_tokens.

bsowell:
It is still possible to pass this in as an extra parameter, but we have observed that in some cases we get better/more reliable output if we don't explicitly limit the output tokens, which was not possible with the previous approach.


HenryL27 (approved):
ok

PR #1354, opened 2025-06-17
Improve ChainedLLM to handle when to move to next llm in chain, add r…

austin-aryn-ai:
…etries to Gemini, remove retries from VLMTableStructureExtractor

### DIFF
--- a/lib/sycamore/sycamore/data/table.py
+++ b/lib/sycamore/sycamore/data/table.py
@@ -185,6 +185,18 @@ class Table:
         d["num_cols"] = self.num_cols
         return d
 
+    @staticmethod
+    def extract_table_block(html_str: str) -> str | None:
+        """
+        Extracts the first <table>...</table> block from the given HTML string.
+        Returns the table block as a string, or None if not found.
+        """
+        import re
+
+        r = html_str.strip()
+        match = re.search(r"<table.*?>.*?</table>", r, re.DOTALL | re.IGNORECASE)
+        return match.group(0) if match else None
+
     # TODO: There are likely edge cases where this will break or lose information. Nested or non-contiguous
     # headers are one likely source of issues. We also don't support missing closing tags (which are allowed in
     # the spec) because html.parser doesn't handle them. If and when this becomes an issue, we can consider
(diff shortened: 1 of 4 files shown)


austin-aryn-ai:
If you're OK with the approach taken here, I can add some tests.

lib/sycamore/sycamore/llms/gemini.py
```diff
@@ -152,9 +152,29 @@ async def generate_async(self, *, prompt: RenderedPrompt, llm_kwargs: Optional[d
...
+
+    @retry.Retry(
+        predicate=retry.if_transient_error,
+        initial=1.0,
+        maximum=60.0,
+        multiplier=2.0,
+        timeout=120.0,
+    )
+    async def generate_content_async(self, model, contents, config):
+        return self._client.models.generate_content(model=model, contents=contents, config=config)
```
bsowell, line 180: Doesn't this need to be `self._client.aio.models.generate_content` or need an await?

lib/sycamore/sycamore/transforms/table_structure/extract.py
```diff
@@ -535,57 +534,48 @@ class VLMTableStructureExtractor(TableStructureExtractor):
...
+
+        try:
+            res: str = self.llm.generate(prompt=prompt)
 
-                table = Table.from_html(res)
-                element.table = table
-                return True, element
+            if res.startswith("```html"):
+                res = res[7:].rstrip("`")
+            res = res.strip()
```
bsowell, line 567: One thought is that you could have the html checker handle this as well, since it's pretty common for llms to use the backtick thing.

bsowell (commented):
Check the async thing. Otherwise looks fine.

lib/sycamore/sycamore/llms/gemini.py
austin-aryn-ai: Ah yes!  I'll add 'await'.

lib/sycamore/sycamore/transforms/table_structure/extract.py
austin-aryn-ai: I don't know if we want the response check to modify the response.  Also, it's only used when the LLM is a ChainedLLM so we still need this code for non-chained LLMs.

lib/sycamore/sycamore/data/table.py
```diff
@@ -185,6 +185,18 @@ def to_dict(self) -> dict[str, Any]:
         d["num_cols"] = self.num_cols
         return d
 
+    @staticmethod
+    def extract_table_block(html_str: str):
```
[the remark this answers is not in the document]
austin-aryn-ai: Switched to using BeautifulSoup.

bsowell (approved):

PR #1358, opened 2025-06-18
adding property derivation

Soeb-aryn:


lib/sycamore/sycamore/utils/derivation_utils.py
```diff
@@ -0,0 +1,202 @@
...
+    def convert(self, value, from_unit, to_unit):
+        rate = self.find_conversion_rate(from_unit, to_unit)
+        return round(value * rate, 6)
+
+    def check(self, from_unit, to_unit, value1, value2):
+        converted_value = self.convert(value1, from_unit, to_unit)
+        return value2 and abs(converted_value - value2) / value2 < 0.05
+
+
+class UnitDerivation:
```
eric-anderson, line 42: On further inspection, I don't think this class is necessary, see comment below in PropertyDerivation.

lib/sycamore/sycamore/utils/derivation_utils.py
```diff
@@ -0,0 +1,202 @@
...
+
+            if self.properties.get(field) is None:
+                self.derive_all_conversion(field, reference_field, unit_map)
+
+
+class PropertyDerivation(UnitConverter, UnitDerivation):
+    def __init__(self, properties, metadata={}, unit_map={}):
+        self.properties = properties
+        self.metadata = metadata
+        UnitConverter.__init__(self)
```
eric-anderson, line 104: I would not do multiple inheritence unless you don't have another choice. It leads to a lot of complexities.
In this case I think it's unnecessary; you can take a unit converter as an argument.

lib/sycamore/sycamore/utils/derivation_utils.py
```diff
@@ -0,0 +1,202 @@
...
+        self.unit_map = unit_map
+        UnitDerivation.__init__(self, self.properties, self.metadata, self, self.unit_map)
+        self.group = []
+
+    def unit_group(self, group):
```
eric-anderson, line 109: I think you can write the necessary conversion code in here and remove the UnitDerivation class entirely.
```
ref_property = None
for g in group:
    assert g in self.unit_map, f"Have a property {g} without units requested for conversion"
for g in group:
     if self.properties.get(g) is not None:
          ref_property = g
          break
ref_value, ref_unit = self.properties[ref_property], self.unit_map[ref_property]
for g in group:
    if self.properties.get(g) is not None:
        continue
      # make __call__ on unit_converter do convert()
    self.properties[g] = self.unit_converter(ref_value, ref_unit, self.unit_map[g])
    if self.properties[g] is None:
        logger.warning(f"Unable to convert {ref_unit} to {self.unit_map[g]} within a unit group for properties {ref_property} to {g}")
# here's where I'd check consistency, but I don't see that happening in the existing code.
```

lib/sycamore/sycamore/utils/derivation_utils.py
```diff
@@ -0,0 +1,202 @@
...
+        temp_group = []
+        for property in group:
+            temp_group.append(property)
+        self.fill_missing_units(temp_group, self.unit_map)
+        self.group.append(temp_group)
+
+    def fill_from_formula(self, formula_string: str):
+        ops = "+-*/"
+        left, right = [part.strip() for part in formula_string.split("=")]
+        if any(op in left for op in ops):
```
eric-anderson, line 119: I'd just force people to write the second form (target = a op b) and assert the condition.

lib/sycamore/sycamore/utils/derivation_utils.py
```diff
@@ -0,0 +1,202 @@
...
+            expr, target = left, right  #  a op b = target
+        else:
+            expr, target = right, left  #  target = a op b
+        for op in ops:
```
eric-anderson, line 123: I'd force the form here as well; just require the spaces.
```
a_sym, op, b_sym = [t.strip() for t in expr.split(" ")]
assert op in ops.

if you want a cleaner error message
parts = [t.strip() for t in expr.split(" ")]
assert len(parts) == 3, f"expected <property> <op> <property> for expression, not '{expr}'"
```

lib/sycamore/sycamore/utils/derivation_utils.py
```diff
@@ -0,0 +1,202 @@
...
+
+        val_a = value_of(a_sym)
+        val_b = value_of(b_sym)
+        val_target = value_of(target)
+
+        known = sum(x is not None for x in (val_a, val_b, val_target))
+        if known != 2:
+            return
+
+        #  compute the missing one
```
eric-anderson, line 142: I'd split this into three functions. 
```
if val_target is None:
   self.properties[target] = self.deriveTarget(val_a, operator_symbol, val_b)
elif val_a is None: 
   self.properties[target] = self.deriveValA(target, operator_symbol, val_b)
elif val_b is None:
   self.properties[target] = self.deriveValB(target, operator_symbol, val_b)
```

Makes it possible to test just the derivation logic:

```
# each checks all three derivations
check("+", 10, 2, 8) 
check("-", 10, 4, -6)
check("*", 15, 3, 5)
check("/", 7, 35, 5)
 ```

Given you added a comment to several of them to work out the derivation (and I agree, those were useful comments), being able to do an exhaustive test makes sure there wasn't a bug (I don't think there was, but I could have missed it).

File: lib/sycamore/sycamore/utils/derivation_utils.py [hunk 7]
```diff
@@ -0,0 +1,202 @@
...
+            elif operator_symbol == "*":
+                val_b = val_target / val_a
+            elif operator_symbol == "/":  # target = a / b
+                val_b = val_a / val_target
+            self.properties[b_sym] = val_b
+
+        return self.properties
+
+
+def main():
```
eric-anderson, line 179: I'd move this into the test.

lib/sycamore/sycamore/tests/unit/utils/test_derivation_utils.py
```diff
@@ -0,0 +1,93 @@
...
+        self.assertEqual(self.conv.convert(123.456, "SF", "SF"), 123.456)
+
+    def test_missing_route_raises(self):
+        with self.assertRaises(ValueError):
+            self.conv.convert(1, "SF", "kg")
+
+    # ------------------------------------------------------------------ helper :check
+    def test_check_method(self):
+        """Value within ±5 % is accepted"""
+        ok = self.conv.check("SF", "SM", 100, 9.5)  # ~2.2 % error
```
eric-anderson, line 37: assert self.conv.check
and assert not self.conv.check?

PR #1367, opened 2025-06-25
Drop support for Python 3.9 and add Python 3.13.

bsowell:
This change does three things:

- Drops support for Python 3.9 and removes it from our Github Actions tests.

- Adds support for Python 3.13. This required updating a few dependencies (primarily scipy, matplotlib, and pandas).

- Adds sycamore to the default ray runtime_env. I am not sure why this showed up for me with this change. Somehow upgrading from 2.43.0 to 2.47.1 (which I did not do explicitly, but must have come when I regenerated the poetry.lock) seems to expose a ModuleNotFound error on the workers. I believe this is the correct fix, based on
https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#library-development, but can't explain what changed.


lib/poetry-lock/poetry-lock-all.sh
```diff
@@ -27,15 +27,15 @@ for i in ${tomls}; do
         (
             echo "--------------------- special casing in $i"
             cd $(dirname "$i")
```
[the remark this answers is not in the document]
bsowell: The current version of poetry no longer takes the --no-update flag.

lib/poetry-lock/poetry-lock-all.sh
```diff
@@ -27,15 +27,15 @@ for i in ${tomls}; do
...
-            poetry lock --no-update || fail "broke on special case 'poetry lock --no-update' for $i"
+            poetry lock || fail "broke on special case 'poetry lock' for $i"
```
eric-anderson, line 30: We should require poetry 2.0 here.  2.0 makes no-update the default, old poetry would change a lot more stuff.

lib/poetry-lock/pyproject.toml
```diff
@@ -16,19 +16,17 @@ fsspec = "2024.2.0"   # sycamore (via datasets)
 pytest = "7.4.0"
 guidance = "0.1.14" # They removed AzureOpenAIChat in https://github.com/guidance-ai/guidance/commit/a75896a851edf8b3cf15534af4fed8d15f277af1
 ray = ">=2.41.0" # Latest version supported on linux/aarch64 https://github.com/ray-project/ray/issues/46104
-pandas = "2.1.1" # Hardcoded into sycamore pyproject.toml. Unclear why.
+pandas = "2.3.0" # Hardcoded into sycamore pyproject.toml. Unclear why.
```
eric-anderson, line 19: Can we update the why?  I"m assuming there's a reason since you changed it.

lib/sycamore/sycamore/executor.py
```diff
@@ -50,12 +50,12 @@ def sycamore_ray_init(**ray_args) -> None:
...
 
     if "runtime_env" not in ray_args:
-        ray_args["runtime_env"] = {}
+        ray_args["runtime_env"] = {"py_modules": [sycamore]}
 
     if "worker_process_setup_hook" not in ray_args["runtime_env"]:
         # logging.error("Spurious log 0: If you do not see spurious log 1 & 2,
         # log messages are being dropped")
-        ray_args["runtime_env"]["worker_process_setup_hook"] = _ray_logging_setup
+        ray_args["runtime_env"]["worker_process_setup_hook"] = _ray_logging_setup  # type: ignore
```
eric-anderson, line 58: it's not an issue that you needed the type ignore?

eric-anderson (approved):

lib/sycamore/sycamore/executor.py
bsowell: Responded offline. The typechecker was inferring`ray_args["runtime_env"]` from the previous assignment. There is probably someway to specify it with an assertion, but it doesn't seem super critical and I was having trouble getting the syntax right.

PR #1379, opened 2025-07-02
Add support for `extract_image_format` when using remote ArynPartitioner

MarkLindblad:
Previously, when using remote `ArynPartitioner` the `extract_image_format` option would be silently ignored.


Soeb-aryn (approved):
LGTM!!

PR #1381, opened 2025-07-02
Add OCR Caching

karanataryn:
Adds the ability to have a configurable OCR Cache.

### DIFF
--- a/examples/ocr_caching_example.py
+++ b/examples/ocr_caching_example.py
@@ -1,0 +1,285 @@
+#!/usr/bin/env python3
+"""
+Example demonstrating OCR caching functionality in Sycamore.
+
+This example shows how to:
+1. Use OCR models with local caching
+2. Use OCR models with S3 caching
+3. Use cache_only mode for offline processing
+4. Monitor cache hit rates
+"""
+
+import tempfile
+from pathlib import Path
+from PIL import Image, ImageDraw, ImageFont
+
+# Import OCR models and caching utilities
+from sycamore.transforms.text_extraction.ocr_models import PaddleOcr, EasyOcr, Tesseract
+from sycamore.transforms.text_extraction.ocr_cache import get_ocr_cache_manager, set_ocr_cache_path
+
+
+def create_test_image(text="Hello World", size=(400, 100), color="white"):
+    """Create a simple test image with text."""
+    # Create image
+    img = Image.new("RGB", size, color)
+    draw = ImageDraw.Draw(img)
+
+    # Try to use a default font, fallback to default if not available
+    try:
+        font = ImageFont.truetype("/System/Library/Fonts/Arial.ttf", 24)
+    except Exception:
+        font = ImageFont.load_default()
+
+    # Calculate text position to center it
+    bbox = draw.textbbox((0, 0), text, font=font)
+    text_width = bbox[2] - bbox[0]
+    text_height = bbox[3] - bbox[1]
+    x = (size[0] - text_width) // 2
+    y = (size[1] - text_height) // 2
+
+    # Draw text
+    draw.text((x, y), text, fill="black", font=font)
+    return img
+
+
+def demonstrate_local_caching():
+    """Demonstrate local caching functionality."""
+    print("=== Local Caching Demo ===")
+
+    # Create temporary cache directory
+    with tempfile.TemporaryDirectory() as temp_dir:
+        cache_path = str(Path(temp_dir) / "ocr_cache")
+
+        # Initialize OCR model with local caching
+        ocr = PaddleOcr(cache_path=cache_path)
+
+        # Create test image
+        img = create_test_image("Hello World")
+
+        # First call - should compute and cache
+        print("First call - computing result...")
+        result1 = ocr.get_text(img)
+        print(f"Result: {result1}")
+
+        # Second call - should use cache
+        print("Second call - using cache...")
+        result2 = ocr.get_text(img)
+        print(f"Result: {result2}")
+
+        # Verify results are identical
+        assert result1 == result2, "Cached result should match original result"
+
+        # Check cache hit rate
+        hit_rate = ocr.cache_manager.get_hit_rate()
+        print(f"Cache hit rate: {hit_rate:.2%}")
+
+        print("Local caching demo completed successfully!\n")
+
+
+def demonstrate_cache_only_mode():
+    """Demonstrate cache_only mode for offline processing."""
+    print("=== Cache-Only Mode Demo ===")
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        cache_path = str(Path(temp_dir) / "ocr_cache")
+
+        # Create OCR model and populate cache
+        ocr = PaddleOcr(cache_path=cache_path)
+        img = create_test_image("Cached Text")
+
+        # Populate cache
+        print("Populating cache...")
+        ocr.get_text(img)
+
+        # Create new OCR model with cache_only=True
+        ocr_cache_only = PaddleOcr(cache_path=cache_path, cache_only=True)
+
+        # Should work with cached data
+        print("Using cache_only mode with cached data...")
+        result = ocr_cache_only.get_text(img)
+        print(f"Result: {result}")
+
+        # Should fail with new image
+        new_img = create_test_image("New Text")
+        print("Trying cache_only mode with new image...")
+        try:
+            ocr_cache_only.get_text(new_img)
+            assert False, "Should have raised CacheMissError"
+        except Exception as e:
+            print(f"Expected error: {e}")
+
+        print("Cache-only mode demo completed successfully!\n")
+
+
+def demonstrate_disable_caching():
+    """Demonstrate disable_caching mode."""
+    print("=== Disable Caching Demo ===")
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        cache_path = str(Path(temp_dir) / "ocr_cache")
+
+        # Create OCR model with caching disabled
+        ocr_disabled = PaddleOcr(cache_path=cache_path, disable_caching=True)
+        img = create_test_image("Disabled Cache Text")
+
+        # Verify cache is disabled
+        print("Verifying cache is disabled...")
+        assert ocr_disabled.cache_manager is None
+        assert ocr_disabled.disable_caching
+
+        # Process image multiple times
+        print("Processing image multiple times with caching disabled...")
+        result1 = ocr_disabled.get_text(img)
+        result2 = ocr_disabled.get_text(img)
+
+        # Results should be identical (same implementation called)
+        assert result1 == result2
+        print(f"Results are identical: {result1}")
+
+        # Create normal OCR model to verify cache is not populated
+        ocr_normal = PaddleOcr(cache_path=cache_path)
+        print("Checking that cache was not populated...")
+
+        # Should not find cached result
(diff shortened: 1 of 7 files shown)


examples/ocr_caching_example.py
```diff
@@ -0,0 +1,288 @@
...
+logger = logging.getLogger(__name__)
+
+
+def create_test_image(text="Hello World", size=(400, 100), color="white"):
+    """Create a simple test image with text."""
+    # Create image
+    img = Image.new("RGB", size, color)
+    draw = ImageDraw.Draw(img)
+
+    # Try to use a default font, fallback to default if not available
```
alexaryn, line 31: The word default used twice here is confusing.

examples/ocr_caching_example.py
```diff
@@ -0,0 +1,288 @@
...
+    draw.text((x, y), text, fill="black", font=font)
+    return img
+
+
+def demonstrate_local_caching():
+    """Demonstrate local caching functionality."""
+    logger.info("=== Local Caching Demo ===")
+
+    # Create temporary cache directory
+    with tempfile.TemporaryDirectory() as temp_dir:
```
alexaryn, line 54: It can be useful to set prefix= to avoid mysterious dirs in /tmp.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -62,15 +185,20 @@ def extract_document(
...
 
 class EasyOcr(OcrModel):
     @requires_modules("easyocr", extra="local-inference")
-    def __init__(self, lang_list=["en"], **kwargs):
+    def __init__(
+        self,
+        lang_list=["en"],
+        cache_path: Optional[str] = None,
+        cache_only: bool = False,
+        disable_caching: bool = True,
```
alexaryn, line 200: These parameters have confusing names.  In essence, the controls we want are:
- enable_cache_read: bool
- enable_cache_write: bool
- fail_on_cache_miss: bool

lib/sycamore/sycamore/transforms/text_extraction/ocr_cache.py
```diff
@@ -0,0 +1,209 @@
...
+                self._version_cache[package_name] = version
+            except importlib.metadata.PackageNotFoundError:
+                self._version_cache[package_name] = "unknown"
+        return self._version_cache[package_name]
+
+    def _image_to_hash(self, image: Image.Image) -> str:
+        """Convert PIL image to a hash string."""
+        # Convert image to bytes for consistent hashing
+        img_bytes = io.BytesIO()
+        image.save(img_bytes, format="PNG")
```
alexaryn, line 33: I would use a more deterministic image format, perhaps zero compression.

lib/sycamore/sycamore/transforms/text_extraction/ocr_cache.py
```diff
@@ -0,0 +1,209 @@
...
+        img_bytes.seek(0)
+        return hashlib.sha256(img_bytes.getvalue()).hexdigest()
```
alexaryn, line 35: You should accept a hash context object and update it here.  That allows the caller to accumulate whatever they want into the hash.  Also, hexdigest is a clumsy way to represent a hash, except perhaps for final rendering into text.  Base64 (or even 36) of` the digest bytes is better.

lib/sycamore/sycamore/transforms/text_extraction/ocr_cache.py
```diff
@@ -0,0 +1,209 @@
...
+        # Hash the kwargs
+        kwargs_hash = self._kwargs_to_hash(kwargs)
+
+        # Get versions of relevant packages
+        versions = {}
+        for package in package_names:
+            versions[package] = self._get_package_version(package)
+
+        # Create the cache key components
+        key_components = {
```
alexaryn, line 72: Just hash all of the things that matter in a deterministic order.  No need to hash hashes.  In fact, try to avoid that.

alexaryn (commented):
Will comment more later.

lib/sycamore/sycamore/transforms/text_extraction/ocr_cache.py
```diff
@@ -0,0 +1,209 @@
...
+
+def set_ocr_cache_path(cache_path: str) -> None:
+    """
+    Set the cache path for OCR operations.
+
+    Args:
+        cache_path: Path to cache directory or S3 URL (e.g., "s3://bucket/path")
+    """
+    global _ocr_cache_manager
+    _ocr_cache_manager = OcrCacheManager(cache_path)
```
alexaryn, line 209: This seems like more racy thread-unsafe code.  Can we just not allow it?

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
-# TODO: Add cache support for OCR per page
-ocr_cache = DiskCache(str(Path.home() / ".sycamore/OcrCache"))
-
 logger = logging.getLogger(__name__)
 
 
 class OcrModel(TextExtractor):
+    """Base class for OCR models with caching support."""
+
+    def __init__(self, cache_path: Optional[str] = None, cache_only: bool = False, disable_caching: bool = True):
```
alexaryn, line 32: I think this should just accept a cache manager object, or None.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
+            self.cache_manager = get_ocr_cache_manager()
+        else:
+            self.cache_manager = None
+
+        self._model_name = str(self.__class__.__name__)
+        self._package_names = self._get_package_names()
+
+    @abstractmethod
+    def _get_package_names(self) -> list[str]:
+        """Return list of package names whose versions should be included in cache key."""
```
alexaryn, line 57: Maybe two functions: one to populate the cache manager with package-version pairs, and one to provide the list of packages.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
+            **kwargs: Additional arguments passed to the OCR implementation
+
+        Returns:
+            Tuple of (text, average_font_size)
+
+        Raises:
+            CacheMissError: If cache_only=True and result not in cache
+        """
+        # If caching is disabled, directly call implementation
+        if self.disable_caching or self.cache_manager is None:
```
alexaryn, line 85: Sometimes if there are two ways to say the same thing, that's a clue.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
+
+        try:
+            cached_result = self.cache_manager.get(
+                image, self._model_name, "get_text", kwargs, self._package_names, self.cache_only
+            )
+            if cached_result is not None:
+                logger.debug(f"Cache hit for {self._model_name}.get_text")
+                return cached_result
+        except CacheMissError as e:
+            if self.cache_only:
```
alexaryn, line 96: This except should just be an (indented) else.  The exception just makes the flow harder to understand.  All that matters is if the cache lookup found None.  It's pretty clear that we could have better separation of concerns when we see that both the exception raising and handling sides both rely on the "cache only" flag.  The duplicate logging below is also a clue.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
+                logger.error(f"Cache miss for {self._model_name}.get_text: {e}")
+                raise
+            else:
+                logger.warning(f"Cache miss for {self._model_name}.get_text: {e}")
+
+        # Cache miss or cache disabled
```
alexaryn, line 102: Disabled is handled above, right?

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
+        logger.debug(f"Cache miss for {self._model_name}.get_text, computing result")
+        result = self._get_text_impl(image, **kwargs)
+
+        # Cache the result
+        self.cache_manager.set(image, self._model_name, "get_text", kwargs, self._package_names, result)
```
alexaryn, line 107: There should be a conditional controlling if we do this.  A read-only cache can be useful.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
+
+        return result
+
+    def get_boxes_and_text(self, image: Image.Image, **kwargs) -> list[dict[str, Any]]:
```
alexaryn, line 111: Is this mostly a copy of the above function?  Can we make a single function that takes the function name and a callable and wraps caching around anything?

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
+        if self.disable_caching or self.cache_manager is None:
+            return self._get_boxes_and_text_impl(image, **kwargs)
+
+        try:
+            cached_result = self.cache_manager.get(
+                image, self._model_name, "get_boxes_and_text", kwargs, self._package_names, self.cache_only
+            )
+            if cached_result is not None:
+                logger.debug(f"Cache hit for {self._model_name}.get_boxes_and_text")
+                assert isinstance(cached_result, list), f"Cached result is not a list: {type(cached_result)}"
```
alexaryn, line 135: Do we really need this assert?

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -18,22 +23,145 @@
...
+                cached_result = [
+                    {"bbox": BoundingBox(*dict_value["bbox"]), **{k: v for k, v in dict_value.items() if k != "bbox"}}
+                    for dict_value in cached_result
+                ]
```
alexaryn, line 139: It's kinda ugly when the object we store in the cache isn't directly what we can return to users.  In this case, I think we can hide/abstract some of it.  Basically, we just need a simplify/reconstruct helper.  Since this varies per-function, I don't think we can push it down into the cache module.

lib/sycamore/sycamore/transforms/text_extraction/ocr_models.py
```diff
@@ -156,16 +293,20 @@ class LegacyOcr(OcrModel):
     """Legacy behavior using Tesseract for text and EasyOcr for tables."""
 
     @requires_modules(["easyocr", "pytesseract"], extra="local-inference")
-    def __init__(self):
-        self.tesseract = Tesseract()
-        self.easy_ocr = EasyOcr()
+    def __init__(self, cache_path: Optional[str] = None, cache_only: bool = False, disable_caching: bool = True):
+        super().__init__(cache_path=cache_path, cache_only=cache_only, disable_caching=disable_caching)
+        self.tesseract = Tesseract(cache_path=cache_path, cache_only=cache_only, disable_caching=disable_caching)
+        self.easy_ocr = EasyOcr(cache_path=cache_path, cache_only=cache_only, disable_caching=disable_caching)
```
alexaryn, line 299: This would be less messy if we just passed a cache manager object around.

PR #1389, opened 2025-07-09
add limit llm operations and require query database processors

HenryL27:


lib/sycamore/sycamore/query/strategy.py
```diff
@@ -229,6 +229,70 @@ def __call__(self, plan: LogicalPlan, **kwargs) -> LogicalPlan:
...
+    less accurate.
+
+    Args:
+        types: List of node types to limit; e.g. "LlmFilter", "SummarizeData"
+        limit: How many documents to limit to
+        message: Optional message to add to the description of the added limit node. Default is
+            "Limit the number of documents going through the following llm operation for interactivity"
+    """
+
+    def __init__(self, types: list[str], limit: int, message: Optional[str] = None):
```
eric-anderson, line 247: See if this looks better as `message: str = "Limit the number of documents going through the following llm operation for interactivity"`
I think it will, but I'm not sure what auto-formatting will do to it.

lib/sycamore/sycamore/query/strategy.py
```diff
@@ -229,6 +229,70 @@ def __call__(self, plan: LogicalPlan, **kwargs) -> LogicalPlan:
...
+        )
+
+    def __call__(self, plan: LogicalPlan, **kwargs) -> LogicalPlan:
+        new_nodes: dict[int, Node] = {}
+        new_nodes_i = 0
+        old_to_new_mapping: dict[int, int] = {}
+        for old_i in sorted(plan.nodes.keys()):
+            node = plan.nodes[old_i]
+            new_inputs = [old_to_new_mapping[inp] for inp in node.inputs]
+            # Don't add a limit after another limit. Note - this is close to but not correct.
```
eric-anderson, line 261: `# TODO: Fix this to set the limit to min(explicit limit, applied limit).  ... (rest of comment)`

On second consideration, why not just fix this?  In the case of a single predecessor limit node, it should just be `new_nodes[new_inputs[0]].num_records = min(...)`
You can leave the TODO on multiple predecessors and some have a limit since that won't happen right now. Probably emit a warning "unable to apply limit bounding"

lib/sycamore/sycamore/query/strategy.py
```diff
@@ -229,6 +229,70 @@ def __call__(self, plan: LogicalPlan, **kwargs) -> LogicalPlan:
...
+        self._types = types
+        self._limit = limit
+        self._message = (
+            message or "Limit the number of documents going through the following llm operation for interactivity"
+        )
+
+    def __call__(self, plan: LogicalPlan, **kwargs) -> LogicalPlan:
+        new_nodes: dict[int, Node] = {}
+        new_nodes_i = 0
+        old_to_new_mapping: dict[int, int] = {}
```
eric-anderson, line 257: The old_to_new stuff is awkward.
```
# TODO: write a helper function to be able to transform plans. It's messy because we don't just have pointers to nodes, but have indexes that identify things.  Not sure what the right API is. 1 possibility is translate to the natural datastructure, and then people can recursively manipulate (like we do for sycamore plans). Other possibility is an api for walking and manipulating -- e = PlanEditor(plan); for n in e.walk_preorder: if n.type is not Limit: e.add_after(n, Limit(...))
```

eric-anderson (approved):

PR #1391, opened 2025-07-14
Add aggregation interface

HenryL27:


lib/sycamore/sycamore/transforms/summarize.py
```diff
@@ -634,3 +635,26 @@ class Summarize(NonCPUUser, NonGPUUser, Map):
 
     def __init__(self, child: Node, summarizer: Summarizer, **kwargs):
         super().__init__(child, f=summarizer.summarize, **kwargs)
+
+
+class CollectToSummaryDoc(Aggregation):
```
eric-anderson, line 640: For discussion: Does order matter? Aggregation doesn't guarantee the order of combine, so the assumption is that combine(a, b) == combine(b, a) or the results are non-deterministic. Unfortunately, you can't even make it a dictionary because the python maintains insertion order, so you can have a == b and a.values() != b.values().

For this specific case, best I can come up with is sort by docid in both accumulate and combine.  For accumulation, we probably need a test version that checks that except for the resulting docid, combine order doesn't change the equality, and accumulate is order independent.  These sorts of non-determinism will be very difficult to debug.

```
>>> a,b={},{}
>>> a[1] = 1
>>> a[2] = 2
>>> b[2] = 2
>>> b[1] = 1
>>> b.keys()
dict_keys([2, 1])
>>> b.values()
dict_values([2, 1])
>>> a == b
True
>>> a.keys()
dict_keys([1, 2])
>>> a.values()
dict_values([1, 2])
>>> assert a == b and a.values() != b.values()
```

lib/sycamore/sycamore/transforms/aggregation.py
```diff
@@ -0,0 +1,259 @@
...
+
+if TYPE_CHECKING:
+    from ray.data import Dataset
+    import numpy as np
+
+
+class AggregationNode(UnaryNode):
+    def __init__(
+        self,
+        child: Optional[Node],
```
eric-anderson, line 18: You have an Aggregation class which defines name ... zero_factory. I'd just take such an object as an argument.

lib/sycamore/sycamore/transforms/aggregation.py
```diff
@@ -0,0 +1,259 @@
...
+                final = self._finalize(partial)
+                ret.extend(update_lineage(from_docs=[partial], to_docs=[final]))
+                ret.append(final)
+        ret.extend(extra_metadata)
+        return ret
+
+
+class Aggregation:
+    def __init__(
+        self,
```
eric-anderson, line 171: I think the common case is going to be people inheriting from Aggregation, so I'd make that case easier.
Also I wouldn't use _ for the interface since it's effectively public.
```
name: Optional[str] = None,
accumulate_docs: Optional[Callable[list[Documnet]], Document] = None,
...
self._name = name or self.__class__.__name__
if accumulate_docs is not None:
    self.accumulate = lambda s, ds: accumulate_docs(ds)
...
def accumulate(self, in: list[Document]) -> Document:
    raise UnimplementedError("Must define accumulate function")
...
def finalize(self, d: Document) -> Document:
    return d

def zero_factory(self) -> Document:
    return Document()
```

lib/sycamore/sycamore/transforms/aggregation.py
```diff
@@ -0,0 +1,259 @@
...
+        import pickle
+
+        doc_n_meta = row[self._name]
+        doc = doc_n_meta["doc"]
+        meta = pickle.loads(doc_n_meta["meta"])
+        return [{"doc": doc}] + [m.to_row() for m in meta]
+
+    def execute(self, **kwargs) -> "Dataset":
+        from ray.data.aggregate import AggregateFnV2
+        import pyarrow as pa
```
eric-anderson, line 52: Can we keep these without shortnames? They're both only used once so the shortening doesn't save anything, and it's more confusing to read.

lib/sycamore/sycamore/transforms/aggregation.py
```diff
@@ -0,0 +1,259 @@
...
+        dataset = self.child().execute()
+
+        class RayAggregation(AggregateFnV2):
+            def __init__(
+                self,
+                syc_agg: AggregationNode,
+                name: str,
+                zero_factory: Callable[[], Document],
+            ):
+                # Idk why I couldn't do super().__init__(...)
```
eric-anderson, line 65: Add something on what error you got?  This does feel weird.

lib/sycamore/sycamore/transforms/aggregation.py
```diff
@@ -0,0 +1,259 @@
...
+                def real_zero_factory():
+                    return {"doc": zero_factory(), "meta": pickle.dumps([])}
+
+                AggregateFnV2.__init__(self, name, real_zero_factory, on=None, ignore_nulls=True)
+                self._syc_agg = syc_agg
+
+            def aggregate_block(self, block: Union[pa.Table, pd.DataFrame]):
+                docs = [
+                    Document.deserialize(
```
eric-anderson, line 74: For discussion: Add comment on why this is necessary.

lib/sycamore/sycamore/transforms/aggregation.py
```diff
@@ -0,0 +1,259 @@
...
+
+        ray_agg = RayAggregation(self, self._name, self._zero_factory)
+        ds = dataset.map(self._to_key_val).groupby("key").aggregate(ray_agg).flat_map(self._unpack)
+        return ds
+
+    def local_execute(self, all_docs: list[Document], do_combine: bool = True) -> list[Document]:
+        import random
+
+        documents, metadata = split_data_metadata(all_docs)
+        split_docs: dict[str, list] = {}
```
eric-anderson, line 132: key_to_docs  (makes it obvious what the keys and values are.

lib/sycamore/sycamore/transforms/aggregation.py
```diff
@@ -0,0 +1,259 @@
...
+                split.append(d)
+            else:
+                split_docs[key] = [d]
+
+        ret: list[Union[Document, MetadataDocument]] = metadata  # type: ignore
+        extra_metadata: list[MetadataDocument] = []
+        with ThreadLocal(ADD_METADATA_TO_OUTPUT, extra_metadata):
+            for key, split in split_docs.items():
+                if do_combine and len(split) > 1:
+                    # This path is mostly for testing. Otherwise combine is not exercised
```
eric-anderson, line 146: This is probably where
```
# Determinism requires that combine is commutative, associative, and has a zero; and that accumulate is order independent.
# Non-crashiness requires it to handle those orders without failing.
# We check a subset of the conditions here and will expand over time.
# TODO: check that the result of combine(a, b) == combine(b, a);  that combine(a, combine(b, c)) = combine(combine(a, b), c); and that combine(a, zero) == a ==  combine(zero, a)
# TODO: check that accumulate(docs) == accumulate(shuffle(docs))
```
should go.

lib/sycamore/sycamore/transforms/aggregation.py
```diff
@@ -0,0 +1,259 @@
...
+            self._name,
+            self._accumulate,
+            self._combine,
+            self._finalize,
+            group_key_fn=group_key_fn,
+            zero_factory=self._zero_factory,
+        )
+
+
+class Reduce(UnaryNode):
```
eric-anderson, line 201: for discussion: why isn't this implemented as an aggregation?  This should be an aggregation (combine to single list) + finalize (call the reduce function on the single list).  Then you don't need the ray and local mode implementations and you can inherit the correctness checking in aggregate.

lib/sycamore/sycamore/tests/unit/transforms/test_aggregation.py
```diff
@@ -0,0 +1,184 @@
...
+        assert counts.fin_calls > 0
+
+        real_out, meta_out = split_data_metadata(out_docs)
+        assert len(real_out) == 1
+        assert real_out[0].properties["sum"] == 21
+
+        assert any("keep" in md.metadata for md in meta_out)
+        assert_lineage(from_docs=Common.docs, to_docs=real_out[0], metadata=meta_out)
+
+    def test_aggregation_gouping(self):
```
eric-anderson, line 106: grouping

eric-anderson (approved):

lib/sycamore/sycamore/transforms/aggregation.py
HenryL27: seems like it works now. Not sure what I was doing wrong.

lib/sycamore/sycamore/transforms/summarize.py
HenryL27: I can just stick the sort in the finalize, right? Should be logically equivalent and glancing at the [algorithm](https://www.wild-inter.net/publications/munro-wild-2018) it should be asymptotically similar right?

lib/sycamore/sycamore/transforms/aggregation.py
HenryL27: bc I didn't think of it.

PR #1401, opened 2025-07-24
Add code for suggest properties.

akarshgupta7:
* Add batch elements step through strategy.

* Add prompt for doing entity extraction with just elements.

* Add suggest_properties method to Docset class.

* Add logic to create schema from the intersection of the field names across documents.

### DIFF
--- a/lib/sycamore/sycamore/transforms/property_extraction/extract.py
+++ b/lib/sycamore/sycamore/transforms/property_extraction/extract.py
@@ -4,7 +4,7 @@ import logging
 
 from sycamore.data.document import Document
 from sycamore.plan_nodes import Node
-from sycamore.schema import Schema
+from sycamore.schema import Schema, SchemaField
 from sycamore.transforms.map import MapBatch
 from sycamore.transforms.property_extraction.strategy import SchemaPartitionStrategy, StepThroughStrategy
 from sycamore.llms.llms import LLM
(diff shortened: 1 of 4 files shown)


lib/sycamore/sycamore/transforms/property_extraction/extract.py
```diff
@@ -78,3 +78,71 @@ async def extract_schema_partition_from_document(self, document: Document, schem
...
+                    # Only append if type matches and value is not None and not already present
+                    if example is not None and v_type == result_dict[k]["type"]:
+                        if example not in result_dict[k]["examples"]:
+                            result_dict[k]["examples"].append(example)
+                    elif v_type != result_dict[k]["type"]:
+                        _logger.warning(
+                            f"Type mismatch for field '{k}': {v_type} vs {result_dict[k]['type']}. "
+                            "Skipping this field."
+                        )
+        return list(result_dict.values())
```
HenryL27, line 148: why not just return result_dict?

lib/sycamore/sycamore/transforms/property_extraction/merge_schemas.py
```diff
@@ -0,0 +1,56 @@
...
+                    "description": field.description,
+                    "examples": field.examples,
+                }
+            else:
+                if field.field_type != merged_fields[name]["type"]:
+                    continue
+                merged_fields[name]["examples"].extend(field["examples"])
+        if temp:
+            field_names.append(temp)
+    common_field_names = set.intersection(*map(set, field_names))
```
HenryL27, line 41: you could make temp a set to begin with

HenryL27 (approved):

lib/sycamore/sycamore/transforms/property_extraction/extract.py
akarshgupta7: I am creating a schema for each document in the function where this function is called where fields is a list hence returning a list. In this function, I created a dictionary to deduplicate schema fields across BatchElements in a document

PR #1403, opened 2025-07-29
[extract] various improvements

HenryL27:
handle lists, add utils for seeing llm utilization, add schemaUpdateStrategy interface, image prompting ability

### DIFF
--- a/lib/sycamore/sycamore/llms/llms.py
+++ b/lib/sycamore/sycamore/llms/llms.py
@@ -210,8 +210,8 @@ class LLM(ABC):
             "model": self._model_name,
             "temperature": kwargs.get("temperature", None),
             "usage": {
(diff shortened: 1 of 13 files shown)


lib/sycamore/sycamore/llms/llms.py
```diff
@@ -210,8 +210,8 @@ def get_metadata(self, kwargs, response_text, wall_latency, in_tokens, out_token
             "model": self._model_name,
             "temperature": kwargs.get("temperature", None),
             "usage": {
-                "completion_tokens": in_tokens,
-                "prompt_tokens": out_tokens,
+                "completion_tokens": out_tokens,
```
akarshgupta7, line 213: Good catch!

PR #1405, opened 2025-07-31
Update Table to HTML Modification

karanataryn:


lib/sycamore/sycamore/data/table.py
```diff
@@ -411,19 +416,47 @@ def to_html(self, pretty=False, wrap_in_html=False, style=DEFAULT_HTML_STYLE):
...
+        for i in range(len(header_rows)):
+            if header_rows[i] != i:
+                break
+            max_header_prefix_row = i
+
+        # Create thead and tbody sections if we have headers
+        has_headers = max_header_prefix_row >= 0
+        if has_headers:
+            thead = ET.SubElement(table, "thead")
+            tbody = ET.SubElement(table, "tbody")
```
bsowell, line 453: It looks like this is taken from the to_pandas method, right? IIRC, the reason I didn't do this initially was that technically you can have some cells in a row that are marked as a header and some that are not. Pandas just doesn't allow that, so I didn't worry about it there, but I do wonder what happens here in that case. If you have a non-header cell in a <thead> section, does it break anything? 

For that matter, do thead and tbody do anything? I guess they are just additional tags you can style? 

This is probably academic. I'm fine with this.

bsowell (approved):

lib/sycamore/sycamore/data/table.py
karanataryn: Discussed offline, going ahead with this for now.

PR #1410, opened 2025-08-01
Add integration test for schema extract.

akarshgupta7:


lib/sycamore/sycamore/tests/integration/transforms/property_extraction/test_schema_extract.py
```diff
@@ -0,0 +1,126 @@
...
+        docs = [doc_0, doc_1]
+        schema_ext = SchemaExtract(
+            None,
+            step_through_strategy=BatchElements(batch_size=50),
+            llm=FakeLLM(),
+            prompt=FakeExtractionPrompt(),
+        )
+        extracted_docs = schema_ext.run(docs)
+        context = sycamore.init()
+        ds = context.read.document(extracted_docs).reduce(intersection_of_fields)
```
HenryL27, line 112: btw, you can rewrite this as 
```python
context = sycamore.init()
read_ds = context.read.document(docs)
ds = Docset(context, SchemaExtract(read_ds.plan, step_through...)).reduce(int_of_fields)
```
Then, if we ever parametrize the prompt in the docset method this can turn into just call that

HenryL27 (approved):
the unit test still exists right?

PR #1418, opened 2025-08-06
[extract] put richproperties in entity_metadata and plain python props in entity

HenryL27:


lib/sycamore/sycamore/transforms/property_extraction/utils.py
```diff
@@ -23,3 +24,25 @@ def create_named_property(prop_data: dict[str, Any], n_examples: Optional[int] =
...
+        for k in ob1.value.keys():
+            if k in ob2.value:
+                rd[k] = stitch_together_objects(ob1.value[k], ob2.value[k])
+            else:
+                rd[k] = ob1.value[k]
+        for k in ob2.value.keys():
+            if k not in rd:
+                rd[k] = ob2.value[k]
+        ret = ob1.model_copy()
+        ret.value = rd
```
akarshgupta7, line 46: missing return statement?

HenryL27: hah! I guess the existing tests are insufficient

PR #1420, opened 2025-08-06
Add support for document filtering when reading a DocSet

austin-aryn-ai:


lib/sycamore/sycamore/connectors/aryn/ArynReader.py
```diff
@@ -21,6 +21,25 @@
...
+    doc_ids: Optional[list[str]] = None
+    sample_ratio: Optional[float] = None
+    seed: Optional[int] = None
+
+    def select(self, doc_list: list[str]) -> list[str]:
+        if self.doc_ids is not None:
+            return [doc_id for doc_id in doc_list if doc_id in self.doc_ids]
+        elif self.sample_ratio is not None:
+            if self.seed is not None:
+                import random
```
akarshgupta7, line 35: why is random defined inside the if statement? it will throw an error if seed is None

PR #1425, opened 2025-08-09
Add support for updating properties only

austin-aryn-ai:


lib/sycamore/sycamore/connectors/aryn/ArynWriter.py
```diff
@@ -21,9 +21,15 @@ def __init__(self, aryn_url: str, api_key: str, **kwargs):
 
 @dataclass
 class ArynWriterTargetParams(BaseDBWriter.TargetParams):
-    def __init__(self, docset_id: Optional[str] = None, update_schema: bool = False):
+    def __init__(
+        self,
+        docset_id: Optional[str] = None,
+        update_schema: bool = False,
```
alexaryn, line 27: Might slap a * argument before update_schema or the new one to force keyword arguments, as positional can get confusing.

lib/sycamore/sycamore/writer.py
```diff
@@ -821,6 +821,7 @@ def aryn(
         aryn_api_key: Optional[str] = None,
         aryn_url: Optional[str] = None,
         update_schema: bool = True,
+        only_properties: bool = False,
```
alexaryn, line 824: I wonder if it's too late to add a * argument here somewhere above.

alexaryn (approved):

lib/sycamore/sycamore/connectors/aryn/ArynWriter.py
austin-aryn-ai: That's a good idea.

lib/sycamore/sycamore/writer.py
austin-aryn-ai: I don't think it's too late.

PR #1429, opened 2025-08-11
[extract] more fixes

HenryL27:
when there are existing properties already on the document (which might be a model_dumped version of the struct we're using), try to handle that


akarshgupta7 (approved):
Looks good

PR #1434, opened 2025-08-14
Change Schema serialization to support backwards compatibility.

bsowell:
This ensures that old versions of the code can read serialized schemas, as long as they don't make use of any of the new incompatible features.

### DIFF
--- a/lib/sycamore/sycamore/schema.py
+++ b/lib/sycamore/sycamore/schema.py
@@ -12,6 +12,8 @@ from pydantic import (
     ValidatorFunctionWrapHandler,
     WrapValidator,
     ValidationError,
+    model_serializer,
+    SerializerFunctionWrapHandler,
 )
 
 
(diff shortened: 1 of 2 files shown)


lib/sycamore/sycamore/schema.py
```diff
@@ -294,3 +296,35 @@ def render_flattened(self) -> str:
...
+                return nxt(self)
+
+            set_fields = p.type.model_fields_set
+            if not set_fields.issubset({"type", "default", "description", "examples"}):
+                return nxt(self)
+
+            fields.append(
+                {
+                    "name": p.name,
+                    "property_type": p.type.custom_type if p.type.type == DataType.CUSTOM else p.type.type,
```
HenryL27, line 323: does the else want to be `p.type.type.value`?

HenryL27 (approved):

PR #1442, opened 2025-08-22
[Extract] Add validators

HenryL27:
At the moment I've hooked it into the update_schema traversal. It should really get its own traversal though (that will let me do things like retry). See previous points about abstracting that and getting confuddled


bsowell (approved):
LGTM as far as I can tell.

PR #1445, opened 2025-08-24
Add Ability to Disable PIL Image Size

karanataryn:


lib/sycamore/sycamore/data/element.py
```diff
@@ -154,6 +154,8 @@ def __init__(
         self.data["properties"]["image_size"] = image_size
         self.data["properties"]["image_mode"] = image_mode
         self.data["properties"]["image_format"] = image_format
+        if kwargs.pop("max_image_size", None) is not None:
+            Image.MAX_IMAGE_PIXELS = kwargs.pop("max_image_size")
```
bsowell, line 158: Isn't it a little strange to change a global (or at least module-level) variable in the element constructor? When would I expect this to be set? What would happen if two different serialized elements had different values for max_image_size? Seems like it would be better for this to be some kind of job-level config, no?

karanataryn: I see your point, where would be the best place to set it then? Within Sycamore .context()?

PR #1456, opened 2025-08-29
[extract] Add retries for validators

HenryL27:
also rework the schema formatting in the prompt to take desscriptions of non-leaf nodes and invalid guesses from the properties


lib/sycamore/sycamore/utils/zip_traverse.py
```diff
@@ -44,12 +44,14 @@ def zip_traverse(
...
         key_sets = [set(ki) for ki in key_iters if ki is not None]
+        if all(len(ks) == 0 for ks in key_sets) and None in key_iters:
+            key_sets.append({None})
     if intersect_keys:
         keys = set.intersection(*key_sets)
     else:
         keys = set.union(*key_sets)
 
-    for k in keys:
+    for k in sorted(keys):  # type: ignore  # Hashable is not necessarily sortable, but strings and ints are
```
bsowell, line 54: Not a huge deal, but then would it be better to change the type to `str | int`.

bsowell: Also do we always want to sort? Will this lead us to reorder the keys in the properties or schema or is it only going to impact the order in which we see the keys??

lib/sycamore/sycamore/transforms/property_extraction/extract.py
```diff
@@ -23,10 +23,13 @@
 from sycamore.utils.threading import run_coros_threadsafe
 from sycamore.transforms.property_extraction.utils import stitch_together_objects, dedup_examples
 from sycamore.transforms.property_extraction.attribution import refine_attribution
+from sycamore.transforms.property_extraction.prompts import format_schema_v2
 from sycamore.utils.zip_traverse import zip_traverse
 
 _logger = logging.getLogger(__name__)
 
+MAX_RETRIES = 10
```
bsowell, line 31: 10 seems like it's probably too high as a default, given how long LLM calls take. Maybe set this to something like 3

bsowell (approved):
I'm not incredibly confident there's not something I didn't catch lurking in here, but mostly seems fine to me.

PR #1460, opened 2025-09-06
Improve extract_json and rendered prompts

eric-anderson:
Handle ```json ``` inside of a string in the returned JSON.
add to_human_readable for rendered prompts.

### DIFF
--- a/lib/sycamore/sycamore/utils/extract_json.py
+++ b/lib/sycamore/sycamore/utils/extract_json.py
@@ -4,28 +4,42 @@ from json import JSONDecodeError
 from typing import Any
 
 
-def extract_json(payload: str) -> Any:
+def extract_json(payload: str, verbose=False) -> Any:
     """Given the provided payload, extract the JSON block from it."""
 
+    orig_payload = payload
+    if verbose:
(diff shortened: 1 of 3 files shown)


lib/sycamore/sycamore/utils/extract_json.py
```diff
@@ -4,28 +4,42 @@
...
+def extract_json(payload: str, verbose=False) -> Any:
     """Given the provided payload, extract the JSON block from it."""
 
+    orig_payload = payload
+    if verbose:
+        print(f"--------------------Extract Json--------------------\n{payload}\n")
+    # It is possible that the LLM response includes a code block with JSON data.
+    # Pull the JSON content out from it.
+    pattern = r"(^|.*\n)```json\n([\s\S]*?\n)```"
+    match = re.match(pattern, payload, re.DOTALL)
```
HenryL27, line 16: I might be reading this wrong but would this match if the llm put backticks with a newline inside the string in the json payload? I guess the newline would have to be escaped itself for the json to be valid so no. ok

HenryL27 (approved):

PR #1472, opened 2025-09-22
Torch speedups from benchmarking: low hanging fruit.

alexaryn:
These dropped the naive case for DETR partitioning from ~18 seconds to ~14.

Using torch.autocast() with the DETR table model caused wrong answers; so, we don't use it there.


karanataryn (approved):
Looks good to me.

PR #1474, opened 2025-09-24
Change `HybridTableStructureExtractor` to default to `deformable_detr` instead of `table_transformer`

MarkLindblad:


HenryL27 (approved):
lgtm but you got a bunch of test fails

PR #1475, opened 2025-09-25
Update bbox_sort and xycut to support an optional reading order.

bsowell:
The default remains "ltr", but if you pass "rtl", it will sort elements. based on that reading order.

### DIFF
--- a/lib/sycamore/sycamore/utils/bbox_sort.py
+++ b/lib/sycamore/sycamore/utils/bbox_sort.py
@@ -6,9 +6,12 @@ TODO:
 - handle bbox not (always) present
 """
 
-from typing import Optional
+from typing import Literal, Optional
 
 from sycamore.data import Element
+
+
+ReadingDirection = Literal["ltr", "rtl"]
 
 
 def elem_top_left(elem: Element) -> tuple:
@@ -18,11 +21,26 @@ def elem_top_left(elem: Element) -> tuple:
     return (0.0, 0.0)
 
 
+def elem_top_right(elem: Element) -> tuple:
+    bbox = elem.data.get("bbox")
+    if bbox:
+        return (bbox[1], -bbox[0])
+    return (0.0, 0.0)
+
+
 def elem_left_top(elem: Element) -> tuple:
     bbox = elem.data.get("bbox")
     if bbox:
         left = int(5 * bbox[0])  # !!! quantize
         return (left, bbox[1])
+    return (0.0, 0.0)
+
+
+def elem_right_top(elem: Element) -> tuple:
+    bbox = elem.data.get("bbox")
+    if bbox:
+        left = int(5 * bbox[0])
+        return (-left, bbox[1])
     return (0.0, 0.0)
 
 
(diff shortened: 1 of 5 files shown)


lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -6,18 +6,28 @@
 - handle bbox not (always) present
 """
 
-from typing import Optional
+from typing import Literal, Optional
 
 from sycamore.data import Element
 
 
+ReadingDirection = Literal["ltr", "rtl"]
```
alexaryn, line 14: I dislike the use of Literal, as mypy tends to complain vociferously if one writes a wrapper function that blindly passing the string through.  We could consider having a boolean "left_to_right", unless we plan to support vertical stuff like old Chinese.  Even then, we could make the horizontal and vertical orthogonal.

lib/sycamore/sycamore/utils/bbox_sort.py
```diff
@@ -6,18 +6,28 @@
...
     bbox = elem.data.get("bbox")
     if bbox:
         return (bbox[1], bbox[0])
     return (0.0, 0.0)
 
 
+def elem_top_right(elem: Element) -> tuple:
+    bbox = elem.data.get("bbox")
+    if bbox:
+        return (bbox[1], -bbox[0])
```
alexaryn, line 27: One of my ideas was to pass through some sort of coordinate transform function (which could be the identity).  Also, I wasn't sure if we wanted negative numbers or we wanted 1.0 - bbox[0].

The other question I had was if it was OK to falsify the data, or if it was better just to invert the sorting and comparison logic.

lib/sycamore/sycamore/utils/element_sort.py
```diff
@@ -43,26 +49,28 @@ def collect_pages(elems: list[Element]) -> list[list[Element]]:
 
 # This allows specification of the "page sort" building block either
 # as a string ("bbox", "xycut") or Callable.
-SortSpec = Union[Optional[str], Callable[[list[Element]], None]]
+SortSpec = Union[Optional[str], PageSorter]
 
 
-def sort_page(elems: list[Element], *, mode: SortSpec = None) -> None:
+def sort_page(elems: list[Element], *, mode: SortSpec = None, reading_direction: ReadingDirection = "ltr") -> None:
```
alexaryn, line 55: It seems like if we're going to the trouble of plumbing a new parameter through, it could be more of a context/options object, rather than this ad-hoc value.

lib/sycamore/sycamore/utils/bbox_sort.py
alexaryn: I think there are some scaling issues that would be improved by the ability to apply a transformation matrix.  For instance, documents with asymmetric or otherwise weird margins, not to mention slight rotations.

My concern about modifying the data in-place is that we may end up reporting bboxes in a different coordinate space than the original document.  Drawing boxes in the UI would look wrong.  We already have this problem for auto-rotation.

lib/sycamore/sycamore/utils/element_sort.py
alexaryn: This is totally about making the next change easier.  I'm often a fan of "install the plumbing once".

PR #1484, opened 2025-09-26
extract bugfix

HenryL27:
this was crashing when doubly nested objects were a thing


lib/sycamore/sycamore/transforms/property_extraction/extract.py
```diff
@@ -159,6 +159,10 @@ async def extract_schema_partition_from_element_batch(
                     continue
                 if v_new is not None and v_new.type is not DataType.OBJECT:
                     p_work.value[k] = v_new
+                if v_new is not None and v_new.type is DataType.OBJECT and v_work is None:
```
karanataryn, line 162: I think we might need a comment or two here explaining the need for this (how this conditional could be hit)

PR #1488, opened 2025-10-01
PDF-to-image iterator that cleans up if interrupted.

alexaryn:
This is both a context manager and an iterator.

Newer version spawns zero threads and it has the core logic in one function for easier understanding.  It's more efficient in buffer handling and has protection against hangs.

Set bufsize to zero in Popen, as we're doing our own buffered IO already.


lib/sycamore/sycamore/tests/manual/pdf2image.py
```diff
@@ -0,0 +1,41 @@
+import sys
```
eric-anderson, line 1: document why a manual test and not an integration test. Or switch to an integration test.

lib/sycamore/sycamore/tests/manual/pdf2image.py
```diff
@@ -0,0 +1,41 @@
...
+        datefmt="%Y%m%d_%H%M%S",
+    )
+    fn = args.pop(0)
+    full(fn)
+    bail(fn)
+    return 0
+
+
+def full(fn: str) -> None:
+    with PdfToImageFiles(pdf_path=fn, file_dir="/tmp") as gen:
```
eric-anderson, line 22: write the files into a TemporaryDirectory, and verify that everything is cleaned up after the iterator exits but before the tempdir does.

lib/sycamore/sycamore/tests/manual/pdf2image.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def bail(fn: str) -> None:
+    with PdfToImageFiles(pdf_path=fn, file_dir="/tmp") as gen:
+        cnt = 0
+        for p in gen:
+            print(cnt, p, flush=True)
+            p.unlink()
+            cnt += 1
+            if cnt > 20:
```
eric-anderson, line 34: document the minimum of 21 pages requirement to test early exit or make it an argument or get the # pages in the pdf and bail out at max(0, min(20, npages - 5))

lib/sycamore/sycamore/tests/manual/pdf2image.py
```diff
@@ -0,0 +1,41 @@
...
+                break
+        print("done loop", flush=True)
```
eric-anderson, line 36: you set up logging, so use it.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -131,10 +131,7 @@ def convert_from_path_streamed_batched(filename: str, batch_size: int) -> Genera
 
 
 def pdf_to_image_files(pdf_path: str, file_dir: Path, resolution: int = 200) -> Generator[Path, None, None]:
-    """Writes the files (streamed) into file_dir.  Caller is responsible for calling
-    path.unlink() to cleanup the files.
-
-    Note: model service will call this to get images for processing"""
+    """Deprecated.  Switch to PdfToImageFiles"""
```
eric-anderson, line 134: Write a followup PR to remove uses of pdf_to_image_files. Then delete the function.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -245,3 +242,152 @@ def read_stderr(fh, q):
...
+    """
+    Writes the files (streamed) into file_dir.  Caller is responsible for calling
+    path.unlink() to cleanup the files.
+
+    Note: model service will call this to get images for processing
+    """
+
+    def __init__(self, *, pdf_path: str | Path, file_dir: str | Path, resolution: int = 200) -> None:
+        self.file_dir = Path(file_dir)
+        self.in_cm = False
```
eric-anderson, line 265: please write this out as in_contextmanager.  It's a rarely used variable and this means I don't have to reason through what it's doing.

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -245,3 +242,152 @@ def read_stderr(fh, q):
...
+        assert proc.stdout
+        assert proc.stderr
+        if proc.poll() is None:
+            self.running = False
+            proc.terminate()
+            try:
+                proc.wait(2)
+            except TimeoutExpired:
+                proc.kill()
+                proc.wait(1)
```
eric-anderson, line 292: log a message if we go through this path. We shouldn't, and if the kill fails (hung fs or something) then the joins are going to silently hang and it will be confusing.

eric-anderson (approved):

lib/sycamore/sycamore/utils/pdf.py
alexaryn: If the wait() after the SIGKILL times out, an unhandled TimeoutExpired exception will be raised, which I thought was sufficient messaging.

lib/sycamore/sycamore/utils/pdf.py
alexaryn: Should we do the same with `convert_from_path_streamed()` at the top of the file which is also marked as deprecated?

eric-anderson (approved):

lib/sycamore/sycamore/utils/pdf.py
```diff
@@ -245,3 +247,129 @@ def read_stderr(fh, q):
...
+        proc = self.proc
+        assert proc.stdout and proc.stderr
+        poller = select.poll()
+        out = NbReader(proc.stdout, poller)
+        err = NbReader(proc.stderr, poller)
+        seen_header = False
+        image_size = 0
+        image_num = 0
+
+        while not (out.eof and err.eof):
```
eric-anderson, line 335:
```
 while not (out.eof and err.eof):
      out_would_block = out.do_read()
      err_would_block = err.do_read()
      if out_would_block and err_would_block:
           ... poll...
```

PR #1603, opened 2026-05-31
Switch LLM cache from pickle to JSON.

bsowell:
Perf impact will be minimal compared to LLM cost, and pickle has a bunch of issues.

### DIFF
--- a/lib/sycamore/sycamore/llms/llms.py
+++ b/lib/sycamore/sycamore/llms/llms.py
@@ -2,8 +2,9 @@ import inspect
 from abc import ABC, abstractmethod
 import copy
 import datetime
+import io
+import json
 import logging
-import pickle
 import base64
 from PIL import Image
 from typing import Any, Optional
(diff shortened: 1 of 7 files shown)


lib/sycamore/sycamore/llms/llms.py
```diff
@@ -140,14 +141,77 @@ def __str__(self):
...
+            return datetime.datetime.fromisoformat(obj["value"])
+        return obj
+
+    @staticmethod
+    def _llm_cache_json_dumps(data: Any) -> str:
+        return json.dumps(data, default=LLM._llm_cache_json_default, sort_keys=True, separators=(",", ":"))
+
+    @staticmethod
+    def _llm_cache_jsonable(data: Any) -> Any:
+        return json.loads(LLM._llm_cache_json_dumps(data))
```
HenryL27, line 209: I do wonder if it would be more efficient to just leave it as a string. I guess double-json-encoding it would be kinda weird too. idk.

HenryL27 (approved):

PR #1606, opened 2026-07-31
Add Ollama backend for local LLM generation and embeddings

### LINKED ISSUE #1605: Add Ollama backend (local LLM + embeddings)
**Is your feature request related to a problem? Please describe.**
All of Sycamore's current LLM (`sycamore.llms`) and embedding (`sycamore.transforms.embed`)
providers are hosted APIs -  OpenAI, Anthropic, Bedrock, Gemini. There's no way to run the LLM
or embedding steps of a DocSet pipeline against a local model server. That's a real gap for
anyone building a fully local-first or offline-capable pipeline - related in spirit to #707,
though that issue is specifically about bundling partitioner model weights rather than
LLM/embedding providers.

I'm building a local-first RAG system (part of an MCP-based project) that already runs
entirely on Ollama for generation and embeddings, and I'd like to use Sycamore's
partitioning/DocSet pipeline without introducing a cloud dependency into an otherwise fully
local stack.

**Describe the solution you'd like**
Add:
- `OllamaLLM` in a new `sycamore/llms/ollama.py`, subclassing the `LLM` ABC, using the
  official `ollama` PyPI client.
- `OllamaEmbedder` in `sycamore/transforms/embed.py`, subclassing `Embedder`, matching
  `OpenAIEmbedder`'s constructor/batching shape.
- A new optional `ollama` extras group in `pyproject.toml`, so this has zero impact on
  existing installs.

I have a working draft of both, plus unit tests that mock the client, and I'm ready to open
a draft PR for early feedback if this is a welcome addition.

**Describe alternatives you've considered**
Ollama also exposes an OpenAI-compatible endpoint (`/v1/chat/completions`, `/v1/embeddings`),
so in principle `OpenAI`/`OpenAIEmbedder` could already point at a local Ollama server with
no new code. I think a native provider is still worth it - `OLLAMA_HOST` env var resolution,
Ollama-native `options` (`num_ctx`, `mirostat`, etc.) not exposed through the compat layer,
and clearer errors when the actual backend is Ollama rather than a generic OpenAI-client
exception - but flagging this in case maintainers would rather just document the
compat-endpoint route instead.

**Additional context**
For `OllamaLLM` I mirrored `anthropic.py` (direct client + `__reduce__` for Ray-picklability)
rather than `openai.py`, since the latter carries Azure/Helicone/batch-API handling Ollama
has no equivalent of. Ollama's model catalog is just "whatever's been `ollama pull`ed" rather
than a fixed list, so `OllamaModels.from_name()` falls back to constructing an ad hoc model
for unrecognized strings instead of raising, unlike the other providers - flagging that as an
intentional deviation in case it should work differently.

Happy to take this on if it's a welcome addition - I'll open a draft PR as soon as I get a
thumbs up.

SkyeVault:
## Summary

Adds Ollama as a native `LLM` and `Embedder` provider so DocSet pipelines can run generation and embedding steps against a local Ollama server instead of a hosted API, for fully local/offline-capable pipelines with no cloud dependency.

Closes #1605

- `sycamore/llms/ollama.py`: `Ollama(LLM)`, using the official `ollama` PyPI client directly.
- `sycamore/llms/config.py`: `OllamaModel`/`OllamaModels`.
- `sycamore/transforms/embed.py`: `OllamaEmbedder(Embedder)`, matching `OpenAIEmbedder`'s constructor/batching shape (Ollama's `/api/embed` natively accepts a batch of inputs in one call).
- New optional `ollama` extras group in `pyproject.toml`, zero impact on existing installs.
- Unit tests mocking the `ollama` client for both the LLM and Embedder (`test_ollama.py`, additions to `test_embed.py`).
- Docs: `llm.rst`, `embed.rst`, `embed.md`.

## Two design calls worth flagging for review

1. **Native `ollama` client vs. the OpenAI-compat endpoint.** Ollama also exposes an OpenAI-compatible `/v1/chat/completions` and `/v1/embeddings` API, so in principle `OpenAI`/`OpenAIEmbedder` could already point at a local Ollama server with no new code. I went with a native provider instead, for `OLLAMA_HOST` env var resolution and Ollama-native `options` (`num_ctx`, `mirostat`, etc.) that aren't exposed through the compat layer, plus clearer errors when the actual backend is Ollama. Flagging in case maintainers would rather just document the compat-endpoint route instead.
2. **Unrecognized model names don't raise.** Every other provider's `FooModels.from_name()` raises on an unrecognized string. Ollama's catalog is just whatever's been `ollama pull`ed rather than a fixed list, so `OllamaModels.from_name()` builds an ad hoc model for unrecognized names instead. Flagging as an intentional deviation in case it should work differently.

For `Ollama` itself I mirrored `anthropic.py` (direct client plus `__reduce__` for Ray-picklability) rather than `openai.py`, since the latter carries Azure/Helicone/batch-API handling that Ollama has no equivalent of.

## Test plan

- [x] `poetry run pytest lib/sycamore/sycamore/tests/unit/llms/` and `.../tests/unit/transforms/test_embed.py`, all passing, including pre-existing tests in both files.
- [x] `ruff check`, `black --check`, `mypy` on all new/changed files, clean.
- [x] Verified end-to-end against a real local Ollama server (not just mocks): `qwen2.5-coder:14b` for title extraction and `nomic-embed-text` for embeddings, on a full `read -> partition -> extract_entity -> explode -> embed` DocSet pipeline against a real PDF.

Opening as a draft since I haven't heard back on #1605 yet. Happy to adjust based on feedback on either design call above.


lib/sycamore/sycamore/llms/ollama.py
```diff
@@ -0,0 +1,183 @@
...
+from sycamore.llms.prompts import RenderedPrompt
+from sycamore.utils.cache import Cache
+from sycamore.utils.image_utils import base64_data
+from sycamore.utils.import_utils import requires_modules
+
+logger = logging.getLogger(__name__)
+
+
+def format_image(image: Image.Image) -> str:
+    return base64_data(image)
```
HenryL27: just call base64_data directly where needed

lib/sycamore/sycamore/llms/ollama.py
```diff
@@ -0,0 +1,183 @@
...
+            "default_mode": self._default_mode,
+            "default_llm_kwargs": self._default_llm_kwargs,
+            "client_args": self._client_args,
+        }
+        return ollama_deserializer, (kwargs,)
+
+    def default_mode(self) -> LLMMode:
+        if self._default_mode is not None:
+            return self._default_mode
+        return LLMMode.ASYNC
```
HenryL27: how could self._default_mode be None?

lib/sycamore/sycamore/transforms/embed.py
```diff
@@ -285,6 +286,80 @@ def embed_texts(self, texts: List[str]) -> List[List[float]]:
...
+        self.host = host
+        self._client_args = client_args or {}
+        self._client: Optional["OllamaClient"] = None
+
+        super().__init__(
+            model_name=model_name,
+            batch_size=batch_size,
+            model_batch_size=model_batch_size,
+            pre_process_document=pre_process_document,
+            device="cpu",
```
HenryL27: seems like it should be a param with a default

HenryL27:
In terms of model registry stuff, a couple thoughts
Is it worth querying the ollama instance to get the list of known models/does my model exist before runtime? In general in sycamore it's good to get errors before executing a pipeline if possible.
Maybe we can add a `pull: bool = False` param somewhere to tell it to pull a model if it's not found? Probably not actually. That would be overcomplicated.

SkyeVault:
On the model registry stuff:

Fail-fast would be nice here, but it means construction needs a live server connection, and even the other providers can't fully validate model names against reality (just against a hardcoded list). If it'd help, happy to add an opt-in check like `validate=False` so it stays fast by default. Let me know if that's useful.

On `pull`: agreed, let's skip it. Auto-pulling a multi-GB model as a side effect of creating an object feels like a bad surprise. I'd rather keep `ollama pull` an explicit, separate step.

